PR #8, opened 2019-04-08
fix(admin): bug fix and new features

kilchenmann:
- [x] new "/system" route with list of projects / users
- [x] 403-error on "/system" route, when user is not sysAdmin
- [x] same toolbar style everywhere
- [x] highlight in each tab, if project is archived
- [x] bug fix: change not own user-password
- [x] bug fix: user's projects
  - [x] request for each project
  - [x] cache this list
  - [x] update cache on changes
  - [x] destroy all user specific cache on logout
- [x] fix permissions: update variables like sysAdmin and projectAdmin and disable functions or buttons in the template
- [x] add button to reactivate project
  - [x] perhaps we need general modify-menu on each list item
- [x] dialog improvements
  - [x] dialog-header
  - [x] confirm dialog (YES / NO)
  - [x] refactor refresh parent method


kilchenmann:
[Kuirl: admin interface](https://trello.com/c/zE2Ti9Lp/69-kuirl-admin-interface)

PR #12, opened 2019-05-03
feat: Get app version and display in header

kilchenmann:
Get the version of the app and of Knora and display them somewhere.


kilchenmann:
There's an issue with reading the header information from Knora api....

kilchenmann:
[Kuirl: server infos](https://trello.com/c/THPMSwkt/72-kuirl-server-infos)

PR #54, opened 2019-08-24
style: Fix position of mat-menu and box-shadow

### LINKED ISSUE #53: style: Position of mat-menu items
**Describe the bug**
The position of mat-menu in the list is wrong when this is the last item (it does not appear) or when the menu appears at the bottom of the screen (it's cropped or appears far away from the button). The shadow effect surrounding the div is also missing.

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to 'http://localhost:4200/system/users'
1. Click on an 'item menu button (icon with 3 dots)'
1. See error

**Expected behaviour**
The mat-menu should appear close by the button (on the top or bottom or left). The box-shadow should be visible. 

**Screenshots**
Screen Shot 2019-08-23 at 14 53 25

**Additional context**
N.A.

flavens:
resolve #53

### DIFF
--- a/src/assets/style/_elements.scss
+++ b/src/assets/style/_elements.scss
@@ -477,9 +477,11 @@ $gc-small: $form-width - $gc-large - 4;
 // dialog box / material overlay
 
 .cdk-overlay-pane {
-  margin-bottom: 120px !important;
-  overflow-y: auto !important;
-  max-height: 80% !important;
+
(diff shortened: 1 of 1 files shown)


kilchenmann:
I changed the height of mat-dialog-container class to max-height, otherwise it looks as follow:
Screenshot 2019-08-24 at 15 14 12

PR #57, opened 2019-08-26
Pre-release v1.4.0

kilchenmann:


flavens:
I am on it!

flavens:
@kilchenmann I can not edit a list item, I get a 400 error, is it only me?

kilchenmann:
@flavens I will test...

kilchenmann:
But why it shouldn't work anymore? I worked on 5 files only in this branch

flavens:
Ok so I can edit a list in the anything project but not in the BEOL project... Maybe there is some permissions on it?! Anyway except this point, I can approve the PR if there is nothing more to fix/add!

kilchenmann:
In the test-data of BEOL is a wrong IRI but I have to read the shortcode from the IRI. In this (only in this) case, the shortcode is not part of the beol iri. This is something we have to fix when we make the list editor better

PR #92, opened 2019-09-27
Fix style issue in lists

kilchenmann:
On large screen we had an issue with the indent of list-item form. This PR fixes this style issue


flavens:
@kilchenmann this branch needs the latest Knora-UI modules version to be tested in Safari!

kilchenmann:
>  this branch needs the latest Knora-UI modules version to be tested in Safari

@flavens why? This is just a styling issue. It will get the latest version, after merging the other two PRs and by updating this branch from master. But it doesn't need a new version of the modules for the lists.

Btw I wasn't able to reproduce the issue with the list items on my small notebook screen. It happend only on the wide screen and bigger browser window.

kilchenmann:
@flavens Thank you!

PR #232, opened 2020-01-09
Add some unit tests and e2e tests

flavens:
First step to add more tests.

Starting with:
- [x] main component: unit + e2e
- [x] help component: e2e
- [x] login component: e2e

To run the tests: `ng e2e --prod=true --protractor-config=./e2e/protractor-ci.conf.js --webdriver-update=false`


src/app/main/main.component.spec.ts
```diff
@@ -17,6 +17,10 @@ import { MainComponent } from './main.component';
 describe('MainComponent', () => {
     let component: MainComponent;
     let fixture: ComponentFixture<MainComponent>;
+    let element: HTMLElement;
+
+    const config = new KnoraApiConfig('http', '0.0.0.0', 3333);
```
kilchenmann: @flavens Why did you removed the global defined apiConfig? `TestConfig.ApiConfig` has the same information as your `const config` here. I thought it's a good idea to define the configuration in one central place. What happens when we have to change the api config e.g. other port or protocol? In this case we have to update the config in each spec file.

flavens: @kilchenmann I did it because I need to create an instance of KnoraApiConfig (outside of the TestBed) to use it in the it tests. But I could not figure out how to mock some data and complete tests. Waiting for the MockFactory from the lib, I will undo these lines..

PR #258, opened 2020-02-07
Add tests to project components

flavens:
- board component


flavens:
@kilchenmann I will need your help to mock the cache service because I face 2 issues:
- I cannot create a partial stub of the cache service because the main properties (cache especially) are declared as private variables
- I have to mock a cache prop of type Map<string, CacheContent> and I have no idea how doing this...

there is maybe an easier solution

flavens:
Waiting for a knora-api-js-lib dev package to be published.

PR #292, opened 2020-09-14
docs: Update README

kilchenmann:
I had do remove a dead link from README


flavens (approved):
I would fix the typo before merging: "Additionally" instead of "Additional" and "GitHub" instead of "github"

PR #295, opened 2020-09-16
Login Success Message

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-331


mdelez:
@flavens do you have an opinion on what we should do about the default message CSS having a background color opacity of 0.4? This leads to text on the page overlapping with the message box. Should we remove the opacity or make it so that the message box doesn't cover any text on the page? We could also make it so that the width of the message box is equivalent to the length of the message text (with a few pixels of padding) which would solve this issue for this case but maybe not on other pages where there is text in the middle of the page.

Screen Shot 2020-09-16 at 10 44 45

mdelez:
change to the css will be done in https://github.com/dasch-swiss/dsp-ui-lib/pull/180

src/app/main/header/header.component.ts
```diff
@@ -45,6 +57,11 @@ export class HeaderComponent {
         });
     }
 
+    ngOnInit() {
+        this.componentCommsSubscription = this._componentCommsService.on(
```
flavens, line 61: you should unsubscribe to the subscription

src/app/main/header/header.component.spec.ts
```diff
@@ -94,4 +100,11 @@ describe('HeaderComponent', () => {
         expect(searchPanel).toBeDefined();
     });
 
+    it('should display the login success message when the LoginSuccess event is emitted', () => {
```
flavens, line 103: you can test the unsubscription

flavens (changes requested):

src/app/main/header/header.component.ts
mdelez: good catch! fixed in 73b3db3

src/app/main/header/header.component.spec.ts
mdelez: fixed in 73b3db3

flavens (approved):

PR #298, opened 2020-09-18
DSP-298 Add app name to header

kilchenmann:
In this PR I tried to add the DSP name to the header. On desktop devices it's called "DaSCH Service Platform" on tablets and smaller devices it's "DSP" only. But we have to refactor the responsive style in the header or in general anyway.
Furthermore I replaced the lamp-logo with the DaSCH logo-icon

Resolves [DSP-298](https://dasch.myjetbrains.com/youtrack/issue/DSP-298)


flavens (commented):
looks good!
should we keep the logo in black? in which case should we use the DaSCH colour palette?
You should also add the link to the YT issue in the description.

kilchenmann:
> looks good!
> should we keep the logo in black? in which case should we use the DaSCH colour palette?

No, but we will change the logo color with the CD user story. DSP-664

> You should also add the link to the YT issue in the description.

Sure, I forgot. Will do it.

flavens (approved):

PR #299, opened 2020-09-21
Fix tiny issues

kilchenmann:
Resolves [DSP-679](https://dasch.myjetbrains.com/youtrack/issue/DSP-679)

- Clean up the github issue templates which should be used by extern reports.
- Fix security issue
- Update prod-config example
- No dirty tag in docker image: This is a temporary solution and will be updated as soon we use bazel everywhere.
- Update badges in README


flavens (approved):
a DSP issue should be related to this PR, if so please make a link

PR #302, opened 2020-09-23
DSP-371 - Improve the visibility of expert and advanced search links

flavens:
resolves [DSP-371](https://dasch.myjetbrains.com/youtrack/issue/DSP-371)

### DIFF
--- a/src/app/main/header/header.component.scss
+++ b/src/app/main/header/header.component.scss
@@ -1,7 +1,10 @@
 @import "../../../assets/style/responsive";
 
+.toolbar-header {
+    height: 72px;
+}
+
 .logo {
-  // margin-left: -16px;
 
   .mat-icon {
     height: 40px;
@@ -11,7 +14,6 @@
     }
 
     &.small {
-      // margin-left: 12px;
       width: 40px;
     }
   }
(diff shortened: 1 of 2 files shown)


src/app/main/header/header.component.scss
```diff
@@ -89,4 +113,12 @@
   .action {
     padding: 0;
   }
+  .advanced-search-link {
+      background-color: lavender;
```
kilchenmann, line 117: Do you want to keep this color?

src/app/main/header/header.component.scss
```diff
@@ -89,4 +113,12 @@
...
+      margin-top: -100px;
+  }
+  .expert-search-link {
+    background-color: indigo;
```
kilchenmann, line 121: Do you want to keep this color?

src/app/main/header/header.component.scss
```diff
@@ -41,6 +43,28 @@
   margin: 0 16px;
 }
 
+.other-search-link {
+    display: grid;
```
kilchenmann, line 47: Why do you use grid here?

kilchenmann (commented):
Looks already good to me. I have just some questions...

src/app/main/header/header.component.scss
flavens: No, I forgot to remove these classes > done in ec8c3ea

src/app/main/header/header.component.scss
flavens: No, I forgot to remove these classes > done in ec8c3ea

src/app/main/header/header.component.scss
flavens: because I prefer this solution to get a blank space > button advanced > button expert, instead of using a class "fill-in-remaining-space" (I tried it though but I could not make it look good as I wanted)
and one day if we want to move the project filter menu here, the grid will be helpful to position it well

kilchenmann: Thanks for explanation

kilchenmann (approved):

PR #307, opened 2020-09-25
DSP-368 - Redundancy with Cancel button and close button in dialog boxes

flavens:
resolves [DSP-368](https://dasch.myjetbrains.com/youtrack/issue/DSP-368)

### DIFF
--- a/src/app/main/dialog/dialog.component.html
+++ b/src/app/main/dialog/dialog.component.html
@@ -201,7 +201,7 @@
     </div>
 
     <div *ngSwitchCase="'editResourceClass'">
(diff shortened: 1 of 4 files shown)


src/app/main/dialog/dialog.component.html
```diff
@@ -212,11 +204,15 @@
         <app-dialog-header [title]="data.title" [subtitle]="data.subtitle" (close)="dialogRef.close()">
```
kilchenmann, line 204: This header still has the `(close)` parameter

flavens: fixed in ee60add

kilchenmann (approved):
I think you replaced it now everywhere

PR #310, opened 2020-09-28
DSP-339 - Add a link to the membership management page from the system project page

flavens:
resolves [DSP-339](https://dasch.myjetbrains.com/youtrack/issue/DSP-339)


src/app/system/projects/projects-list/projects-list.component.ts
```diff
@@ -95,12 +95,27 @@ export class ProjectsListComponent implements OnInit {
...
             () => this._router.navigate(['/project/' + code])
         );
     }
 
+    /**
+     * Navigate to the collaboration / project members page
+     *
+     * @param code
+     */
+    gotoProjectProjectMembers(code: string) {
```
kilchenmann, line 114: It would be nice if we have a combination with the method "gotoProjectBoard". You can name it for example as openProject(code: string, page: 'collaboration' | 'ontology'). etc. In this case we are more flexibel and can reuse the function. Otherwise we have to add a method for more links. Maybe someone wants a link to the ontology from this list of projects. Do we have to create then an addition method "gotoProjectOntologies"?

There's anyway a typo (project is twice) in the method name:  gotoProjectProjectMembers

flavens: Do you want to implement now the link to the project ontology page?

kilchenmann: We can add the link now when it's necessary. It was just an idea to combine methods which do the (almost) same thing. This is why we have methods.

flavens: methods combination done in 7c9773a

kilchenmann (approved):

PR #321, opened 2020-10-15
DSP-851 New resource form

flavens:
resolves [DSP-851](https://dasch.myjetbrains.com/youtrack/issue/DSP-851)

this PR has been developed with dsp-api v13.0.0-rc.17


flavens:
@mdelez the yalc process to get the required local dsp-ui package:
1. checkout the branch `wip/dsp-881-export-ValueOperationEventService`
2. in your terminal in the dsp-ui-lib repo, run `npm run yalc-publish-lib`
3. in your dsp-app repo, run `yalc add @dasch-swiss/dsp-ui && npm install`
4. in dsp- app, run `npm run ng s` and see error

mdelez:
@flavens I got something working. You'll need the changes I made to the UI-Lib as well [here](https://github.com/dasch-swiss/dsp-ui-lib/pull/210). I recommend using a form with not too many properties to test the validation, I've been using a website resource class in the biblio ontology of the BEOL project.

I've been trying to find a way to add each FormGroup to the parent FormGroup with a unique name since they need to be unique or else it will be overwritten...unsuccessful so far. This will be important when we implement a plus button to allow users to add multiple values per property.

mdelez:
**Breaking change incoming from DSP-UI [PR 214](https://github.com/dasch-swiss/dsp-ui-lib/pull/214)**

renaming of ValueTypeService to ValueService

affected code: https://github.com/dasch-swiss/dsp-app/blob/4b2920d1200ea929131ba1e550f10233aba2dfb8/src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts#L36

mdelez:
> **Breaking change incoming from DSP-UI [PR 214](https://github.com/dasch-swiss/dsp-ui-lib/pull/214)**
> 
> renaming of ValueTypeService to ValueService
> 
> affected code:
> 
> https://github.com/dasch-swiss/dsp-app/blob/4b2920d1200ea929131ba1e550f10233aba2dfb8/src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts#L36

fixed in 54c4670

kilchenmann:
Just some details I figured out:
- In case of system admin, all projects should be accessible.
- In case user is member in only one project, the user doesn't have to select. Same for ontology if there's only one ontology in the project, the user doesn't need a select box or it should already be selected. Same for resource class.
- Label input field could be confusing: In the old salsah we used the value (stringified) of the first property as the label. There was no Input field for label. This is something we have to discuss together and with @lrosenth.
- When clicking on "back" button, the form is empty. It can be frustrating to select again all needed fields.

src/app/workspace/resource/resource-instance-form/select-properties/switch-properties/switch-properties.component.html
```diff
@@ -0,0 +1,20 @@
...
+    <dsp-interval-value #createVal *ngSwitchCase="constants.IntervalValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"></dsp-interval-value>
+    <dsp-time-value #createVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"></dsp-time-value>
+    <dsp-geoname-value #createVal *ngSwitchCase="constants.GeonameValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"></dsp-geoname-value>
+    <dsp-link-value #createVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"
+                [parentResource]="parentResource" [propIri]="property.id"></dsp-link-value>
+    <dsp-date-value #createVal *ngSwitchCase="constants.DateValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"></dsp-date-value>
+    <dsp-list-value #createVal *ngSwitchCase="constants.ListValue" [mode]="mode" [parentForm]="parentForm" [formName]="formName"
+                [propertyDef]="property"></dsp-list-value>
+    <span *ngSwitchDefault>
+        <p>Cannot match any value component.</p>
```
kilchenmann, line 18: Could you add the property.objectType here? So that we know, which value component is missing here?
e.g.

```html
    <span *ngSwitchDefault>
        <p>Cannot match any value component for <strong>{{property.objectType}}</strong></p>
    </span>
```

kilchenmann (commented):
It looks already nice. Just some comments 😉

mdelez:
> Just some details I figured out:
> 
> * In case of system admin, all projects should be accessible.
> * In case user is member in only one project, the user doesn't have to select. Same for ontology if there's only one ontology in the project, the user doesn't need a select box or it should already be selected. Same for resource class.
> * Label input field could be confusing: In the old salsah we used the value (stringified) of the first property as the label. There was no Input field for label. This is something we have to discuss together and with @lrosenth.
> * When clicking on "back" button, the form is empty. It can be frustrating to select again all needed fields.

How crucial would you consider these things for the first implementation of this feature? I think Flavie and I are eager to get what we have so far into the next release

src/app/workspace/resource/resource-instance-form/select-properties/switch-properties/switch-properties.component.html
mdelez: added in 5f4bb1d

kilchenmann:
> How crucial would you consider these things for the first implementation of this feature? I think Flavie and I are eager to get what we have so far into the next release

I agree. Just a comment; will move it to the user story 😉

mrivoal:
Sorry to barge in like this :-) 
@kilchenmann 

> Label input field could be confusing: In the old salsah we used the value (stringified) of the first property as the label. There was no Input field for label. This is something we have to discuss together and with @lrosenth.

I can't remember how the old Salsah worked but in the version that we use now (Salsah with Knora as back-end), there is an input field for the label.

And I agree with you, it is confusing. For every new project, the reason for this label must be explained to users: when you want to create a link between 2 resources, the link property opens a search box which uses the label of the resource you want to point to to create the link...

In some cases, using the first property as label would not provide enough information. Most of the time, the label chosen by users is a selection of at least 2 properties and sometimes much more. Actually, in one case, a plugin was even developed by a user so that each resource class receives a label consisting of a custom concatenation of properties. 

I thought it could be useful for you to know.

flavens:
@kilchenmann @mrivoal about the resource label, I agree that we could take the first property value as a resource label. However, it is currently not possible. In some ontologies, the first property does not make sense to use as a resource label. It could be set/updated in the ontology editor but this is not implemented yet. A short discussion should be planned to decide what to do. Meanwhile, we will keep the current behaviour, it will be improved as soon as we make a decision.

kilchenmann (approved):
I checked out the branch and tested it and it worked well.

PR #324, opened 2020-10-27
DSP-861 Create resource property

kilchenmann:
resolves DSP-861 and DSP-862 and closes DSP-751

With this PR we will have a first version of the ontology creator. This means you can create and delete an ontology. Create and delete a resource class and create resource properties. The properties will be created in case they do not exist. And it's possible to use existing properties in each resource class. It still needs some improvements. One missing thing is the permission part. You can't set permission on resource class neither on resource property yet. Comments on properties are also missing at the moment. This can be solved in separate tasks (s. DSP-1145).

But you can set cardinality and gui-order, select other resource classes and lists in corresponding properties (links and lists).

See following screen record with an example of a simple ontology with two resource classes: "Person" and "Photograph". 
<https://dasch.myjetbrains.com/youtrack/api/files/7-225?sign=MTYwNzQxNDcwMDAwMHwxLTl8Ny0yMjV8X2x2emhoc1o3amk1cjJaUmxhRGNpNjJSN3VLOVhYN05j%0D%0ASTNmWnlHVGVyMA0K%0D%0A&updated=1607381473832>
p.s. if the link to the video does not work, you'll find it also in the comments of DSP-861

Btw the name or the iri of resource class and resource property is created from the label (with a lot of regex 😉).


tobiasschweizer:
@kilchenmann Sorry, I accidentally clicked "Update branch" on this PR.

kilchenmann:
> @kilchenmann Sorry, I accidentally clicked "Update branch" on this PR.

no worries

src/app/project/ontology/ontology.component.html
```diff
@@ -1,5 +1,7 @@
 <div *ngIf="projectAdmin" class="desktop-only">
 
+    <p class="note warning center">This is a first version of the ontology editor. Not all features are implemented yet.
+        Be careful when using it!</p>
```
flavens, line 4: "Be careful when using it!": this turn of phrase is quite scary for the user...

flavens (changes requested):
I've got different strange behaviours:
1. I updated an existing data model (when creating a new data model, I set different type of names, with or without capital, long/short but the form stays invalid without any error message. I have no idea what is going on.)

2. during the creation of the resource's prop, when I selected a name that I have used before with the same type, the name and type are then disabled. It is really confusing.
Screenshot 2020-12-08 at 09 34 30

And when submitting the form with this "disabled" prop, I got a 400 error from dsp-api:
Screenshot 2020-12-08 at 09 36 35

kilchenmann:
> I've got different strange behaviours:
> 
> 1. I updated an existing data model (when creating a new data model, I set different type of names, with or without capital, long/short but the form stays invalid without any error message. I have no idea what is going on.)
> 2. during the creation of the resource's prop, when I selected a name that I have used before with the same type, the name and type are then disabled. It is really confusing.

This will be explained in an info box (hint) as well in the guidelines. When you use an already defined property, you can't change the type anymore, because it's already define. You can reuse it as it is and change only the cardinality incl. guiOrder and permission (permissions are not yet implemented).

> 
> Screenshot 2020-12-08 at 09 34 30
> 
> And when submitting the form with this "disabled" prop, I got a 400 error from dsp-api:
> Screenshot 2020-12-08 at 09 36 35

The error message comes from the API. I know it's not clear what's happening there. This is something we have to improve in DSP-Api but also in DSP-UI in notification service. Can you check the error message form API directly in developer tools "network"?

src/app/project/ontology/ontology.component.html
kilchenmann: As you know, we can always change my english sentences. I'm open for suggestions 😉

kilchenmann: "use it on your own risk" 😂
No, just kidding. I thought it would be good to give a hint, that we know about the improvements and it's still work in progress. But it would be good to have an ontology create to start with simple ontologies on prod server. I need this tool for some projects in January...

mdelez: "Some features may not work as intended" is usually the corporate way of saying "Please don't be an idiot" 😂

kilchenmann:
@flavens I'm not able to reproduce the error. But it always depends on the already defined property type. The problem in the ontology creator is, that we have to make so many checks to be sure what will be allowed and what not from the API side. I'll try to implement a better error handler.

src/app/project/ontology/ontology.component.html
kilchenmann: Done in 122c8b8

mdelez:
I noticed that when I enter a name for a new data model, the create button does not enable until I make a change to the label. I think this is due to automatically entering the label for the user based on the name which messes with the form validation. A possible solution could be to remove formControl of the label and immediately add it again every time a change to the name is made, which isn't exactly elegant.

kilchenmann:
> I noticed that when I enter a name for a new data model, the create button does not enable until I make a change to the label. I think this is due to automatically entering the label for the user based on the name which messes with the form validation. A possible solution could be to remove formControl of the label and immediately add it again every time a change to the name is made, which isn't exactly elegant.

Yes, this was a known issue. I solved it in 3a3d2a8 as follow:
Ontology label is not required anymore. In case it's empty, it uses the name as value. 
This quick solution can of course still be improved.

kilchenmann:
>> I've got different strange behaviours:

>> I updated an existing data model (when creating a new data model, I set different type of names, with or without capital, long/short but the form stays invalid without any error message. I have no idea what is going on.)
during the creation of the resource's prop, when I selected a name that I have used before with the same type, the name and type are then disabled. It is really confusing.

> This will be explained in an info box (hint) as well in the guidelines. When you use an already defined property, you can't change the type anymore, because it's already define. You can reuse it as it is and change only the cardinality incl. guiOrder and permission (permissions are not yet implemented).

@flavens I solved it as follow: By select an existing property it displays a little hint:

Screen Shot 2020-12-08 at 14 12 42

Resolved in 4bf19e5

src/app/project/ontology/default-data/default-properties.ts
```diff
@@ -131,22 +146,25 @@ export class DefaultProperties {
...
-                    icon: 'compare_arrows',
-                    label: 'External resource',
-                    subPropOf: 'knora-api:ExternalResValue',
-                    gui_ele: 'salsah-gui:SimpleText',
+                    subPropOf: Constants.HasLinkTo,
+                    objectType: Constants.Resource,
+                    gui_ele: Constants.SalsahGui + Constants.Delimiter + 'Searchbox',    // 'Autocomplete',
                     group: 'Link'
                 },
+                // {
```
flavens, line 154: is this type waiting for something to make it operational?

flavens (approved):
The overall looks good, maybe some commented lines without explanation could be removed if not necessary for future improvements. There are no unit tests, this is the missing point. 
I would suggest to wait for a 2nd review before merging.

src/app/project/ontology/default-data/default-properties.ts
kilchenmann: Yes this is something we have to discuss. I'm not sure if we need it. The same for duration and period. It does not exist in knora-api resp. in salsah gui ontology. It could be handled as gui attribute. I want to include client services into the discussion as soon, they can test it.
I just created DSP-1154 about this topic.

lrosenth:
Looks good from my side. Code -as far as I understand it – is clean and meaningful. Go ahead!

lrosenth (approved):
OK!

PR #329, opened 2020-11-13
DSP-1062 - Form "back" button functionality polish

mdelez:
resolves DSP-1062

### DIFF
--- a/src/app/workspace/resource/resource-instance-form/resource-instance-form.component.html
+++ b/src/app/workspace/resource/resource-instance-form/resource-instance-form.component.html
@@ -26,6 +26,7 @@
     <!-- select one resource -->
     <span *ngIf="resourceClasses?.length > 0">
         <app-select-resource-class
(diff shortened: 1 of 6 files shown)


flavens (changes requested):
It looks good!
A couple of remarks:
- the resource label is not reset
- when I select the incunabula ontology as a second choice, I get an error:
Screenshot 2020-11-16 at 13 22 03
but it may be a different issue

mdelez:
> It looks good!
> A couple of remarks:
> 
> * the resource label is not reset
> * when I select the incunabula ontology as a second choice, I get an error:
> 
> Screenshot 2020-11-16 at 13 22 03
> 
> but it may be a different issue

first remark is fixed in bf70fb0

Second remark will be handled in [DSP-1061](https://dasch.myjetbrains.com/youtrack/issue/DSP-1061)

flavens (approved):

PR #331, opened 2020-11-18
DSP-1061 Form select box functionality polish

flavens:
resolves DSP-1061


mdelez:
If I'm a normal user and I'm only a member of one project or the project only has one ontology or the ontology only has one resource class, I get the `ExpressionChangedAfterItHasBeenCheckedError` error when the value is automatically chosen.

Screen Shot 2020-11-19 at 09 39 59

I simulated this by reassigning the `ontologies` and `resourceClassDefinitions` array to only the first value of the array
Screen Shot 2020-11-19 at 09 40 21

Did you run into this error as well? If not, how did you test only having one ontology or resource class to choose from?

mdelez:
Actually this happens to a sys admin as well

Screen Shot 2020-11-19 at 09 57 02

flavens:
@mdelez thanks for your comment, I have just spotted the issue but I have no idea of what is going on. I will need a hand on this if you have time.

flavens:
@kilchenmann there is still the described error message posted by Mike. We could not find any solution, we need a new look on it.

kilchenmann (approved):

PR #333, opened 2020-11-19
DSP-1098 - Bump JS-Lib Version and Fix Breaking Changes

mdelez:
resolves DSP-1098

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1631,9 +1631,9 @@
             }
         },
         "@dasch-swiss/dsp-js": {
-            "version": "1.0.0-rc.15",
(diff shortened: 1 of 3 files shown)


mdelez:
@flavens I used the MockResource to get a test resource in the resource-instance-form spec file. I renamed these functions in the most recent js-lib rc (rc.16) so I fixed the breaking changes.

flavens (approved):
ok - the compatibility matrix should updated

mdelez:
> ok - the compatibility matrix should updated
oh right, so I should just increase the version number in the sprint 17 row, right?

PR #334, opened 2020-11-20
DSP-943 - Resource Form Add Additional Value

mdelez:
resolves DSP-943


mdelez:
@flavens I've added what we did during our call as an initial starting point as well as many `TODO` comments. I think you will want to make the key value pairs the same way you did in the `_makeResourceProperties` method.

flavens (approved):
it looks good!! 
I suggest to find a solution for the css class in a task related to dsp-1100
```
::ng-deep .form-content .large-field {
  min-width: 0;
}
```

PR #340, opened 2020-12-09
DSP-1146 Use new methods for ontology info processing

flavens:
resolves DSP-1146


src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
```diff
@@ -350,66 +335,4 @@ export class ResourceInstanceFormComponent implements OnInit, OnDestroy {
...
-    private convertPropObjectAsArray() {
-        // represent the properties as an array to be accessed by the template
-        const propsArray = [];
-
-        for (const propIri in this.properties) {
-            if (this.properties.hasOwnProperty(propIri)) {
-                const prop = this.properties[propIri];
-
-                // only list editable props that are not link value props
-                if (prop.isEditable && !prop.isLinkProperty) {
```
tobiasschweizer, line 405: I could reproduce the problem locally. I think you need to put this logic in `ResourceInstanceFormComponent.selectProperties`:

```ts
    selectProperties(resourceClassIri: string) {

        ...

        // filter out all props that cannot be edited or are link props 
        this.properties = onto.getPropertyDefinitionsByType(ResourcePropertyDefinition).filter(prop => prop.isEditable && !prop.isLinkProperty);
        
          
        ....
```

tobiasschweizer: Otherwise the form creates entries for system props like `knora-api:hasIncomingLinkValue` which you cannot edit.
In the case at hand, the problem was that `knora-api:hasIncomingLinkValue` had no corresponding def. for the link prop. (`knora-api:hasIncomingLink`)

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts
```diff
@@ -17,7 +17,7 @@ export class SelectPropertiesComponent implements OnInit {
```
tobiasschweizer, line 17: I think now you should able to remove

```ts
export interface Properties {
    [index: string]: ResourcePropertyDefinition;
}
```

I think there is only an used import in `src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.spec.ts`

tobiasschweizer (changes requested):
I think you should be able to remove the `typeGuard` from `src/app/workspace/resource/resource-instance-form/select-resource-class/select-resource-class.component.spec.ts` and use the new methods instead.

I think this very similar to what I did in https://github.com/dasch-swiss/dsp-ui-lib/pull/239/files, see `projects/dsp-ui/src/lib/search/advanced-search/select-resource-class/select-resource-class.component.spec.ts`

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.spec.ts
```diff
@@ -3,20 +3,19 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
...
-
-// https://dev.to/krumpet/generic-type-guard-in-typescript-258l
-type Constructor<T> = { new(...args: any[]): T };
-
-const typeGuard = <T>(o: any, className: Constructor<T>): o is T => {
-    return o instanceof className;
-};
-
-let propArrayLength = 0;
+export interface Properties {
```
tobiasschweizer, line 16: Is this still needed?

tobiasschweizer: I believe I got rid of this in dsp-ui serach

tobiasschweizer (approved):
Great, less code to maintain :-)

PR #342, opened 2020-12-10
feat(new-resource-form): make visible the required prop fields (DSP-1115)

flavens:
resolves DSP-1115

the PR must be reviewed with the main unpublished version of dsp-ui using yalc


mdelez:
Note that we currently need to use yalc in order to use [a change](https://github.com/dasch-swiss/dsp-ui-lib/pull/245) made in UI to complete this task.

mdelez:
The asterisks next to required fields seems to work well as well as submitting forms with non-required fields not filled out. I can still click on the "save" button when a required field is not filled out which was planned for. However it results in a console error and nothing is presented to the user. I think it would be good if we could somehow jump to the first required field that the user didn't fill out in this case.

flavens:
> ... I think it would be good if we could somehow jump to the first required field that the user didn't fill out in this case.

I will try this tuto: [Medium - How to scroll to the first invalid control once a form has been submitted](https://medium.com/javascript-everyday/how-to-scroll-to-first-invalid-control-once-a-form-has-been-submitted-eb47d9fbc6e)

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts
```diff
@@ -65,6 +72,23 @@ export class SelectPropertiesComponent implements OnInit {
         );
     }
 
+    isPropRequired(propId: string): void {
```
mdelez, line 75: I think it would make more sense for this method to return a boolean in case we need this logic for other booleans in the future.

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts
```diff
@@ -45,13 +47,18 @@ export class SelectPropertiesComponent implements OnInit {
                     // each property will also have a filtered array to be used when deleting a value.
                     // see the deleteValue method below for more info
                     this.propertyValuesKeyValuePair[prop.id + '-filtered'] = [0];
+
+                     // check the cardinality to know if the prop is required or not
+                     this.isPropRequired(prop.id);
+                     this.propertyValuesKeyValuePair[prop.id + '-cardinality'] = [this.isRequiredProp ? 1 : 0];
```
mdelez, line 53: Can you add a comment explaining why we are adding this to the KeyValuePair?

src/app/workspace/resource/resource-instance-form/select-properties/switch-properties/switch-properties.component.ts
```diff
@@ -20,23 +20,20 @@ export class SwitchPropertiesComponent implements OnInit, AfterViewInit {
...
 
     ngOnInit(): void {
-        // console.log('prop', this.property);
-    }
-
-    ngAfterViewInit() {
-        // console.log('createValueComponent', this.createValueComponent);
-        // this.saveNewValue();
+        // convert from boolean (1/0) to boolean (true/false)
+        this.isRequiredProp = !!+this.isRequiredProp;
```
mdelez, line 32: Maybe you can explain why this is necessary (Input provided by KeyValuePair is stored as a number)

mdelez (changes requested):

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.spec.ts
```diff
@@ -98,7 +98,7 @@ describe('SelectPropertiesComponent', () => {
         }
 
         // each property has two entries in the keyValuePair object
```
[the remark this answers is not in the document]
flavens: fixed in 5dc1abe

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts
flavens: done in 7858d95

src/app/workspace/resource/resource-instance-form/select-properties/switch-properties/switch-properties.component.ts
flavens: done in fb3a064

src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts
flavens: done in f131477

mdelez (approved):
Looks good!

PR #343, opened 2020-12-11
feat(monorepo): Optimization of the workflow for DSP library dependencies

kilchenmann:
resolves DSP-1158

It includes the DSP-JS and DSP-UI with git submodules. This way, we do not need the dependency of those releases.


kilchenmann:
@subotic What do you think? Should we close this PR without merging? At the moment I'm moving a lot of DSP-UI components directly into the app.

kilchenmann:
DSP-1848 will resolve a part of the issues we have with the dependencies

PR #345, opened 2020-12-16
DSP-1128 Trim whitespace from resource label

flavens:
resolves DSP-1128


src/app/workspace/resource/resource-instance-form/select-resource-class/select-resource-class.component.html
```diff
@@ -12,5 +12,8 @@
     <mat-form-field class="large-field resLabel">
         <mat-label>Resource label</mat-label>
         <input matInput formControlName="label" [value]="label" class="label">
+        <mat-error *ngIf="form.controls.label.hasError('pattern')">
+            <span class="custom-error-message">Your label must contain and start with at least one character or one number. Special characters allowed.</span>
```
mdelez, line 16: "Your label must start with a letter or a number and cannot contain spaces."

I think it should be obvious to a user that they can use special characters except for the first character (which we warn them about). They will try to use them anyways if they want them 😅 Is the "no spaces allowed anywhere in the label" intended or does the regex need to be modified?

flavens: fixed in df6a7ae

mdelez (changes requested):
Putting a space at the end of the label also makes it invalid. Should we just trim the whitespace at the end for the user in this case instead of showing them an error message? If not, we should add that the label cannot end with a space to the error message. But I think doing this automatically for the user would be better.

flavens:
> Putting a space at the end of the label also makes it invalid. Should we just trim the whitespace at the end for the user in this case instead of showing them an error message? If not, we should add that the label cannot end with a space to the error message. But I think doing this automatically for the user would be better.

We can trim the whitespace at the end, but how do you do that?

mdelez:
> > Putting a space at the end of the label also makes it invalid. Should we just trim the whitespace at the end for the user in this case instead of showing them an error message? If not, we should add that the label cannot end with a space to the error message. But I think doing this automatically for the user would be better.
> 
> We can trim the whitespace at the end, but how do you do that?

You can modify your regex to allow trailing white space and then just use `.trim()` on the string when the user clicks on `next`. This will also trim leading whitespace but that won't be valid due to the regex anyways.

mdelez (approved):

PR #346, opened 2020-12-16
Wip/dsp 1168 Release please action instead release drafter

kilchenmann:
resolves DSP-1168

- Bring back an empty changelog file. This will be updated by "release please action" after merging this PR
- Add "release please action" and remove "release drafter"

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,16 +29,18 @@ jobs:
           npm run webdriver-update
           npm run test-e2e-protractor
 
-  update_release_draft:
-    name: Update next release draft
-    needs: build-test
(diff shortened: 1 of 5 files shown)


kilchenmann:
@flavens for the review: This is nothing you can test yet. The action script runs when the PR is merged into main branch. It's the same configuration as I already did in DSP-JS and DSP-UI-Lib.
When this PR is merged we have to create one last release manually. Because the new [release-please-action](https://github.com/marketplace/actions/release-please-action) could have problem with all the release candidates. So we have to be careful, but it's not an issue.

flavens (approved):

PR #311, opened 2020-09-30
DSP-710 Show service unavailable in case of api error

kilchenmann:
resolves [DSP-710](https://dasch.myjetbrains.com/youtrack/issue/DSP-710)

PR #355, opened 2021-01-09
feat(error): Server error handler (DSP-710)

kilchenmann:
resolves DSP-710 (s. discussion)

closes #311 

- Implementation of error handler service with following job:
  - depending on error status show message from dsp-ui notification service (mat snackbar) or show full-size error page (case 500 and 503 / server error)

- Add error handler to all components and services where we do api requests (incl. refactoring of Imports)
- Error test page `/teapot` with status 418 ([easteregg](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418))
  - shows the default app error; default means it's not one of the most common error we have: 403, 404, 500 or 503
  - change the status number in `app-routing.module.ts` from 418 to another one (e.g. 403, 404, 500 or 503) to see the different error pages
 
- Test the implementation of error handler:
  - run the app and turn off knora-stack (error 500) or at least fuseki (error 503): it should show a full-size "Service unavailable" error message
  - turn on knora-stack or fuseki again and press "reload page" link in the error message and it should load the page again
  - with turned-off fuseki the only page that works is the `/help` page because there we do not do db-relevant requests.

Comment: We could add a link to <https://status.dasch.swiss> in one of the error 5xx page. So, user will see what's going on on the server. At the moment this service is only implemented for the prod server and not yet for the test (or staging) server which would be <https://status.test.dasch.swiss> (<https://status.staging.dasch.swiss>). Btw with this status page the ops-developer team will be informed when a service doesn't run anymore. So, we don't need a contact button anymore on the server error page to send this kind of error report.


mpro7:
image
Before you get the error page it is possible to see for milliseconds the page under it:
image

src/assets/style/_elements.scss
```diff
@@ -540,7 +540,7 @@ $gc-small: $form-width - $gc-large - 4;
 
 .cdk-overlay-pane {
   .mat-dialog-container {
-    max-height: 80vh !important;
+    // max-height: 80vh !important;
```
mpro7, line 543: why not to remove?

mpro7 (approved):
Seems ok for me. Only that blinking thing while loading the error page I've mentioned in previous comment. If not easy thing to fix, might be moved to another task/bug.

src/assets/style/_elements.scss
kilchenmann: wasn't sure if we still need it and we have to fix the height in another way. But yes, let's delete this line 😉

kilchenmann: done in 924fc3d

kilchenmann:
> image
> Before you get the error page it is possible to see for milliseconds the page under it, with loading animation and snackbar initiated too I think. Here is the recording:
> 
>  Screen.Recording.2021-01-11.at.15.20.51.copy.mov

Yes, I know this issue. For the moment we have to live with it. As you written above, it would be complicated to solve it in a quick way. This full-screen error page opens only in case of server errors. I hope, that we'll not have too many server errors 😆🙈
We should solve it in a separate task...

flavens (approved):
it looks good but I strongly suggest that Mike should check the error messages.

mpro7:
> > image
> > Before you get the error page it is possible to see for milliseconds the page under it, with loading animation and snackbar initiated too I think. Here is the recording:
> > Screen.Recording.2021-01-11.at.15.20.51.copy.mov
> 
> Yes, I know this issue. For the moment we have to live with it. As you written above, it would be complicated to solve it in a quick way. This full-screen error page opens only in case of server errors. I hope, that we'll not have too many server errors 😆🙈
> We should solve it in a separate task...

Ok, could you please create the task if not already did.

kilchenmann:
> it looks good but I strongly suggest that Mike checks the error messages.

As @flavens suggested: @mdelez can you check the error messages? Thanks!

src/app/main/error/error.component.ts
```diff
@@ -8,14 +17,90 @@ import { Title } from '@angular/platform-browser';
...
+        },
+        {
+            status: 503,
+            message: "Service unavailable",
+            description: `The DaSCH Service Platform is not available at the moment.<br>
+            The server is currently unavailable (overloaded or down).`,
+            action: 'reload',
+            image: 'dsp-error-503.svg'
+        }
+    ];
```
kilchenmann, line 65: @mdelez Here are the error messages defined...

src/app/main/error/error.component.ts
```diff
@@ -8,14 +17,90 @@ import { Title } from '@angular/platform-browser';
...
+            message: "Undefined Error",
+            description: `Maybe I'm a teapot, maybe not.<br>
+            But anyway, something undefined went wrong.`,
+            action: 'goback',
+            image: 'dsp-error.svg'
+        },
+        {
+            status: 403,
+            message: "Forbidden",
+            description: `This is not the content you were looking for.<br>
```
mdelez, line 35: I would say "Invalid permissions." instead of "This is not the content you were looking for." because the user was likely looking for this content but just doesn't have the permissions.

src/app/main/error/error.component.ts
```diff
@@ -8,14 +17,90 @@ import { Title } from '@angular/platform-browser';
...
+            Your request was valid but you do not have the<br>
+            necessary permissions to access it.`,
+            action: 'goback',
+            image: 'dsp-error-403.svg'
+        },
+        {
+            status: 404,
+            message: "Not found",
+            description: `This is not the content you were looking for.<br>
```
mdelez, line 44: This description can simply say "The content you were looking for cannot be found." No need for the second sentence as it is clear nothing was found.

src/app/main/error/error.component.ts
```diff
@@ -8,14 +17,90 @@ import { Title } from '@angular/platform-browser';
...
+            But we couldn't find anything with this request.`,
+            action: 'goback',
+            image: 'dsp-error-404.svg'
+        },
+        {
+            status: 500,
+            message: "Internal Server Error",
+            description: `The DaSCH Service Platform is not available at the moment.<br>
+            An error has occured in a server side script, a no more specific message is suitable.`,
```
mdelez, line 53: you can remove "a no more specific message is suitable". The correct way to say it would be "a more specific message is not suitable" but I think it's clear enough without this part.

mdelez (changes requested):
Wording improvements :)

src/app/main/error/error.component.ts
kilchenmann: Done in 686f1ce

src/app/main/error/error.component.ts
kilchenmann: Done in 686f1ce

src/app/main/error/error.component.ts
kilchenmann: Done in 686f1ce

kilchenmann:
It seems that the tests failing from time to time. The problem is the dialog box:
`Error: Found the synthetic listener @dialogContainer.start. Please include either "BrowserAnimationsModule" or "NoopAnimationsModule" in your application.` I'll add the suggested module now.

mpro7:
@kilchenmann did adding the module in a49b63f help?

mpro7 (approved):

mdelez (approved):

PR #357, opened 2021-01-11
feat (project landing page): download metadata in JSON format (DSP-1184)

waychal:
resolves DSP-1184

It adds the functionality on project landing page to download metadata in JSON format.


flavens (commented):
why did you push your package-lock?

waychal:
> why did you push your package-lock?

I accidentally added it. I have reverted back the changes.

src/app/project/board/board.component.ts
```diff
@@ -262,6 +262,22 @@ export class BoardComponent implements OnInit {
         this.keywords = this.projectMetadata['keywords'];
     }
 
+    // download metadata
+    downloadMetadata() {
```
flavens, line 266: question: will you use this method only for JSON or as a generic one for the download format?

waychal: Once the endpoints for other formats will be available, this method will be converted into generic one where format will be passed as a parameter.

src/app/project/board/board.component.ts
```diff
@@ -262,6 +262,22 @@ export class BoardComponent implements OnInit {
...
+        const blob: Blob = new Blob([JSON.stringify(this.metadata)], {type: 'application/json'});
+        const fileName: string = 'metadata.json';
+        const objectUrl: string = URL.createObjectURL(blob);
+        const a: HTMLAnchorElement = document.createElement('a') as HTMLAnchorElement;
```
mpro7, line 270: You don't need to annotate type if the value of properties it's already initialised. But I'm not sure which convention we follow in DSP-APP. It's generally not bad, but redundant.

flavens (approved):

src/app/project/board/board.component.ts
mpro7: I think there will be no separate endpoint for downloads, but just different parameter to pass to the GET method. But this needs to be confirmed with BE team.

src/app/project/board/board.component.ts
waychal: I also don't know. May be @flavens can comment it and I will use it next time.

src/app/project/board/board.component.ts
waychal: We should create issue for adding endpoints for different formats in backend and prioritise it.

PR #361, opened 2021-01-15
feat(file value uploads): use upload from for file value types

tobiasschweizer:
resolves DSP-1203


tobiasschweizer:
@flavens I integrated the upload form. However, now we need to figure out how to get the tmp. file name from Sipi and then create the `CreateStillImageFileValue` class. I guess we would need some help from @mpro7.

src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
```diff
@@ -372,4 +378,34 @@ export class ResourceInstanceFormComponent implements OnInit, OnDestroy {
...
+    private _hasBinaryRepresentation(properties: ResourcePropertyDefinition[]): string[] {
+
+        const binaryTypes = [Constants.StillImageFileValue, Constants.AudioFileValue, Constants.DocumentFileValue, Constants.MovingImageFileValue];
+
+        // remove file value properties from property list
+        this.properties = properties.filter(
+            prop => !binaryTypes.includes(prop.objectType)
+        );
+
+        // TODO: add a FormControl representing the file value to propertiesParentForm
```
tobiasschweizer, line 398: @flavens @mpro7 I added [Outputs to the upload file component](https://github.com/dasch-swiss/dsp-ui-lib/pull/264).

My idea is that on a successful file upload, a `CreateStillImageFileValue` can be created. However, I am not sure where this should happen. Would `SelectPropertiesComponent` be the right place to handle this?

I have added some logic for the upload to `ResourceInstanceFormComponent`, but maybe this should be moved to  `SelectPropertiesComponent`.

tobiasschweizer: Or maybe instead of the Outputs,  `UploadFileComponent`'s form could be used: as long as there is no successful upload to Sipi, the form is invalid.  `UploadFileComponent` would have to be added to `propertiesParentForm` then.

But then instead of the binary file (`File`), the `UploadedFileResponse` obtained from Sipi should be stored in the `FormControl`.

mpro7: @tobiasschweizer IMO FormControl has to be used anyway, so second options sounds good for me.

tobiasschweizer: ok, I'll try that approach next week.

flavens:
@tobiasschweizer please do not update this branch from main (we will explain during the daily scrum)

tobiasschweizer:
> @tobiasschweizer please do not update this branch from main (we will explain during the daily scrum)

I figure it is because you removed everything but the admin part from main.

kilchenmann:
This is done in #472

PR #363, opened 2021-01-19
feat(select-resource-class): allow accented character (DSP-1241)

flavens:
Resolves DSP-1241


mdelez (commented):
It might be better to use ranges instead of hardcoded values here unless we only want support for german and french in which case you might also want to add ß. https://stackoverflow.com/a/26900132
https://en.wikipedia.org/wiki/List_of_Unicode_characters#Basic_Latin

mdelez (approved):
Looks good to me! One thing to note is that this allows for a few math symbols (÷, ×). Not sure if it's a big issue though.

flavens:
> Looks good to me! One thing to note is that this allows for a few math symbols (÷, ×). Not sure if it's a big issue though.

I do not think so. I am not sure these symbols will ever be used in labels.

PR #370, opened 2021-01-27
refactor!: set up the login page as a starting page (DSP-1292)

flavens:
resolves DSP-1292

### DIFF
--- a/src/app/main/login/login.component.html
+++ b/src/app/main/login/login.component.html
@@ -1,3 +1,4 @@
 <div class="login-page">
-    <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form>
+    <dsp-login-form (loginSuccess)="refresh()"></dsp-login-form>
+   <!--  <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form> -->
 </div>
--- a/src/app/main/login/login.component.html
+++ b/src/app/main/login/login.component.html
@@ -1,4 +1,4 @@
 <div class="login-page">
     <dsp-login-form (loginSuccess)="refresh()"></dsp-login-form>
-   <!--  <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form> -->
+   <!-- will be reactivated: <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form> -->
 </div>
--- a/src/app/main/login/login.component.html
+++ b/src/app/main/login/login.component.html
@@ -1,4 +1,3 @@
 <div class="login-page">
-    <dsp-login-form (loginSuccess)="refresh()"></dsp-login-form>
-   <!-- will be reactivated: <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form> -->
+    <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form>
 </div>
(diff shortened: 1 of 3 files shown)


src/app/main/login/login.component.html
```diff
@@ -1,3 +1,4 @@
 <div class="login-page">
-    <dsp-login-form (loginSuccess)="refresh($event)"></dsp-login-form>
+    <dsp-login-form (loginSuccess)="refresh()"></dsp-login-form>
```
kilchenmann, line 2: The `LoginComponent` emits a boolean value in `loginSuccess`. I do not understand why you removed it?

src/app/main/login/login.component.ts
```diff
@@ -26,13 +28,21 @@ export class LoginComponent implements OnInit {
     ngOnInit() {
     }
 
-    refresh(status: boolean) {
+    refresh() {
+        // check if a session is active
+        if (this._session.getSession()) {
+            this._router.navigate(['dashboard']);
+        }
```
kilchenmann, line 35: After successful login the user will always be redirected to the dashboard page, right? 
Why not keeping `this._router.navigate([this.returnUrl]);`? In case a user bookmarked his project page e.g. `admin.dasch.swiss/project/1111/info` but (s)he's no longer logged-in, then it would be nice to be redirected to this page after logging-in. 
But it's just a detail... maybe not important 😉

src/app/main/login/login.component.ts
```diff
@@ -26,13 +28,21 @@ export class LoginComponent implements OnInit {
     ngOnInit() {
     }
 
-    refresh(status: boolean) {
+    refresh() {
+        // check if a session is active
+        if (this._session.getSession()) {
```
kilchenmann, line 33: could be `if(status) {`, because you get the status from login-form

src/app/main/login/login.component.ts
flavens: this is why I redirect explicitly to the dashboard, after logging in, I get this page (which I do not want):
Screenshot 2021-01-27 at 15 56 58

flavens:
@kilchenmann could you explain: `this._componentCommsService.emit(new EmitEvent(Events.LoginSuccess, true));` from the refresh method?

src/app/main/login/login.component.ts
kilchenmann: haha...yes true. The root path `/` was handled in the main component before. This component showed the landing page or the dashboard. Yes, as I said "it's just a little detail". I can live with it.
Another solution would have been to integrate the login page instead of the landing page. I.e. that still the main component handles the redirection. 
But it's fine 😉

kilchenmann:
> @kilchenmann could you explain: `this._componentCommsService.emit(new EmitEvent(Events.LoginSuccess, true));` from the refresh method?

I think this is what @mdelez implemented. It resolved DSP-331

src/app/main/login/login.component.html
flavens: reversed in 27bc05c

src/app/main/login/login.component.ts
flavens: refactored in 34e4acf

kilchenmann (approved):

PR #378, opened 2021-02-09
feat(list-editor): add deletion functionality (DSP-1334)

mdelez:
resolves DSP-1334


mdelez:
@flavens this PR actually only deals with deleting a child node and not a root node (the entire list itself). Do you think I should also do that in this PR or in a separate one?

flavens:
> @flavens this PR actually only deals with deleting a child node and not a root node (the entire list itself). Do you think I should also do that in this PR or in a separate one?

You can do it in a separate one.

flavens:
@mdelez so we cannot delete the root node but also the last or only child node, right?

flavens (approved):

mdelez:
> @mdelez so we cannot delete the root node but also the last or only child node, right? is it the correct current behaviour?

You should be able to delete the last or only child node as of [DSP-API 13.3.1](https://github.com/dasch-swiss/dsp-api/pull/1817)

PR #382, opened 2021-02-15
feat(ontology): default language for property label

kilchenmann:
resolves DSP-1346

In this PR I implemented a select form element to select one default language for the property label.


flavens (changes requested):
On my screen, the field label is cropped.. It may be too long as well.

Screenshot 2021-02-16 at 16 48 04

And when I update the corresponding props in English first and then I open again the edit form and change the language to Deutsch. How do I know in which language is my props, is there a mix? If a user wants to enter the props in 2 languages, can s.he do that? If yes, it is not clear to me.

kilchenmann:
> On my screen, the field label is cropped.. It may be too long as well.

What do you mean "it's cropped"?

> 
> And when I update the corresponding props in English first and then I open again the edit form and change the language to Deutsch. How do I know in which language is my props, is there a mix? If a user wants to enter the props in 2 languages, can s.he do that? If yes, it is not clear to me.

Yes you're right it's not clear. It's a first try and has to be well documented, but it has to be improved too. In this first version of ontology editor you have to select one language for all properties. You can add more labels in other languages later in the edit property part (this is another branch I'm working on). But for now this is a fast solution to create properties with a language.

kilchenmann:
In my version it's not cropped

Screen Shot 2021-02-16 at 17 09 42

kilchenmann:
> On my screen, the field label is cropped.. It may be too long as well.

Shortened the placeholder in e5b5a94

flavens (approved):
The language selector issue should be improved quite soon because even disabled, I am not that a user will understand the situation...

kilchenmann:
> The language selector issue should be improved quite soon because even disabled, I am not that a user will understand the situation...

I totally agree and I'm open for better solutions!

Thanks for approving 🎉

PR #386, opened 2021-02-16
feat(list-editor): delete list root node (DSP-1356)

mdelez:
resolves DSP-1356

### DIFF
--- a/src/app/project/list/list.component.html
+++ b/src/app/project/list/list.component.html
@@ -70,6 +70,10 @@
                     (click)="$event.stopPropagation(); openDialog('editListInfo', list.labels[0].value, list.id)">
                     Edit info
                 </button>
+                <button mat-stroked-button class="delete"
+                    (click)="$event.stopPropagation(); openDialog('deleteList', list.labels[0].value, list.id)">
+                    Delete list
+                </button>
             </mat-toolbar-row>
         </mat-toolbar>
 
--- a/src/app/project/list/list.component.html
+++ b/src/app/project/list/list.component.html
@@ -66,13 +66,15 @@
                 <p class="mat-body" [matTooltip]="list.comments | dspStringifyStringLiteral" matTooltipPosition="above">
                     {{list.comments | dspStringifyStringLiteral | dspTruncate:64}}</p>
                 <span class="fill-remaining-space"></span>
-                <button mat-stroked-button
-                    (click)="$event.stopPropagation(); openDialog('editListInfo', list.labels[0].value, list.id)">
-                    Edit info
+                <button mat-button
+                    (click)="$event.stopPropagation(); openDialog('editListInfo', list.labels[0].value, list.id)"
+                    matTooltip="Edit info">
+                    <mat-icon>edit</mat-icon>
                 </button>
-                <button mat-stroked-button class="delete"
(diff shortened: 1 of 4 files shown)


mdelez:
@kilchenmann @flavens This PR adds this button to delete a list root node
Screen Shot 2021-02-17 at 12 44 23

kilchenmann:
Looks good, but just a comment: In the ontology editor we don't use icons on the root level. There we have two buttons to edit and delete the whole ontology.

Screen Shot 2021-02-17 at 13 58 54

In the first version of list editor I also had a button with the label "Edit". We should keep it consistent...

mdelez:
> In the first version of list editor I also had a button with the label "Edit". We should keep it consistent...
> 
> It makes sense to have icons instead words when we do not have the space for a big button. But on this level we could have buttons with labels instead icons.

I agree on keeping it consistent. I kinda like the icons as opposed to text since we have these icons in a few places now so the user should be familiar with them. But like you said, at this level we have space for buttons with text. I think using icons could also come to our advantage when dealing with multiple languages. Since the icon will always be the same, the way the button looks will always be the same. For example, you don't have to worry about how text in every provided language will look in regards to the CSS applied to it. Having said this, idk how easy it is to provide tooltips in multiple languages.

@flavens any opinion? If not, I'll change the buttons back to text.

kilchenmann:
> > In the first version of list editor I also had a button with the label "Edit". We should keep it consistent...
> > It makes sense to have icons instead words when we do not have the space for a big button. But on this level we could have buttons with labels instead icons.
> 
> I agree on keeping it consistent. I kinda like the icons as opposed to text since we have these icons in a few places now so the user should be familiar with them. But like you said, at this level we have space for buttons with text. I think using icons could also come to our advantage when dealing with multiple languages. Since the icon will always be the same, the way the button looks will always be the same. For example, you don't have to worry about how text in every provided language will look in regards to the CSS applied to it. Having said this, idk how easy it is to provide tooltips in multiple languages.
> 
> @flavens any opinion? If not, I'll change the buttons back to text.

I totally agree what you said but an icons will always need a text label or at least a tooltip. There are several articles about this topic e.g. <https://ux.stackexchange.com/questions/1795/when-to-use-icons-vs-icons-with-text-vs-just-text-links>

Anyway it's something we have to decide when do the big refactor/redesign task. This is something we have to mockup with Figma. For now we should have the same setup in all admin views. I can also convert the edit- / delete-buttons into icons in the ontology editor.

flavens:
@kilchenmann @mdelez Mike has a good point, with icons (and there are very common so easy to understand), we won't bother to translate the button label. Because I do not know when the interface translation will be planned in the roadmap, probably not soon. You have to stay consistent between the 2 UI view.

kilchenmann:
So, someone has to decide which type we should use 😅

kilchenmann:
I suggest to keep the icons here. I will update the view in the ontology editor. Okay?

kilchenmann (approved):

mdelez:
> I suggest to keep the icons here. I will update the view in the ontology editor. Okay?

Okay :) I think this decision will actually have to come from Ivan, Lukas, or the Client Service Team.

kilchenmann:
Btw. I like the combination of the buttons in youtrack:

Screen Shot 2021-02-17 at 14 59 31

PR #387, opened 2021-02-17
feat(permission): set simple permissions

flavens:
resolves DSP-1355

### DIFF
--- a/src/app/app-global.ts
+++ b/src/app/app-global.ts
@@ -19,8 +19,8 @@ export class AppGlobal {
             icon: 'group'
         },
         {
-            label: 'Permission groups',
-            shortLabel: 'Groups',
+            label: 'Permissions',
+            shortLabel: 'Permissions',
             route: 'permissions',
             icon: 'lock_open'
         },
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -91,6 +91,7 @@ import { AddressTemplateComponent } from './project/board/address-template/addre
 import { OrganisationTemplateComponent } from './project/board/organisation-template/organisation-template.component';
 import { EditListItemComponent } from './project/list/list-item-form/edit-list-item/edit-list-item.component';
 import { PropertyInfoComponent } from './project/ontology/property-info/property-info.component';
+import { DisplayPermissionComponent } from './project/permission/display-permission/display-permission.component';
 
 // translate: AoT requires an exported function for factories
 export function HttpLoaderFactory(httpClient: HttpClient) {
@@ -164,7 +165,8 @@ export function HttpLoaderFactory(httpClient: HttpClient) {
         AddressTemplateComponent,
         OrganisationTemplateComponent,
         EditListItemComponent,
-        PropertyInfoComponent
+        PropertyInfoComponent,
+        DisplayPermissionComponent
     ],
     imports: [
         AppRoutingModule,
--- a/src/app/project/permission/display-permission/display-permission.component.html
+++ b/src/app/project/permission/display-permission/display-permission.component.html
@@ -1,8 +1,8 @@
 <!-- display all granted permissions -->
 <td>{{role}}</td>
-<td><input type="checkbox" [checked]="grantedPermissions[0] !== undefined"></td>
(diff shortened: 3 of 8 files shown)


tobiasschweizer:
coordinated requests:

```ts
                const calls = [
                    this._dspApiConnection.admin.permissionsEndpoint.getAdministrativePermissions(this.project.id).pipe(
                        map((res: ApiResponseData<AdministrativePermissionsResponse>) => res.body)
                    ),
                    this._dspApiConnection.admin.permissionsEndpoint.getDefaultObjectAccessPermissions(this.project.id).pipe(
                        map((res: ApiResponseData<DefaultObjectAccessPermissionsResponse>) => res.body)
                    )
                ];

                forkJoin(calls).subscribe(
                    res => console.log(res)
                );
```

tobiasschweizer:
@flavens I added the comp. `display-permission`.
Given a permission string, it displays all granted permissions:

```ts
<app-display-permission [displayPermission]="'V'"></app-display-permission>
```

Screenshot 2021-02-25 at 15 31 47


So far it uses checkboxes and it does not display any title info. Could you help me turn this into a grid?

flavens:
> So far it uses checkboxes and it does not display any title info. Could you help me turn this into a grid?

Yes I can help. I am trying to get the same display, but I get only one set of permissions (V passed as displayPermission). I may need to understand what you did first.

tobiasschweizer:
I just got the DOAP's permissions and put them in a variable: `perms: Permission[];`.

And then I juts loop over them: 

```ts
        <table>
            <tr *ngFor="let perm of perms"><th>{{perm.additionalInformation}}</th><td><app-display-permission [displayPermission]="perm.name"></app-display-permission></td></tr>
        </table>
```

So for each row,  `app-display-permission` can be used.

tobiasschweizer:
https://github.com/dasch-swiss/dsp-app/pull/387/commits/7e719862866dc4ece4546d6ef6f6730ba1365055 uses the child comp. for DOAP

Screenshot 2021-02-25 at 17 16 08

tobiasschweizer:
After today's discussion: we learned that the create permission is **not** necessarily implicitly granted when the user has modify or delete permissions.

We initially had the idea to merge administrative and default object access permissions into one data structure (combining the results of `getDefaultObjectAccessPermissions` and `getAdministrativePermissions`) that then would be displayed. However, now it seems favourable to keep the two types of permissions separate internally, but to display them side by side in the UI. The user need not know that there are different types of permissions.

The component displaying permissions could have separate inputs:

```ts
<app-display-permission [role]="..." [doaPermission]="..." [adminPermission]="...">
```

I think, both types of permissions could fit into simple data structure like this:

```ts
export interface IPermission {
    role: string;
    permission:string;
}
```

Per group and role, there could be permissions belonging to admin and such belonging to doap. They should be kept separate and passed to the respective inputs.

Example:

```
AdministrativePermission
   forGroup: "http://www.knora.org/ontology/knora-admin#ProjectMember"
   forProject: "http://rdfh.ch/projects/0001"
   hasPermissions: Array(1)
       0: Permission
           additionalInformation: undefined
           name: "ProjectResourceCreateAllPermission"
           permissionCode: undefined
   id: "http://rdfh.ch/permissions/0001/005-a1"

DefaultObjectAccessPermission
   forGroup: "http://www.knora.org/ontology/knora-admin#ProjectMember"
   forProject: "http://rdfh.ch/projects/0001"
   forProperty: undefined
   forResourceClass: undefined
   hasPermissions: Array(4)
      0: Permission {additionalInformation: "http://www.knora.org/ontology/knora-admin#Creator", name: "CR", permissionCode: 8}
      1: Permission {additionalInformation: "http://www.knora.org/ontology/knora-admin#ProjectMember", name: "M", permissionCode: 6}
      2: Permission {additionalInformation: "http://www.knora.org/ontology/knora-admin#KnownUser", name: "V", permissionCode: 2}
      3: Permission {additionalInformation: "http://www.knora.org/ontology/knora-admin#UnknownUser", name: "RV",    permissionCode: 1}
   id: "http://rdfh.ch/permissions/0001/005-d1"

```

tobiasschweizer:
@flavens Could we make a mockup that shows how the two sets of permissions in the example above would be shown to the user?

PR #388, opened 2021-02-17
feat(list-editor): reposition a child node amongst its siblings (DSP-1340)

mdelez:
resolves DSP-1340


mdelez:
I discovered I was using quite an old version of npm which was creating conflicting package-lock.json files. Should be fixed now.

mdelez:
@kilchenmann This PR adds these two buttons to the "action bubble" and the up and down arrows hide if the node cannot move in that direction. I did not implement it as drag and drop for version 1. [Explained in more detail here](https://dasch.myjetbrains.com/youtrack/issue/DSP-1340).
Screen Shot 2021-02-18 at 13 35 26

src/app/project/list/list-item-form/list-item-form.component.html
```diff
@@ -23,6 +23,20 @@
...
+            <button mat-button
+                    *ngIf="position > 0"
+                    class="reposition up"
+                    title="move up"
+                    (click)="$event.stopPropagation(); repositionNode('up')">
+                <mat-icon>arrow_upward</mat-icon>
+            </button>
+            <button mat-button
+                    *ngIf="!lastPosition"
+                    class="reposition down"
```
waychal, line 35: According to the conventions written [here](https://docs.google.com/document/d/1SIVMxoV0N8xYqhVk_0hhuqSuUF47D1YYrDESqdmipfM/edit#heading=h.9a4ce1jq4p7q) class should be at the end.

waychal (changes requested):
Everything else look ok to me.

src/app/project/list/list-item-form/list-item-form.component.html
mdelez: done in 4f8909d

waychal (approved):

PR #391, opened 2021-02-22
feat(ontology): separate list of ontology properties (DSP-1364)

kilchenmann:
resolves DSP-1364

This is a first version of a separate list of properties. In another task we will add a filter and sort button.

### DIFF
--- a/src/app/project/ontology/ontology.component.scss
+++ b/src/app/project/ontology/ontology.component.scss
@@ -27,7 +27,6 @@ $width: 304px;
   max-width: 1280px;
   min-height: calc(100vh - #{$header-height} - #{$tab-bar-margin} - 100px);
   border: 1px dotted $primary_200;
-  background: $primary_200;
   position: relative;
   margin: 0 auto 60px auto;
 }
@@ -45,8 +44,13 @@ $width: 304px;
     height: $sub-header-height;
   }
 
+  .ontology-editor-sidenav {
+    width: 160px;
+  }
+
   .ontology-editor-canvas {
     min-height: calc(100vh - #{$sub-header-height} - #{$header-height} - #{$tab-bar-margin} - 100px);
+    background: $primary_200;
 
     .ontology-editor-grid {
       display: grid;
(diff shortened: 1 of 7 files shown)


src/app/project/ontology/ontology.component.scss
```diff
@@ -45,15 +44,29 @@ $width: 340px;
...
+
+    button {
+      width: 100%;
+      text-align: left;
+    }
+  }
+
   .ontology-editor-canvas {
     min-height: calc(100vh - #{$sub-header-height} - #{$header-height} - #{$tab-bar-margin} - 100px);
+    // background: $primary_200;
```
flavens, line 58: when there are unused css classes, we should clean them up or add a comment (remark valid for all the scss files)

flavens (approved):

src/app/project/ontology/ontology.component.scss
kilchenmann: Some of the css classes are not defined in the scss file, that's true. But they are needed for the tests.

Or are you talking about the commented css property here? I can clean them up...

kilchenmann: Done in ac7c3cf

mdelez:
Screen Shot 2021-02-24 at 12 23 37

The numbering seems to be a little messed up 🤔

kilchenmann:
> Screen Shot 2021-02-24 at 12 23 37
> 
> The numbering seems to be a little messed up 🤔

Yes I know. I'm using the number from `gui order`. This Gui order is defined in the ontology; it's part of the res class cardinality. In case of missing or doubled numbers it's a mistake in the ontology. The test data are not well done.

mdelez (approved):
I would stress that we need to have some unit tests for the ontology component but I don't think this is the PR for that so I'll approve this one :)

kilchenmann:
> I would stress that we need to have some unit tests for the ontology component but I don't think this is the PR for that so I'll approve this one :)

I'm working on it 😉

PR #394, opened 2021-02-24
feat(eslint): migrate tslint to eslint (DSP-1372)

kilchenmann:
resolves DSP-1372

The most important file that has to be reviewed in this PR is the [.eslintrc.js (eslint config file)](https://github.com/dasch-swiss/dsp-app/blob/b68a4b2994eb761ae9bf884e7b4e15cc51d6c507/.eslintrc.js).


src/app/main/cache/cache.service.ts
```diff
@@ -12,21 +12,21 @@ interface CacheContent {
...
 })
 /**
- * Cache Service is an observables based in-memory cache implementation
+ * cache Service is an observables based in-memory cache implementation
  * Keeps track of in-flight observables and sets a default expiry for cached values
  * @export
  * @class CacheService
  */
 export class CacheService {
+    readonly defaultMaxAge: number = 3600000;  // 3600000ms => 1 Stunde
```
mdelez, line 21: should we keep readonly class members as all caps with underscore? I'm quite used to that formatting.

mdelez (commented):
Nifty conversion tool! Just a small question.

src/app/main/cache/cache.service.ts
kilchenmann: I think we decided to use underscore in private variables only. 
The ESLint rule for variable names is always **camelCase**. 
**PascalCase** is not allowed or we have to add a specific rule...

kilchenmann: @mdelez Do you mean in this case it should be like

```typescript
readonly _DefaultMaxAge: number
```

OR

```typescript
readonly _defaultMaxAge: number
```

kilchenmann: This could be interesting:
<https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md>

mpro7:
In my opinion we should agree on the rules first, which can be later applied.

kilchenmann:
> In my opinion we should agree on the rules first, which can be later applied.

I agree. We can use the same rules in DSP-APP as defined in DSP-UI-Lib. We should expand the existing list of conventions: <https://docs.google.com/document/d/1SIVMxoV0N8xYqhVk_0hhuqSuUF47D1YYrDESqdmipfM/edit#heading=h.5fm7povww30c>

src/app/main/cache/cache.service.ts
mdelez: I'm talking about UPPER_CASE for `readonly` members. The migration tool changed `DEFAULT_MAX_AGE` to `defaultMaxAge`. I agree on keeping the leading underscore for private variables. Apparently this a source of contention within the JS community however.

kilchenmann: 🙈 Yes, that is what I have now also seen and understood 🤪

kilchenmann: I added some naming convention options to the eslint configuration (932ddcf). Now it should support our naming conventions from ui-lib design document.

mdelez (approved):
Looks good!

src/app/main/directive/invalid-control-scroll.directive.spec.ts
```diff
@@ -37,7 +37,7 @@ class TestLinkHostComponent implements OnInit {
     }
 
     onSubmit() {
-        console.log('form submitted');
+        console.warn('form submitted');
```
mpro7, line 40: Does it make difference have it as a `warn` instead of `log`?

.eslintrc.js
```diff
@@ -156,7 +156,6 @@ module.exports = {
             "error",
             {
                 "allow": [
```
mpro7, line 158: why?

src/app/main/directive/invalid-control-scroll.directive.spec.ts
kilchenmann: This is the question: https://docs.google.com/document/d/1SIVMxoV0N8xYqhVk_0hhuqSuUF47D1YYrDESqdmipfM/edit?disco=AAAALqCScG4

mpro7: I see, but it doesn't answer my question ;)
Why you exclude error and log but keep the others @flavens ?

Different console options can have different usages. In the example above we get warn the code is submitted? It should be rather info or log :)

All console methods can be divided in some groups. The group with `log info warn error` is to display messages. The only one difference here are is visual, which should help to identify errors quicker. The common practice is to use `log`. And I think it's fine, unless there are some sophisticated rules to use all others (which is always nice to have ;).

Anyways, I find this linting rule just a matter of preferences, because at the end, no console logs, warns, errors, should be on the production (there is a flag to disable these). But strange would be to get most commonly used 'log' and 'error' w/o having the rules set for the others. Just my 2 Rappens ;)

kilchenmann: @mpro7 Thanks for the feedback. Would you like to copy this comment to the DSP-UI design documentation on google drive? Because we agreed last Friday to have the discussion there. 
About "no console.log, console.warn and console.error": it was just an idea. And as you said we all have to agree to the conventions. I can revert the rule. No problem. But there was no other feedback in the design documentation than the one from @flavens.

kilchenmann: Okay.... I misunderstood the comment from @flavens 
I will fix it. I'm sorry 🙈

mpro7: Ok, I will.

kilchenmann: This is now the "correct" (has to be discussed) solution for `console.log, .error and .warn`: 4f3d7cc and fb7ded2
I'm sorry again!

kilchenmann: > Ok, I will but this is why I wanted to have rather some ideas rose on the chat, where we can get notification. I' busy with other things and don't have much time to go through all rules, but I can always add my 2 Rappens :D

Yes, I updated the design documentation, added my comment and then I opened the new thread in the R&D chat room. But nobody answered in the chat; only Flavie added a comment to the design documentation. So, how we can we all agree to the rules (now)?

flavens: if all thre rules have been defined in the eslint file in this PR, each approval = agreement

mpro7: @kilchenmann did you go through all the rules and if, did you find anything which would overwhelming?

kilchenmann: I went through most of the options and looked for inconsistencies. In this PR `eslintrc.js` was created from `tslint.json` where we had already most of the rules configured. ESLint is just more restrict and this is why we have so many changes in this PR.

.eslintrc.js
```diff
@@ -0,0 +1,205 @@
...
+        "id-blacklist": "off",
+        "id-match": "off",
+        "import/no-deprecated": "warn",
+        "jsdoc/no-types": "off",
+        "max-len": [
+            "error",
+            {
+                "code": 200
+            }
+        ],
```
mpro7, line 152: I don't how how the whole translation process looks like but it seems there is some bug here. It should be `max-lenght`. So was thats manually copied by you @kilchenmann or this translation tool bug?

Another question here is, if that rule, and maybe others here (which I can recognise from old TSLing) are applied, because this one for instance isn't in the rules directory:
https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin/docs/rules

src/app/main/directive/invalid-control-scroll.directive.spec.ts
mpro7: That's fine. Have a look on [my newest comment](https://github.com/dasch-swiss/dsp-app/pull/394#discussion_r585524067).

.eslintrc.js
kilchenmann: No, I think `max-len` is correct? <https://eslint.org/docs/rules/max-len>

kilchenmann: This was the rule in `tslint.json`:

```json
"max-line-length": [
    true,
    200
],
```

mpro7: Ok, thanks, then it's me this time being blind ;)

.eslintrc.js
```diff
@@ -0,0 +1,205 @@
...
+
+Happy linting! 💖
+*/
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
+        "node": true
+    },
+    "ignorePatterns": [
```
flavens, line 20: should we also add ".yalc"?

kilchenmann: Good point. Added it to `.eslintignore` in 8bfa437

flavens (approved):

kilchenmann:
For the reviewers: I just added new rules about spaces in objects which is defined here:
https://docs.google.com/document/d/1SIVMxoV0N8xYqhVk_0hhuqSuUF47D1YYrDESqdmipfM/edit#heading=h.awf6ef8h98ze

PR #397, opened 2021-02-25
fix(ontology): set the cache earlier in case of only one ontology (DSP-1374)

kilchenmann:
resolves DSP-1374


mdelez (approved):

kilchenmann:
Thank you @mdelez for approving. But @flavens figured out that with some ontologies we get some errors in the developer tools' console. I have to refactor the cache concept in the `ontology.component`.

kilchenmann:
@flavens @waychal @mdelez I finished this task. It would be nice if someone can do the review (again). Thanks

src/app/project/ontology/ontology.component.ts
```diff
@@ -193,36 +192,91 @@ export class OntologyComponent implements OnInit {
...
+                    this._errorHandler.showMessage(error);
+                }
+            );
             await callback(ontologies[i]);
         }
     }
 
+    async loadAndCache(ontologies: OntologyMetadata[]) {
+        await this.asyncForEach(ontologies, async (onto: OntologyMetadata) => {
+            await this.waitFor(200);
```
mdelez, line 253: Can you clarify why this method is needed? I'm not a huge fan of `waitFor` and we have stated in our [design doc](https://docs.google.com/document/d/1SIVMxoV0N8xYqhVk_0hhuqSuUF47D1YYrDESqdmipfM/edit#heading=h.jjkobsuiogsh) that we shouldn't use setTimeout in production code. How will this behave with a rather slow internet connection?

src/app/project/ontology/ontology.component.ts
```diff
@@ -511,8 +507,7 @@ export class OntologyComponent implements OnInit {
 
     setCache() {
         // set cache for current ontology
-        this._cache.set('currentOntology', this.ontology);
-        this._cache.set('currentProjectOntologies', this.ontologies);
+        // this._cache.set('currentOntology', this.ontology);
```
mdelez, line 510: is this needed?

mdelez (changes requested):

src/app/project/ontology/ontology.component.ts
kilchenmann: no, I forgot to delete it

src/app/project/ontology/ontology.component.ts
waychal: Whats happens if wait is over but data is still not ready?

kilchenmann: Yes I know, that we have this rule in our conventions. But I have no idea how to solve this async requests in another way. I found this solution in a tutorial. Maybe someone of you has a better idea?

mdelez: I'll pull the changes from main and check it out

waychal: One way is:
1. set Boolean variable to false
2. once your data is ready, set this variable to true
3. add watch on the boolean variable defined above and so when it becomes true, call the function to want run after the Ajax data is ready

kilchenmann: I'll try that...

src/app/project/ontology/ontology.component.ts
kilchenmann: done in 0f6438d

src/app/project/ontology/ontology.component.ts
mdelez: idk if you've already found this @kilchenmann but the issue isn't that the data isn't arriving in time from the API, it's that `this.ontologies` isn't modified in time for the length check within `loadAndCache` so its length is always 0.

Without `waitFor`:
Screen Shot 2021-02-26 at 16 44 37

With `waitFor`:
Screen Shot 2021-02-26 at 16 47 18

I think it's due to your callback. Admittedly I'm quite horrible at dealing with async/await things but I'm quite certain it's an issue with your callback getting triggered earlier than it should.

kilchenmann: Thank you @mdelez 
I'm open for better solution than async/await. At the moment I'm lost. Even with the solution from @waychal I'm not successful. I'm pretty sure I'm doing something wrong, but I have no idea what and how to solve it 😞

kilchenmann: ok, I came up with another solution without async/await. See my last commit (d174ba2).

src/app/project/ontology/ontology.component.ts
```diff
@@ -375,7 +362,7 @@ export class OntologyComponent implements OnInit {
     openResourceClassForm(mode: 'createResourceClass' | 'editResourceClass', resClassInfo: DefaultClass): void {
 
         // set cache for ontology and lists
-        this.setCache();
+        // this.setCache();
```
mdelez, line 365: Can this be removed? Same with the other similar commented out line below

flavens (approved):
in general, it looks good. I have just spotted one error message when I open the Leibniz onotology (and only this one):
Screenshot 2021-03-01 at 13 36 26
The error may come from somewhere else than the cache or your code (?!).

src/app/project/ontology/ontology.component.ts
kilchenmann: done in 68380d1

kilchenmann:
> in general, it looks good. I have just spotted one error message when I open the Leibniz onotology (and only this one):
> Screenshot 2021-03-01 at 13 36 26
> The error may come from somewhere else than the cache or your code (?!).

@flavens no, you're right. It was a logical issue. Resolved in 2be2a49

mdelez (approved):
Looks good to me!

waychal (approved):
In file `src/app/project/ontology/ontology.component.html` line 45: according to DSP-APP code conventions class attribute should be at the end.

Everything else looks fine to me.

kilchenmann:
> In file `src/app/project/ontology/ontology.component.html` line 45: according to DSP-APP code conventions class attribute should be at the end.
> 
> Everything else looks fine to me.

Thanks for the info. I think this HTML code convention is not yet fully integrated in DSP-APP. We should plan a separat HTML refactoring task for this issue.

PR #427, opened 2021-04-22
refactor(ontology): disable ontology graph view (DSP-1560)

kilchenmann:
resolves DSP-1560


flavens (approved):
N.B. I would have used the word "disable [the graph feature]" instead of "remove" in the comments

PR #442, opened 2021-05-19
fix: disable progress bar if search results are empty (DSP-1575)

waychal:
resolves DSP-1575

@flavens When page is refreshed with search string in the url, it displays search results on the page but same search string is not displayed in the search input box.

This bug requires changes in dsp-ui-lib and can be fixed in separate PR.

Screenshot 2021-05-19 at 13 25 50

### DIFF
--- a/src/app/workspace/results/results.component.html
+++ b/src/app/workspace/results/results.component.html
@@ -1,5 +1,5 @@
 <!-- In case if results present -->
-<div class="content" *ngIf="searchParams && ((!loading && (numberOfAllResults > 0)) || (searchMode === 'gravsearch'))">
+<div class="content" *ngIf="searchParams && !loading && (numberOfAllResults > 0)">
     <as-split direction="horizontal">
         <as-split-area [size]="40">
             <dsp-list-view [search]="searchParams" [displayViewSwitch]="false" (resourceSelected)="openResource($event)"></dsp-list-view>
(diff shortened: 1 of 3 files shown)


src/app/workspace/results/results.component.html
```diff
@@ -9,3 +10,15 @@
         </as-split-area>
     </as-split>
 </div>
+
+<!-- In case of 0 result -->
+<div class="no-results" *ngIf="!loading && numberOfAllResults === 0">
+    <p>Your search <span *ngIf="searchMode === 'fulltext'">- <strong> {{searchQuery}}</strong> -</span> did not match any documents.</p>
```
flavens, line 16: this message should be centered in the middle of the page instead of being displayed on the left side

flavens (changes requested):
It looks good for the fulltext search - just a little css position to change. However, I have tried the advanced search and here, nothing: no result, no message, no console log.. It may not be related but it should be investigated in case the display problem comes from this PR.

src/app/workspace/results/results.component.html
waychal: Done.

waychal:
> It looks good for the fulltext search - just a little css position to change. However, I have tried the advanced search and here, nothing: no result, no message, no console log.. It may not be related but it should be investigated in case the display problem comes from this PR.

I have fixed the issue in commit [7d5443e](https://github.com/dasch-swiss/dsp-app/pull/442/commits/7d5443e63482b9d0a2dbfa9cd2e95b850f9b74ad)

flavens (commented):
When I test the advanced search, I get the results as before but the loading spiner is running whereas you have set it to false...

https://user-images.githubusercontent.com/29532462/119500810-7c9f5f80-bd68-11eb-90b8-56481d0d387c.mov

PR #445, opened 2021-05-21
feat(resource): display region annotations in still images (DSP-1585)

kilchenmann:
resolves DSP-1585

### DIFF
--- a/src/app/material-module.ts
+++ b/src/app/material-module.ts
@@ -3,6 +3,7 @@ import { DragDropModule } from '@angular/cdk/drag-drop';
 import { NgModule } from '@angular/core';
 // material modules
 import { MatAutocompleteModule } from '@angular/material/autocomplete';
+import { MatBadgeModule } from '@angular/material/badge';
 import { MatButtonModule } from '@angular/material/button';
 import { MatButtonToggleModule } from '@angular/material/button-toggle';
 import { MatCardModule } from '@angular/material/card';
@@ -23,8 +24,8 @@ import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { MatRadioModule } from '@angular/material/radio';
 import { MatSelectModule } from '@angular/material/select';
 import { MatSidenavModule } from '@angular/material/sidenav';
-import { MatSlideToggleModule } from '@angular/material/slide-toggle';
 import { MatSliderModule } from '@angular/material/slider';
+import { MatSlideToggleModule } from '@angular/material/slide-toggle';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 import { MatSortModule } from '@angular/material/sort';
 import { MatTableModule } from '@angular/material/table';
@@ -36,6 +37,7 @@ import { MatTreeModule } from '@angular/material/tree';
 const matModules = [
     DragDropModule,
     MatAutocompleteModule,
+    MatBadgeModule,
     MatButtonModule,
     MatButtonToggleModule,
     MatCardModule,
@@ -68,9 +70,9 @@ const matModules = [
 ];
 
 @NgModule({
-    imports: [ matModules ],
+    imports: [matModules],
 
-    exports: [ matModules ]
+    exports: [matModules]
 })
 export class MaterialModule {
 }
--- a/src/app/workspace/resource/representation/still-image/still-image.component.html
+++ b/src/app/workspace/resource/representation/still-image/still-image.component.html
@@ -5,13 +5,13 @@
 
         <!-- openseadragon (osd) viewer -->
         <div class="osd-container">
-            <div class="action horizontal top">
+            <div class="action horizontal top overlay">
                 <!-- caption -->
-                <span class="overlay caption mat-caption">{{imageCaption}}</span>
+                <span class="caption mat-caption">{{imageCaption}}</span>
                 <span class="fill-remaining-space"></span>
                 <!-- image action tools e.g. zoom, rotate and flip -->
                 <!-- zoom buttons -->
-                <span class="overlay">
+                <span>
                     <button mat-icon-button id="DSP_OSD_ZOOM_OUT" matTooltip="Zoom out">
                         <mat-icon>remove_circle_outline</mat-icon>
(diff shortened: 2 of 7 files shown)


kilchenmann:
https://user-images.githubusercontent.com/4253580/119170747-fd5a1500-ba63-11eb-8b26-63916df5fd34.mov

kilchenmann:
https://user-images.githubusercontent.com/4253580/119171075-62156f80-ba64-11eb-9056-1254972bebf4.mov

Region in compound object

kilchenmann:
@flavens @mdelez @waychal if you like to try this locally you can do it with the following steps:

* Run the app with the test-server environment: `ng s --configuration=test-server` This way you have also all the images.

**1. Resource with annotation**
* Do a fulltext search with "Med_MS_p0098" (<http://localhost:4200/search/fulltext/Med_Ms_p0098>)
* You should see the regions incl. the corresponding metadata

**2. Compound object**
* Open a compound object from "Part of" property link (Meditationes) from previous object (<http://localhost:4200/resource/http:%2F%2Frdfh.ch%2F0801%2FHSmk8KWbQjy6YCqnBrwYgA>)
* Now you could scroll through the book and you should see the regions on each page.

Or you could search for other resources which have annotations. Good Luck

p.s. about the design. The design of the annotation metadata could be better but it has to be developed first in Figma.

kilchenmann:
> I have a few questions:
> 
> * why having the Annotations tab displayed in a resource without a media representation?
> * when I want to show all props of an annotation, it opens them for all the existing annotations, but I would have expected to open the complete list of props only for the selected annotation. What is the expected behaviour here? If it is not the correct one, I would like a fix in this PR.
> * about the info of the project, is it possible that 2 different annotations belong to 2 different projects? If not, I would clean up the info (after each annotation header).

1. Because it will also be used to create a new annotation. In this case we have a starting point. 
2. Yes it opens all. This is because they are handled in one component. Any suggestion how to solve it? I will also think about it but I'm happy for all constructive inputs.
3. ok, makes sense. But the line will still stay there because of the creator information. Is that correct?

kilchenmann:
@flavens about the second item: I found a solution! My fault. I will fix it...

flavens:
> 1. Because it will also be used to create a new annotation. In this case we have a starting point.

ok

> 3. ok, makes sense. But the line will still stay there because of the creator information. Is that correct? --> btw this is the point I mentioned about the "design" here [#445 (comment)](https://github.com/dasch-swiss/dsp-app/pull/445#issuecomment-847644731)

yes

flavens:
> @flavens about the second item: I found a solution! My fault. I will fix it...but it will not be a quick solution! Maybe it would be better to do it in a separate PR. Would you like to create a task on youtrack? Thanks

ok I will create the issue

waychal (commented):
I have few queries:

1. When I select annotation on image, in result it shows me the selected one in annotation tab. Is it possible to do it other way round? I would like to select the annotation in annotation tab and in result it should highlight the selected one on image.
2. When I tried to remove property from annotation, it asked me for the confirmation which is fine. In confirmation dialog, it shows value in html format (see attached screen shot). Is it a requirement or can we display value in text format?
3.  The annotation belongs to compound object is shown in red colour. When I checked annotation tab first time, it was confusing to me why 2 colours are used until I read your comment in this PR.

Screenshot 2021-05-25 at 17 12 22

kilchenmann:
> I have few queries:
> 
> 1. When I select annotation on image, in result it shows me the selected one in annotation tab. Is it possible to do it other way round? I would like to select the annotation in annotation tab and in result it should highlight the selected one on image.

Yes, that's also what I thought. But it should be done in a separate task. (s. p.s. in [comment](https://github.com/dasch-swiss/dsp-app/pull/445#issuecomment-847644731))

> 2. When I tried to remove property from annotation, it asked me for the confirmation which is fine. In confirmation dialog, it shows value in html format (see attached screen shot). Is it a requirement or can we display value in text format?

This is not part of this PR and belongs to the value component. Would you like to create an issue task on Youtrack? Thanks

> 3. The annotation belongs to compound object is shown in red colour. When I checked annotation tab first time, it was confusing to me why 2 colours are used until I read your comment in this PR.
> 

What do you mean by 2 colours? The color of the region can be selected by the creator or editor.

waychal:
> Yes, that's also what I thought. But it should be done in a separate task. (s. p.s. in [comment](https://github.com/dasch-swiss/dsp-app/pull/445#issuecomment-847644731))
> 
Ok.

> This is not part of this PR and belongs to the value component. Would you like to create an issue task on Youtrack? Thanks

Sure.

> > 1. The annotation belongs to compound object is shown in red colour. When I checked annotation tab first time, it was confusing to me why 2 colours are used until I read your comment in this PR.
> 
> What do you mean by 2 colours? The color of the region can be selected by the creator or editor. But create and edit region is not part of this PR.

Sorry, my mistake. `Farbe` in German is `Colour`. I didn't know that. I thought red is used for compound object and green for others.

Screenshot 2021-05-25 at 18 24 40

waychal:
So far everything is good from my side. If you add me as reviewer again, I will approve the PR.

gautschr (commented):
It works in both cases. After a bit of trial I figured out how to search for resources which have annotations. I also realised what was mentioned already concerning the hide/show properties option, that it doesn't work for single annotations but for all on a certain page. It took me a little bit of time to figure out which polygon corresponds to which annotation below in cases where there were e.g. three green polygons.

What I wonder about: would it be possible easily to avoid that the html color code is printed so prominently on top of the color in the case of all color values? To me it's rather distracting. 
Bildschirmfoto 2021-05-25 um 19 53 16

kilchenmann:
> What I wonder about: would it be possible easily to avoid that the html color code is printed so prominently on top of the color in the case of all color values? To me it's rather distracting.
>

I totally agree. The color value is done in the value component and so, it has to be resolved in an other task. Would you like to create an issue on youtrack? Thanks

gautschr:
Yes, I'll create a new task.

gautschr (approved):

PR #448, opened 2021-05-21
fix(resource): add if condition (DSP-1655)

kilchenmann:
resolves DSP-1655

### DIFF
--- a/src/app/workspace/resource/resource-properties/resource-properties.component.ts
+++ b/src/app/workspace/resource/resource-properties/resource-properties.component.ts
@@ -106,8 +106,10 @@ export class ResourcePropertiesComponent implements OnInit, OnDestroy {
         // subscribe to the ValueOperationEventService and listen for an event to be emitted
         this.valueOperationEventSubscriptions.push(this._valueOperationEventService.on(
             Events.ValueAdded, (newValue: AddedEventValue) => {
-                this.addValueToResource(newValue.addedValue);
-                this.hideAddValueForm();
+                if (newValue) {
+                    this.addValueToResource(newValue.addedValue);
+                    this.hideAddValueForm();
+                }
             }));
 
         this.valueOperationEventSubscriptions.push(this._valueOperationEventService.on(


kilchenmann:
@mdelez What do you think about this quick solution?

mdelez:
In what scenario do you find yourself in a situation where a `ValueAdded` event is raised but there is no value?

kilchenmann:
> In what scenario do you find yourself in a situation where a `ValueAdded` event is raised but there is no value?

only in the test... yes maybe we should fix it in the test. But I have no clue about this setup. The test failed because "newValue" is undefined

kilchenmann:
@mdelez Maybe I'll need your help here. Any suggestion?

mdelez:
> @mdelez Maybe I'll need your help here. Any suggestion?

I can take a look at it later today

kilchenmann:
Maybe @flavens has also an idea!?

mdelez:
Strangely, if I run `npm run test-ci` locally, I get a different error message. I can't get the same `ValueAdded` error 🤔 What happens when you run this command locally? No errors?

Screen Shot 2021-05-25 at 11 47 11

kilchenmann:
> Strangely, if I run `npm run test-ci` locally, I get a different error message. I can't get the same `ValueAdded` error 🤔 What happens when you run this command locally? No errors?
> 
> 

Yes, this is the strange thing. I do not have an error, but sometimes I have the same error as on Github CI. But sometimes I also had your error message, but test were successful. This is what happened right now on main branch...

kilchenmann:
On Github CI the tests failed on the mentioned error `Cannot read property 'addedValue' of undefined`, but when re-running all the tests, they are successful. But this could not be the solution, right?

mdelez:
> On Github CI the tests failed on the mentioned error `Cannot read property 'addedValue' of undefined`, but when re-running all the tests, they are successful. But this could not be the solution, right?

This is definitely a solution but I feel like we should figure out why this is actually happening at some point. My guess is that it's a timing issue. There is probably an async BeforeEach somewhere that is not ready in time sometimes. For now, I'll approve it.

mdelez (approved):

kilchenmann:
> > On Github CI the tests failed on the mentioned error `Cannot read property 'addedValue' of undefined`, but when re-running all the tests, they are successful. But this could not be the solution, right?
> 
> This is definitely a solution but I feel like we should figure out why this is actually happening at some point. My guess is that it's a timing issue. There is probably an async BeforeEach somewhere that is not ready in time sometimes. For now, I'll approve it.

Yes absolutely, I agree.

PR #449, opened 2021-05-25
docs: update documentation about contribution (DSP-1657)

kilchenmann:
resolves DSP-1657


flavens (approved):
It looks good. 
Just one comment about the link to the DSP-API doc in README.md, we end up on a 404 page... and it is not the best...

kilchenmann:
> It looks good.
> Just one comment about the link to the DSP-API doc in README.md, we end up on a 404 page... and it is not the best...

Ok, I see. This is another task, but I can fix it here. Yes, there's still the open question how we publish all the documentation at one place (DSP-1066). Since we cleaned up the dsp-docs, but we did not resolved the monorepo-documentation build, we have dead links which is really really bad.

PR #455, opened 2021-05-28
feat(ontology): new cardinality workflow (DSP-1652)

kilchenmann:
resolves DSP-1652, DSP-1656 and DSP-976

This PR contains the new workflow to create resource class and to add properties. It means: the user creates first a resource class. Then s/he has to go to this resource class and add  the properties one by one. The GUI order can be changed at any time. It's the first iteration that helps to create new resource classes and to add properties incl. the cardinality. The update of a cardinality of existing properties in a resource class will be part of the next iteration.

I had to clean up some old (or now outdated) components:
* There's no need anymore for a `resource-class-form.service` and the `resourceClassFormSub` because we have separated the cardinality from the resource-class. But I moved some of the general methods to handle the data in the ontology editor from this service to a new service called `ontology.service`. 
* The old component `resource-class-property-form.component` is not needed anymore, because we can use the `property-form.component` to handle the cardinality.


kilchenmann:
This is how it works at the moment:


Uploading Screen Recording 2021-06-04 at 11.07.44.mov…

waychal (commented):
I tried to test the functionality but my properties are not updating correctly in the resource class and for every action it gives me `Bad request(400)` error. 
am I doing something wrong here?

https://user-images.githubusercontent.com/25031092/120892982-1db3d300-c611-11eb-9cb3-3843d46c3122.mov

kilchenmann:
> I tried to test the functionality but my properties are not updating correctly in the resource class and for every action it gives me `Bad request(400)` error. 
> 
> am I doing something wrong here?
> 
> 

As mentioned already in one of the meetings, it doesn't work with the test data like the anything ontology. Because this ontology is already in use. It means that the properties have already data in it. You have to start from scratch and create your own example ontology.

mdelez:
Screen Shot 2021-06-07 at 11 08 31

For me, it works but I have to refresh the page to see any changes I've made. However the response has the correct data I believe. I was trying to add a Date of Birth (DOB) property to my class. I'm using DSP-API v13.9.2 and there are no console errors 🤔

src/app/project/ontology/property-form/property-form.component.ts
```diff
@@ -375,4 +403,45 @@ export class PropertyFormComponent implements OnInit {
...
+                this.loading = false;
+                this.closeDialog.emit();
+            },
+            (error: ApiResponseError) => {
+                this._errorHandler.showMessage(error);
+            }
+        );
+
+        this.loading = false;
+        this.closeDialog.emit();
```
kilchenmann, line 444: `this.closeDialog.emit();` should run only once namely only the one on line 436.

kilchenmann:
> For me, it works but I have to refresh the page to see any changes I've made. However the response has the correct data I believe. I was trying to add a Date of Birth (DOB) property to my class. I'm using DSP-API v13.9.2 and there are no console errors 🤔

It reloads the view when I connect to the API on the test-server (v13.9.1), but not when using the local instance (v13.9.2). It has nothing to do with the API version, but maybe with an async time issue and for sure with the `closeDialog.emit()` command in the `property-form.component.ts`. It runs twice (s. my comment [here](https://github.com/dasch-swiss/dsp-app/pull/455/files#r646620193)). I'll fix it...

src/app/project/ontology/property-form/property-form.component.ts
kilchenmann: fixed in b56a43a

mdelez (approved):
Well done :)

PR #457, opened 2021-06-07
feat(ontology): check if an ontology, a class or a property can be deleted (DSP-1750)

kilchenmann:
resolves DSP-1700
is required for DSP-1623

In this PR I added the new DSP-API routes (DSP-1622) and DSP-JS methods (DSP-1673) to check if an ontology, a resource class or a property can be deleted or not.


src/app/project/ontology/property-info/property-info.component.html
```diff
@@ -52,7 +52,7 @@
                 <mat-icon>tune</mat-icon>
             </button>
             <span
-                [matTooltip]="(resClasses.length > 0 ? 'The property can\'t be removed because it\'s in use' : 'Remove property from resource class')">
+                [matTooltip]="((resClasses.length > 0 || !propCanBeDeleted) ? 'The property can\'t be removed because it\'s in use' : 'Remove property from resource class')">
```
mdelez, line 55: I think this logic needs to be refactored. I can delete the property even though it says I can't. AS a side note, I think the delete icon should be the trash can for consistency reasons. The icon currently being used is normally used for "backspace".

Screen Shot 2021-06-21 at 11 04 00

mdelez (changes requested):

waychal (commented):
It is same with me. I am able to delete the property even if it says that `This property can't be removed because it's in use`. I would suggest to disable the delete icon if the property / class can't be deleted. 
I am also able to to delete classes. Can you give example in which case I can not delete the class?

kilchenmann:
@waychal and @mdelez Thanks for the report. Yes, there's something wrong with the logic. In this case the tooltip needs a different text, because here we want to remove (this is why I don't use the trash-bin icon) the property from the resource class; it's not deleting the property. This part has to be done in the properties section / `.../property` route itself. I'm using in both view the same component, but the functionality is different. I will fix it

kilchenmann:
I figured out what happened: I added the `propCanBeDeleted` condition to the wrong place. Resolved in 5db5353

The "remove property from resource class" will be implemented / fixed in the next iteration...

src/app/project/ontology/property-info/property-info.component.html
kilchenmann: resolved in 5db5353

mdelez (approved):

kilchenmann:
Thank you @mdelez

PR #461, opened 2021-06-10
chore(cookie-policy): reactivate the cookie policy banner (DSP-1727)

kilchenmann:
resolves DSP-1727

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -8,4 +8,17 @@
         <router-outlet></router-outlet>
     </div>
 
+    <!-- cookie information banner -->
+    <div class="cookie-banner" *ngIf="showCookieBanner">
+        <p class="note">
+            This web-application uses cookies to provide you with great user experience. By using the application you
+            accept our
+            <a href="./cookie-policy">use of cookies</a>.
+        </p>
+        <div class="action">
+            <button mat-flat-button color="accent" (click)="closeCookieBanner()">
+                ACCEPT
+            </button>
+        </div>
+    </div>
 </div>
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -11,7 +11,7 @@
     <!-- cookie information banner -->
     <div class="cookie-banner" *ngIf="showCookieBanner">
         <p class="note">
-            This web-application uses cookies to provide you with great user experience. By using the application you
+            This web-application uses cookies to provide you with a greater user experience. By using the application you
             accept our <span class="link" (click)="goToCookiePolicy()">use of cookies</span>.
         </p>
         <div class="action">
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -12,8 +12,7 @@
     <div class="cookie-banner" *ngIf="showCookieBanner">
         <p class="note">
             This web-application uses cookies to provide you with great user experience. By using the application you
-            accept our
-            <a href="./cookie-policy">use of cookies</a>.
+            accept our <span class="link" (click)="goToCookiePolicy()">use of cookies</span>.
         </p>
         <div class="action">
             <button mat-flat-button color="accent" (click)="closeCookieBanner()">
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -1,3 +1,5 @@
+@import "../assets/style/theme";
+
 
 .cookie-banner {
   position: fixed;
@@ -13,8 +15,8 @@
   .note {
     text-align: center;
 
-    a {
-      color: inherit;
+    .link {
+      color: $accent;
       text-decoration: underline;
     }
   }
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -1,0 +1,26 @@
+
+.cookie-banner {
+  position: fixed;
+  bottom: 0;
+  height: 128px;
+  width: 100%;
+  background: rgba(0, 0, 0, 0.75);
+  color: #fff;
+  z-index: 100;
+  // display: inline-flex;
+  padding: 24px 0;
+
+  .note {
+    text-align: center;
+
+    a {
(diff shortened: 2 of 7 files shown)


kilchenmann:
The "history back" button doesn't work anymore. Will fix it...

src/app/app.component.html
```diff
@@ -8,4 +8,16 @@
         <router-outlet></router-outlet>
     </div>
 
+    <!-- cookie information banner -->
+    <div class="cookie-banner" *ngIf="showCookieBanner">
+        <p class="note">
+            This web-application uses cookies to provide you with great user experience. By using the application you
```
mdelez, line 14: with a greater user experience

mdelez (changes requested):
I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.

kilchenmann:
> I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.

But what should happen, when the user click on "reject"? Redirect the user to a google page? Because in this case s/he can't use the app. 
This is anyway a temporary solution only. Daniela and Marchin are working on a better implementation, where the user can select which cookies should be enabled.

src/app/app.component.html
kilchenmann: rewritten in fd2ecf2

mdelez:
> > I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.
> 
> But what should happen, when the user click on "reject"? Redirect the user to a google page? Because in this case s/he can't use the app.
> This is anyway a temporary solution only. Daniela and Marchin are working on a better implementation, where the user can select which cookies should be enabled.

The website should still be useable even if the user rejected all **non-essential** cookies. We can still use cookie that are essential to the website operating. The user just needs to be allowed the reject any cookie that may have identifying data. https://gdpr.eu/cookies/#:~:text=To%20comply%20with%20the%20regulations,cookies%20except%20strictly%20necessary%20cookies.&text=Allow%20users%20to%20access%20your,the%20use%20of%20certain%20cookies

kilchenmann:
> > > I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.
> > 
> > 
> > But what should happen, when the user click on "reject"? Redirect the user to a google page? Because in this case s/he can't use the app.
> > This is anyway a temporary solution only. Daniela and Marchin are working on a better implementation, where the user can select which cookies should be enabled.
> 
> The website should still be useable even if the user rejected all **non-essential** cookies. We can still use cookie that are essential to the website operating. The user just needs to be allowed the reject any cookie that may have identifying data. https://gdpr.eu/cookies/#:~:text=To%20comply%20with%20the%20regulations,cookies%20except%20strictly%20necessary%20cookies.&text=Allow%20users%20to%20access%20your,the%20use%20of%20certain%20cookies

Yes I know and you're right, but in DSP-APP we have only **essential** cookies (the one for the login). When the user reject them s/he can't use the app. This is why I wrote: "If you want to use the app, you have to accept the cookie policy".

mdelez:
> > > > I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.
> > > 
> > > 
> > > But what should happen, when the user click on "reject"? Redirect the user to a google page? Because in this case s/he can't use the app.
> > > This is anyway a temporary solution only. Daniela and Marchin are working on a better implementation, where the user can select which cookies should be enabled.
> > 
> > 
> > The website should still be useable even if the user rejected all **non-essential** cookies. We can still use cookie that are essential to the website operating. The user just needs to be allowed the reject any cookie that may have identifying data. https://gdpr.eu/cookies/#:~:text=To%20comply%20with%20the%20regulations,cookies%20except%20strictly%20necessary%20cookies.&text=Allow%20users%20to%20access%20your,the%20use%20of%20certain%20cookies
> 
> Yes I know and you're right, but in DSP-APP we have only **essential** cookies (the one for the login). When the user reject them s/he can't use the app. This is why I wrote: "If you want to use the app, you have to accept the cookie policy".

That's fair. So then at that point I don't think we even need a cookies banner. You could also just add a "Reject non-essential cookies" button that does nothing but dismiss the banner since we don't have any non-essential cookies.

kilchenmann:
> > > > > I think there needs to be a "Reject" button as well, no? There is currently no way to dismiss the banner besides clicking the "Accept" button.
> > > > 
> > > > 
> > > > But what should happen, when the user click on "reject"? Redirect the user to a google page? Because in this case s/he can't use the app.
> > > > This is anyway a temporary solution only. Daniela and Marchin are working on a better implementation, where the user can select which cookies should be enabled.
> > > 
> > > 
> > > The website should still be useable even if the user rejected all **non-essential** cookies. We can still use cookie that are essential to the website operating. The user just needs to be allowed the reject any cookie that may have identifying data. https://gdpr.eu/cookies/#:~:text=To%20comply%20with%20the%20regulations,cookies%20except%20strictly%20necessary%20cookies.&text=Allow%20users%20to%20access%20your,the%20use%20of%20certain%20cookies
> > 
> > 
> > Yes I know and you're right, but in DSP-APP we have only **essential** cookies (the one for the login). When the user reject them s/he can't use the app. This is why I wrote: "If you want to use the app, you have to accept the cookie policy".
> 
> That's fair. So then at that point I don't think we even need a cookies banner. You could also just add a "Reject non-essential cookies" button that does nothing but dismiss the banner since we don't have any non-essential cookies.

As Subotic said yesterday, I should reactivate the banner. So that's what I did.

mdelez (approved):

kilchenmann:
Thank you @mdelez

PR #472, opened 2021-06-23
feat(upload)!: add upload form for still images (DSP-1761)

kilchenmann:
resolves DSP-1761

In this PR I copied the upload file component from DSP-UI directly to DSP-APP.
Additional I fixed an issue in the resource instance form in the project selection. A user should be able to select only the active projects (where s/he's member of) and the Knora and the System project should also not be displayed.

This PR contains a breaking change: The config file needs now the URL of our iiif service (sipi).


kilchenmann:
https://user-images.githubusercontent.com/4253580/123646346-6a14bc00-d827-11eb-9aac-dec37425bfae.mov

subotic (approved):

PR #478, opened 2021-07-02
feat(ontology): change gui element for text value properties

kilchenmann:
resolves DSP-1624

and 
- disable property form when adding existing property to resource class
- disable period property because there's no difference to a single date property (as discussed with @gautschr)

incl. some bug fixes:
- cardinality "multiple" is not allowed in boolean values &rarr; disabled form element in this case
- display correct icon in list of properties (it was wrong in case of decimal value property)


subotic (approved):

waychal (approved):
Everything looks fine after testing the functionality using test server.

PR #481, opened 2021-07-07
feat(resource): upload pdf document (DSP-1776)

kilchenmann:
resolves DSP-1776

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@angular/platform-browser-dynamic": "^11.2.9",
         "@angular/router": "^11.2.9",
         "@ckeditor/ckeditor5-angular": "^1.2.3",
-        "@dasch-swiss/dsp-js": "^2.5.0",
+        "@dasch-swiss/dsp-js": "^2.6.0",
         "@dasch-swiss/dsp-ui": "^1.6.0",
         "@ngx-translate/core": "^12.1.2",
         "@ngx-translate/http-loader": "5.0.0",
@@ -2257,9 +2257,9 @@
       }
     },
     "node_modules/@dasch-swiss/dsp-js": {
-      "version": "2.5.0",
-      "resolved": "https://registry.npmjs.org/@dasch-swiss/dsp-js/-/dsp-js-2.5.0.tgz",
-      "integrity": "sha512-Zd13732lmd/yS/LrfkJj2KJ1TnGWOuXsvffTMXMYpiYtTxdWl90GyFtgbt7IVtdcPyOaGrg8gZpTVssslUSMBw==",
+      "version": "2.6.0",
+      "resolved": "https://registry.npmjs.org/@dasch-swiss/dsp-js/-/dsp-js-2.6.0.tgz",
+      "integrity": "sha512-l6Aa5jsD8on/QHwH4egom8V9o0bgf2LbibMqpZxw4vgcY/pbxSuCSnOnJw+DwIS+cDUvrDI4GU1eT8TyV1hylw==",
       "dependencies": {
         "@types/jsonld": "^1.5.0",
         "json2typescript": "1.4.1",
@@ -22060,9 +22060,9 @@
       }
     },
     "@dasch-swiss/dsp-js": {
-      "version": "2.5.0",
-      "resolved": "https://registry.npmjs.org/@dasch-swiss/dsp-js/-/dsp-js-2.5.0.tgz",
-      "integrity": "sha512-Zd13732lmd/yS/LrfkJj2KJ1TnGWOuXsvffTMXMYpiYtTxdWl90GyFtgbt7IVtdcPyOaGrg8gZpTVssslUSMBw==",
+      "version": "2.6.0",
+      "resolved": "https://registry.npmjs.org/@dasch-swiss/dsp-js/-/dsp-js-2.6.0.tgz",
+      "integrity": "sha512-l6Aa5jsD8on/QHwH4egom8V9o0bgf2LbibMqpZxw4vgcY/pbxSuCSnOnJw+DwIS+cDUvrDI4GU1eT8TyV1hylw==",
       "requires": {
         "@types/jsonld": "^1.5.0",
         "json2typescript": "1.4.1",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@angular/platform-browser-dynamic": "^11.2.9",
     "@angular/router": "^11.2.9",
     "@ckeditor/ckeditor5-angular": "^1.2.3",
-    "@dasch-swiss/dsp-js": "^2.5.0",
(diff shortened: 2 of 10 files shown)


waychal (commented):
I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.


https://user-images.githubusercontent.com/25031092/124908597-c3c87380-dfe9-11eb-97d2-163996854f17.mov

kilchenmann:
> I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.
> 
>  Screen.Recording.2021-07-08.at.12.33.49.mov

Thanks @waychal for your comment. There are two things:

The resource instance form has a bug in the label: it does not allow to use special characters. For me it's not understandable, why the label is restricted this way. But it's another topic and has to be resolved in another task. This task is only the implementation of an upload form for documents and that leads me to the second point: The viewer for a document is not implemented yet and will be handled in a separate task.

waychal:
> > I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.
> > Screen.Recording.2021-07-08.at.12.33.49.mov
> 
> Thanks @waychal for your comment. There are two things:
> 
> The resource instance form has a bug in the label: it does not allow to use special characters. For me it's not understandable, why the label is restricted this way. But it's another topic and has to be resolved in another task. This task is only the implementation of an upload form for documents and that leads me to the second point: The viewer for a document is not implemented yet and will be handled in a separate task.

Ok. But then in this case, how can I test the functionality? How can I know that the file is actually uploaded? I would like to get at least some alert or console message saying that if it is successful or there is an error.

kilchenmann:
> > > I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.
> > > Screen.Recording.2021-07-08.at.12.33.49.mov
> > 
> > 
> > Thanks @waychal for your comment. There are two things:
> > The resource instance form has a bug in the label: it does not allow to use special characters. For me it's not understandable, why the label is restricted this way. But it's another topic and has to be resolved in another task. This task is only the implementation of an upload form for documents and that leads me to the second point: The viewer for a document is not implemented yet and will be handled in a separate task.
> 
> Ok. But then in this case, how can I test the functionality? How can I know that the file is actually uploaded? I would like to get at least some alert or console message saying that if it is successful or there is an error.

I know this is a problem, but the upload is needed to start with the document viewer. You can have a look in the `Network` tab in browser's developer tools, if the API responses with the correct object and if this object has information about the file value. Additional it should be possible to open the pdf with sipi. But there, in some cases we have some permission problems. When you test it locally, you can check in finder if the file is in the sipi image folder `dsp-api/sipi/images/[project-shortcode-folder]/`. In this case the file upload was successful.

waychal:
> > > > I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.
> > > > Screen.Recording.2021-07-08.at.12.33.49.mov
> > > 
> > > 
> > > Thanks @waychal for your comment. There are two things:
> > > The resource instance form has a bug in the label: it does not allow to use special characters. For me it's not understandable, why the label is restricted this way. But it's another topic and has to be resolved in another task. This task is only the implementation of an upload form for documents and that leads me to the second point: The viewer for a document is not implemented yet and will be handled in a separate task.
> > 
> > 
> > Ok. But then in this case, how can I test the functionality? How can I know that the file is actually uploaded? I would like to get at least some alert or console message saying that if it is successful or there is an error.
> 
> I know this is a problem, but the upload is needed to start with the document viewer. You can have a look in the `Network` tab in browser's developer tools, if the API responses with the correct object and if this object has information about the file value. Additional it should be possible to open the pdf with sipi. But there, in some cases we have some permission problems. When you test it locally, you can check in finder if the file is in the sipi image folder `dsp-api/sipi/images/[project-shortcode-folder]/`. In this case the file upload was successful.

Ok. I do not have dsp-api running locally because of new MacBook processor. So I can not test it. 
I can see the requests in network tab but I am not very sure about this.
For now I am approving this PR but it would be good if someone else also have a look at it.

kilchenmann:
> > > > > I am running DSP-APP locally using test server configuration. When I upload file, I do not see it once I save the changes. I have attached recording below. Also When I add resource label including `-`(e.g. `file-test`, it disables `next` button and gives me wrong error in red text below. If I remove `-`, it works fine.
> > > > > Screen.Recording.2021-07-08.at.12.33.49.mov
> > > > 
> > > > 
> > > > Thanks @waychal for your comment. There are two things:
> > > > The resource instance form has a bug in the label: it does not allow to use special characters. For me it's not understandable, why the label is restricted this way. But it's another topic and has to be resolved in another task. This task is only the implementation of an upload form for documents and that leads me to the second point: The viewer for a document is not implemented yet and will be handled in a separate task.
> > > 
> > > 
> > > Ok. But then in this case, how can I test the functionality? How can I know that the file is actually uploaded? I would like to get at least some alert or console message saying that if it is successful or there is an error.
> > 
> > 
> > I know this is a problem, but the upload is needed to start with the document viewer. You can have a look in the `Network` tab in browser's developer tools, if the API responses with the correct object and if this object has information about the file value. Additional it should be possible to open the pdf with sipi. But there, in some cases we have some permission problems. When you test it locally, you can check in finder if the file is in the sipi image folder `dsp-api/sipi/images/[project-shortcode-folder]/`. In this case the file upload was successful.
> 
> Ok. I do not have dsp-api running locally because of new MacBook processor. So I can not test it.
> I can see the requests in network tab but I am not very sure about this.
> For now I am approving this PR but it would be good if someone else also have a look at it.

Sure. Thank you @waychal . I'm pretty sure that @mdelez will also have a look at it.

mdelez (commented):
Screen Shot 2021-07-09 at 11 33 12

The upload works well! However I've noticed that you can save the file without providing a title, is this intended?

Also it would be nice to have title label fit nicely. Does it look okay on your end @kilchenmann?
Screen Shot 2021-07-09 at 11 11 02

mdelez (approved):

PR #483, opened 2021-07-12
fix(ontology): fix regex pattern in ontology form (DSP-1139)

kilchenmann:
resolves DSP-1139

We had an issue in the "create ontology" form. The regex didn't work as expected. As the DSP-API documentation says: "An ontology name must be a valid XML NCName. [...] A user-created ontology name may not start with the letter v followed by a digit". Special characters (incl. Umlauts etc), spaces and leading numbers are not allowed.
In this PR I fixed the regex and turned off the autocomplete (because it hides the error message below the field).


mdelez (approved):
Nice :)

PR #485, opened 2021-07-16
feat(resource): add document viewer with download (DSP-1791)

kilchenmann:
resolves DSP-1791

In this PR I actually wanted to integrate a simple download button for pdf file representations. But the main work was to generalise how we handle the different representation types. Previous everything was prepared for still images only. This is why I had to create a general `FileRepresentation` type. Additional I integrated a simple angular pdf viewer ([ng2-pdf-viewer](https://www.npmjs.com/package/ng2-pdf-viewer)) and styled it the same way as we already did for the opeansedragon viewer for still images. I just had to adjust the still image viewer resp. I had to move the style to our main styling files to use the same setup for the new document viewer.


kilchenmann:
This is how the still image viewer looks like:
Screen Shot 2021-07-16 at 16 26 12

And here's the new document viewer:
Screen Shot 2021-07-16 at 16 26 02

waychal (commented):
For me `open file in new tab` button is not working but it is expected as I am using the test server for dsp-api. Other than that everything looks fine to me.

I will let @mdelez to approve the PR as he could also test the `open file in new tab` functionality.

mdelez (approved):
Works great and I can open my files in a new tab :)

PR #486, opened 2021-07-19
feat(resource): upload audio (DSP-1799)

kilchenmann:
resolves DSP-1799

~~Upload audio file works, but we do not have an audio player yet; this will be implemented in the next iteration. It's the old chicken or the egg dilemma. It means for the reviewers, when you want to test it (even it's a code review only), you have to check the API answer in the developer tools or you should find the file in the `dsp-api/sipi/images/[project]/`-folder~~


mdelez (changes requested):
This works well if I use it exactly how I'm supposed to but I noticed that if I upload a file (via drag and drop) and then delete it and click save, it scrolls to the upload section. If I then upload a file and click save I get this error and no matter what I do I cannot save the resource:
Screen Shot 2021-07-22 at 10 50 26

Another thing I noticed is that if I edit my data model, the changes aren't reflected immediately if I click on the plus button on the top right to create a new resource. But maybe this could be a separate PR.
Screen Shot 2021-07-22 at 10 44 23

kilchenmann:
> This works well if I use it exactly how I'm supposed to but I noticed that if I upload a file (via drag and drop) and then delete it and click save, it scrolls to the upload section. If I then upload a file and click save I get this error and no matter what I do I cannot save the resource:

--> I'm able to reproduce the issue, but I don't get the error message. But I figured out, that the submit button is not disabled anymore in case the form is wrong / has missing values in required fields 

> Another thing I noticed is that if I edit my data model, the changes aren't reflected immediately if I click on the plus button on the top right to create a new resource. But maybe this could be a separate PR.

--> this is indeed a separate task. Would you like to write a bug report in youtrack and assign to me?

mdelez:
> > This works well if I use it exactly how I'm supposed to but I noticed that if I upload a file (via drag and drop) and then delete it and click save, it scrolls to the upload section. If I then upload a file and click save I get this error and no matter what I do I cannot save the resource:
> 
> --> I'm able to reproduce the issue, but I don't get the error message. But I figured out, that the submit button is not disabled anymore in case the form is wrong / has missing values in required fields
> 
> > Another thing I noticed is that if I edit my data model, the changes aren't reflected immediately if I click on the plus button on the top right to create a new resource. But maybe this could be a separate PR.
> 
> --> this is indeed a separate task. Would you like to write a bug report in youtrack and assign to me?

1. Yes, ideally the save button should be disabled if the form is invalid but I wasn't sure if this was set up already
2. Bug report opened -> DSP-1810

src/app/workspace/resource/representation/upload/upload.component.ts
```diff
@@ -283,6 +295,9 @@ export class UploadComponent implements OnInit {
             case 'document':
                 this.allowedFileTypes = this.supportedDocumentTypes;
                 break;
+            case 'audio':
+                this.allowedFileTypes = this.supportedAudioTypes;
+                break;
             default:
```
Vijeinath, line 301: Not severe: In this file there are incongruent blank lines within switch blocks. Either use a blank line after every case block or no blank line after case block

Vijeinath (approved):
I'm happy with it

src/app/workspace/resource/representation/upload/upload.component.ts
kilchenmann: resolved in 8f6c0e4

kilchenmann:
> 1. Yes, ideally the save button should be disabled if the form is invalid but I wasn't sure if this was set up already

--> should be resolved now

mdelez (approved):
Save button works appropriately now :)

PR #489, opened 2021-07-20
feat(resource): delete and erase resource (DSP-1228)

kilchenmann:
resolves DSP-1228

What I did:

* new section in dialog component for delete (with comment) and for erase (without comment)
* integration of delete and erase methods from js-lib
* after deleting it displays `(deleted)` behind the label
* app-specific message banner to display info, that the resource has been deleted
* if resource doesn't exist (anymore) the api returns a 404 error. In this case display text, that the resource does not exist or could not be found.
* check if user is system or project admin; this information is needed to enable or disable the erase button.
* not sure how to handle permissions: the edit menu is activated in case a user is logged-in


mdelez (commented):
If I edit the value of a resource property, I can no longer delete or erase the resource. Seems to be an issue with the API:
Screen Shot 2021-07-22 at 11 27 34

kilchenmann:
> If I edit the value of a resource property, I can no longer delete or erase the resource. Seems to be an issue with the API:

Thank you for this report. Yes, in this case we should take the last modification date somehow into account.

kilchenmann:
@mdelez I fixed the issue you mentioned (<https://github.com/dasch-swiss/dsp-app/pull/489#pullrequestreview-712538176>) in commit ad54fa8. But it's not the best solution yet. I think we should emit the last modification date from ui-lib's property-value component. I wrote already a corresponding issue on youtrack: DSP-1813

kilchenmann:
@mdelez should be ready for review again

mdelez (approved):
Nice :)

PR #492, opened 2021-07-27
feat(resource): update resource's label (DSP-1801)

kilchenmann:
resolves DSP-1801

### DIFF
--- a/src/app/main/dialog/dialog.component.html
+++ b/src/app/main/dialog/dialog.component.html
@@ -281,21 +281,24 @@
     <div *ngSwitchCase="'updateCardinality'">
         <app-dialog-header [title]="data.title" [subtitle]="data.subtitle">
         </app-dialog-header>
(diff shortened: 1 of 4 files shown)


kilchenmann:
There are two issues which can't be resolved right now. Both are described in the Youtrack's task:

- <https://dasch.myjetbrains.com/youtrack/issue/DSP-1801#focus=Comments-4-3279.0-0>
- <https://dasch.myjetbrains.com/youtrack/issue/DSP-1801#focus=Comments-4-3280.0-0>

kilchenmann:
@mdelez what should we do with the two issues mentioned in previous comment? In second case it needs an update of the API or should I do a second request to get the necessary information? Any suggestions?

kilchenmann:
This is the first iteration. It has to be improved as soon as DSP-API resolved the "update metadata" response. At the moment it's not possible to edit the label twice without refreshing the whole page. 
For the moment we can't wait for new release of DSP-API because this will be v14.0.0 which contains a lot of breaking changes.

mdelez (approved):
looks good :)

PR #504, opened 2021-08-13
refactor(resource): migrate viewer from UI-lib (DSP-1850)

mdelez:
resolves DSP-1850


mdelez:
idk what github wants from me. It works on my machine. Seems like a timing issue since it's not always the same test that fails when I re-run the tests.

kilchenmann (approved):

PR #507, opened 2021-08-13
feat(resource): display incoming links (DSP-1846)

Vijeinath:
resolves DSP-1846

I added also a pagination to use in case there are more than 25 incoming links

### DIFF
--- a/src/app/workspace/resource/properties/properties.component.html
+++ b/src/app/workspace/resource/properties/properties.component.html
@@ -156,9 +156,21 @@
                     </h3>
                 </div>
                 <div class="property-value">
-                    <!-- the value(s) of the incoming links -->
-                    <div *ngFor="let inRes of incomingLinkResources">
-                        <a [routerLink]="['/resource', inRes.id]" target="_blank">{{inRes.label}}</a>
+                    <div *ngIf="loading">
+                        ...Loading
+                    </div>
+                    <div *ngIf="!loading">
+                        <!-- the value(s) of the incoming links -->
+                        <div *ngFor="let inRes of incomingLinkResources">
+                            <a [routerLink]="['/resource', inRes.id]" target="_blank">{{inRes.label}}</a>
+                        </div>
+                        <mat-paginator
+                                [length]=numberOffAllIncomingLinkRes
+                                [pageSize]="25"
+                                [hidePageSize]="true"
+                                [pageIndex]="pageEvent.pageIndex"
+                                (page)="goToPage($event)">
+                        </mat-paginator>
                     </div>
                 </div>
             </div>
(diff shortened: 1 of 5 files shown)


src/app/workspace/resource/properties/properties.component.html
```diff
@@ -146,6 +146,36 @@ <h3 class="label mat-title">
...
+        <div *ngIf="incomingLinkResources.length > 0">
+            <div class="property border-top">
+                <div class="property-label">
+                    <!-- label of the incoming links -->
+                    <h3 class="label mat-subheading-1">
+                        has incoming link
+                    </h3>
+                </div>
+                <div class="property-value">
+                    <div *ngIf="loading">
```
kilchenmann, line 159: You should use the `dsp-progress-indicator` in case of loading

src/app/workspace/resource/properties/properties.component.scss
```diff
@@ -166,6 +166,15 @@
   color: rgba(0, 0, 0, 0.54);
 }
 
+:host::ng-deep .mat-paginator-container {
```
kilchenmann, line 169: `ng-deep` is deprecated. It would be a better way to define/overwrite component's style in one of the global stylesheets in the assets folder. But for the moment you can keep it as it is.

src/app/workspace/resource/properties/properties.component.spec.ts
```diff
@@ -17,7 +18,7 @@ import {
     MockUsers,
     ProjectsEndpointAdmin,
     ReadLinkValue,
-    ReadResource,
+    ReadResource, ReadResourceSequence,
```
kilchenmann, line 21: should go to a new line 😉

Vijeinath: Good

src/app/workspace/resource/properties/properties.component.scss
Vijeinath: Ok, I keep this for the moment. Is there an alternative?

src/app/workspace/resource/properties/properties.component.html
Vijeinath: I think right now its not in the dsp-app

mdelez: you can either import it from dsp-ui or leave a comment to replace it with the progress indicator once it's been migrated to the app

src/app/workspace/resource/properties/properties.component.scss
mdelez: There's no alternative for `ng-deep` and it doesn't appear as if there ever will be an alternative so you'll have to overwrite the component's style either in one of the global style sheets (recommended) or directly in the stylesheet of the component

mdelez (approved):

PR #509, opened 2021-08-17
refactor(action): migrate action module (DSP-1852)

mdelez:
resolves DSP-1852


mdelez:
@subotic it's a ton of code but it's all code that we already know works. This PR is a simple migration of the code from the `ui-lib`, it **does not** update the imports within `dsp-app` to use this migrated code. That will happen in a future PR since I will basically touch every file.

subotic (approved):

PR #524, opened 2021-09-07
feat(resource): draw regions (DSP-1845)

EricSommerhalder:
resolves DSP-1845


src/app/workspace/resource/resource.component.html
```diff
@@ -9,6 +9,7 @@
                 [images]="representationsToDisplay"
                 [imageCaption]="(incomingResource ? resource.res.label + ': ' + incomingResource.res.label : resource.res.label)"
                 [compoundNavigation]="compoundPosition" (goToPage)="compoundNavigation($event)"
+                             [resourceIri]="incomingResource ? incomingResource.res.id : resource.res.id" [project]="resource.res.attachedToProject"
```
kilchenmann, line 12: Just a tiny detail. The indention is wrong resp. could you place each property on a separate line. The `(goToPage)` should be the last one. We defined (somewhere) the rule that we have first the normal parameters followed by the events. But as I said it's just a detail. Otherwise it looks good.

kilchenmann (commented):
The code looks good, but I also want to try it

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
```diff
@@ -0,0 +1,42 @@
+<form [formGroup]="regionForm">
+    <mat-form-field>
+        <mat-label>Comment</mat-label>
```
kilchenmann, line 3: Add an asterisk to the label, so the user knows that a value is required.

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
```diff
@@ -0,0 +1,42 @@
...
+        <input matInput formControlName="comment">
+    </mat-form-field>
+    <mat-form-field>
+        <mat-label>Label</mat-label>
```
kilchenmann, line 7: Add an asterisk to the label, so the user knows that a value is required.

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
```diff
@@ -0,0 +1,42 @@
+<form [formGroup]="regionForm">
+    <mat-form-field>
+        <mat-label>Comment</mat-label>
+        <input matInput formControlName="comment">
```
kilchenmann, line 4: Comment field should be a textarea. But I think the form style could also be done in a separate, style refactoring PR.

kilchenmann:
After creating a region, it does not update the resource viewer. Only after refreshing the page I got the region in the annotation tab. 

And for any reason, I don't know why, it does not display the "is region of" in the region properties. I've checked and it's not an issue of the create region function. This has to be resolved in another PR; in the same PR we should also fix the typo in the german label "is Region von".

Screen Shot 2021-09-09 at 07 47 23

src/app/workspace/resource/representation/add-region-form/add-region-form.component.ts
```diff
@@ -0,0 +1,24 @@
...
+    @Input() resourceIri: string;
+    regionForm: FormGroup;
+    colorPattern = '^#[a-f0-9]{6}$';
+    constructor(private _fb: FormBuilder) {
+    }
+
+    ngOnInit(): void {
+        this.regionForm = this._fb.group({
+            color: ['#ff3333', [Validators.required, Validators.pattern(this.colorPattern)]],
+            comment: [null, Validators.required],
```
mdelez, line 19: is a comment actually required to submit the region?

src/app/workspace/resource/representation/still-image/still-image.component.ts
```diff
@@ -91,12 +114,14 @@ export class StillImageComponent implements OnChanges, OnDestroy {
 
     @Output() regionClicked = new EventEmitter<string>();
 
+    private _regionDrawMode: Boolean = false; // stores whether viewer is currently drawing a region
+    private _regionDragInfo; // stores the information of the first click for drawing a region
+    private _mouseTracker; // stores the MouseTracker that allows drawing regions
```
mdelez, line 119: is this used anywhere? My IDE says it's unused so it can probably be removed

mdelez (changes requested):

mdelez:
> After creating a region, it does not update the resource viewer. Only after refreshing the page I got the region in the annotation tab.
> 

+1 this should be handled in this PR. I've also noticed that editing the color of the region afterwards does not change the actual color of the region, even after refreshing but this can be handled in a different PR.

src/app/workspace/resource/resource.component.html
EricSommerhalder: Thanks! Done.

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
EricSommerhalder: Done

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
EricSommerhalder: Done

src/app/workspace/resource/representation/add-region-form/add-region-form.component.html
EricSommerhalder: I changed it to a textarea for now, but I agree that a style refactoring PR will be needed.

src/app/workspace/resource/representation/add-region-form/add-region-form.component.ts
EricSommerhalder: Yes, otherwise the api is not happy. Not sure why though.

src/app/workspace/resource/representation/still-image/still-image.component.ts
EricSommerhalder: At some point this got the whole thing running, but apparently it is not needed anymore. I have removed it.

mdelez (approved):
Thanks!

kilchenmann (approved):

PR #532, opened 2021-09-12
feat(resource): new date picker (DSP-1769)

kilchenmann:
resolves DSP-1769


kilchenmann:
@mdelez I just fixed the `expressionChangedAfterItHasBeenCheckedError` we had in date period. Additional I've added the new date picker to the advanced search. Now everything should work and is ready to be reviewed.

PR #534, opened 2021-09-15
refactor: openseadragon prod build fix (DSP-1779)

mdelez:
resolves DSP-1779


kilchenmann (approved):
wow! LGTM. Let's try....

EricSommerhalder:
Looks good to me as well. I feel like we should document somewhere that we only support rectangular regions at the moment though. I'm not sure where.

EricSommerhalder (approved):

mdelez:
> Looks good to me as well. I feel like we should document somewhere that we only support rectangular regions at the moment though. I'm not sure where.

Yeah, that's a good point. I'll check the docs and see if they can be updated.

kilchenmann:
> > Looks good to me as well. I feel like we should document somewhere that we only support rectangular regions at the moment though. I'm not sure where.
> 
> Yeah, that's a good point. I'll check the docs and see if they can be updated.

The whole part about "Working with resources" is not yet written (or only partial). Here's the (data) [user guide](https://docs.dasch.swiss/DSP-APP/user-guide/data/) which has the be improved completely.

PR #535, opened 2021-09-16
refactor(modules): clean up imports and npm packages

kilchenmann:
Sort imports, declarations and modules in app and material modules.
Run `npm prune --legacy-peer-deps` which refactored the package-lock.json


mdelez (approved):
Nice, thanks :)

PR #538, opened 2021-09-17
fix(resource): increase width of space between entries of incoming links (DSP-1908)

kilchenmann:
resolves DSP-1908


mdelez (approved):
Looks good! Thanks for fixing my CSS 😆

PR #540, opened 2021-09-20
fix(error): improve the current error handler in DSP-APP (DSP-1911)

kilchenmann:
resolves DSP-1911

You can test it as follows:

- Start the app with the test-server configuration
- Open the following page: <http://0.0.0.0:4200/search/fulltext/Stygimoloch>
- If you're logged-in, you should see the correct error message from JS-Lib instead the 503 error mentioned in DSP-1910

Screen Shot 2021-09-20 at 22 42 55

Or

- Start the app with your local api stack environment
- Turn off the API
- You should see the 503 server error


mdelez (approved):
Works great!

PR #542, opened 2021-09-21
fix: gravsearch results don't appear after page refresh (DSP-1675)

mdelez:
resolves DSP-1675


kilchenmann (approved):
works perfect 🎉  thanks

PR #547, opened 2021-10-04
fix(resource-instance-form): reload cached resource (DSP-1810)

mdelez:
resolves DSP-1810


kilchenmann (approved):
Seems to work perfect

PR #550, opened 2021-10-14
fix: disable edit/delete action in deactivated projects (DEV-52)

kilchenmann:
resolves DEV-52


kilchenmann:
@mdelez the issue should be resolved now (e9b1fb50)

PR #571, opened 2021-11-01
fix(value): display ckEditor in case of rich-text property (DEV-182)

kilchenmann:
resolves DEV-182 incl. update of ckEditor version and style refactoring of value components in general


mdelez (approved):
Looks good, thanks!

PR #572, opened 2021-11-02
refactor: move datadog rum methods to service (DEV-190)

mdelez:
resolves DEV-190


src/app/app.component.ts
```diff
@@ -14,12 +15,15 @@ export class AppComponent implements OnInit {
...
-        private _titleService: Title
+        private _titleService: Title,
+        private _datadogRumService: DatadogRumService
     ) {
 
         // set the page title
         this._titleService.setTitle('DaSCH Service Platform');
 
+        // init datadog RUM
+        this._datadogRumService.initializeRum();
```
subotic, line 26: Why is the `datadogRumService` initialized here and not at time of construction?

src/app/main/services/datadog-rum.service.ts
```diff
@@ -0,0 +1,48 @@
...
+                    // collect a RUM resource's response headers
+                    if (event.type === 'resource' && event.resource.type === 'xhr') {
+                        event.context = { ...event.context, responseHeaders: (context as RumFetchResourceEventDomainContext).response.body };
+                    }
+                },
+            });
+        }
+    }
+
+    setActiveUser(identifier: any, identifierType: 'iri' | 'email' | 'username'): void {
```
subotic, line 38: you need to check if datadog.Rum is initialized, and only then set the user. Otherwise you can probably ignore it.

src/app/main/services/datadog-rum.service.ts
```diff
@@ -0,0 +1,48 @@
...
+        datadogRum.setUser({
+            id: identifier,
+            identifierType: identifierType
+        });
+    }
+
+    removeActiveUser(): void {
+        datadogRum.removeUser();
```
subotic, line 46: likewise, check if datadog.Rum is initialized.

subotic (changes requested):

src/app/app.component.ts
mdelez: changed in a48ea3d

src/app/main/services/datadog-rum.service.ts
mdelez: changed in a48ea3d

src/app/main/services/datadog-rum.service.ts
mdelez: changed in a48ea3d

subotic (approved):

PR #581, opened 2021-11-09
feat(resource)!: new resource route (DEV-196)

kilchenmann:
resolves DEV-196


src/app/workspace/resource/services/resource.service.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+                    useValue: appInitSpy
+                }
+            ]
+        });
+        service = TestBed.inject(ResourceService);
+    });
+
+    it('should be created', () => {
+        expect(service).toBeTruthy();
+    });
```
mdelez, line 28: I would add some tests here for the service methods. They can be simple tests but since this service is quite crucial to the whole app running, it would be good to ensure these methods are working as intended.

mdelez (changes requested):

src/app/workspace/resource/services/resource.service.spec.ts
kilchenmann: Done in f4a6b170

src/app/workspace/resource/services/resource.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                    provide: AppInitService,
+                    useValue: appInitSpy
+                }
+            ]
+        });
+        service = TestBed.inject(ResourceService);
+    });
+
+    it('should be created', () => {
+        expect(service).toBeTruthy();
```
mdelez, line 64: you can move this `expect` to line 61 and remove this `it` since it's a bit better to ensure the service is created correctly during each `BeforeEach()`. You can do the same with the `it` on line 26

mdelez (changes requested):

src/app/workspace/resource/services/resource.service.spec.ts
kilchenmann: Done in c89eef9

PR #605, opened 2021-12-08
fix(resource): create resource iri from route only in certain cases (DEV-306)

kilchenmann:
resolves DEV-306


mdelez (approved):
works well on my end, thanks!

PR #610, opened 2021-12-08
chore(deps): use correct nginx-server (DEEV-263)

kilchenmann:
resolves DEV-263


subotic (approved):
could you please change the Email to 400790+subotic@users.noreply.github.com

PR #616, opened 2021-12-09
fix(resource): update lastModificationDate after editing label (DEV-315)

kilchenmann:
resolves DEV-315


mdelez (approved):
works well, thanks!

PR #619, opened 2021-12-14
fix(upload): fix thumbnail preview (DEV-268)

kilchenmann:
resolves DEV-268


mdelez (approved):
works well on localhost, thanks!

PR #632, opened 2022-01-12
feat(resource): display deleted resource (DEV-299)

kilchenmann:
resolves DEV-299


kilchenmann:
This is how it works:

*Normal Resource* Screen Shot 2022-01-14 at 09 23 41

*Click on Delete Resource* 
Screen Shot 2022-01-14 at 09 23 48

*Dialog box to confirm and to set a comment*
Screen Shot 2022-01-14 at 09 24 54

*Status after deleting resource*
Screen Shot 2022-01-14 at 09 25 01

*Deleted resource when opening from ARK-URL*
Screen Shot 2022-01-14 at 09 25 11

src/app/workspace/resource/resource.component.html
```diff
@@ -34,7 +34,7 @@
         </div>
 
         <!-- tabs -->
-        <mat-tab-group animationDuration="0ms" [(selectedIndex)]="selectedTab" (selectedTabChange)="tabChanged($event)">
+        <mat-tab-group *ngIf="!resource.res.isDeleted; else deletedResource" animationDuration="0ms" [(selectedIndex)]="selectedTab" (selectedTabChange)="tabChanged($event)">
```
mdelez, line 37: I think we should also have a similar `ngIf` for the representation div of this file. As it is now, if I delete a resource with some sort of resource representation (audio, document, image, archive, etc.), I still see the representation. The resource representation is only hidden once I refresh the page.

src/app/workspace/resource/resource.component.html
```diff
@@ -69,6 +69,11 @@
 
             </mat-tab>
         </mat-tab-group>
+        <ng-template #deletedResource>
+            <app-properties *ngIf="resource.res" [resource]="resource" [displayProjectInfo]="true"
+                [adminPermissions]="adminPermissions" [editPermissions]="editPermissions" [valueUuidToHighlight]="valueUuid">
```
mdelez, line 74: are the `adminPermissions`, `editPermissions`, and `valueUuidToHighlight ` needed here?

mdelez (changes requested):

src/app/workspace/resource/resource.component.html
kilchenmann: oh yes, you're right. I'll update asap

src/app/workspace/resource/resource.component.html
kilchenmann: yes,  because it should be possible (at least in the future) to re-activate (or to erase) a resource. In this case we have to know the permissions in the propertiesComponent

mdelez: I was also wondering if further editing of a resource would eventually be possible. Good to know :)

src/app/workspace/resource/resource.component.html
kilchenmann: Yes, that's right. When deleting a resource, it's still visible as it was before (incl. all property values etc) because I do not get the updated resource information yet. In comparison to the current view of a deleted resource does not have any property and value information. But Rebecca is working on a better version, that we still get the whole resource information even when it's deleted. The current implementation in this PR is only a temporary solution to display something. In the previous version, the user was getting a blank page only.

mdelez (approved):

PR #633, opened 2022-01-12
fix(advanced search): update comparison operators in case of rich text (DEV-326)

kilchenmann:
resolves DEV-326


irinaschubert (approved):
LGTM, I assume you've tested it? Or should I do it?

kilchenmann:
> LGTM, I assume you've tested it? Or should I do it?

Thanks. Yes, I've tested it

PR #643, opened 2022-01-24
fix(ontology): support subproperty in ontology editor (DEV-332)

kilchenmann:
resolves DEV-332

### DIFF
--- a/src/app/project/ontology/default-data/default-properties.ts
+++ b/src/app/project/ontology/default-data/default-properties.ts
@@ -118,15 +118,6 @@ export class DefaultProperties {
                     guiEle: Constants.GuiDatePicker,
                     group: 'Date / Time'
                 },
-                // {
-                //     icon: 'date_range',
-                //     label: 'Period',
-                //     description: 'A period of time between two dates',
-                //     subPropOf: Constants.HasValue,
-                //     objectType: Constants.DateValue,
-                //     guiEle: Constants.GuiDatePicker',
-                //     group: 'Date / Time'
-                // },
                 {
                     icon: 'access_time',
                     label: 'Timestamp',
--- a/src/app/project/ontology/default-data/default-properties.ts
+++ b/src/app/project/ontology/default-data/default-properties.ts
@@ -33,6 +33,18 @@ export interface DefaultProperty {
 }
 
 export class DefaultProperties {
+
+    public static unsupported: DefaultProperty = {
+        icon: 'warning_amber',
+        label: 'Unsupported property type',
+        description: 'This property type is not supported in the ontology editor',
+        subPropOf: undefined,
+        objectType: undefined,
+        guiEle: undefined,
+        group: 'Warning'
+    };
+
+
     public static data: PropertyCategory[] = [
         {
             group: 'Text',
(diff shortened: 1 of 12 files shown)


src/app/project/ontology/default-data/default-properties.ts
```diff
@@ -112,7 +124,7 @@ export class DefaultProperties {
                 //     description: 'A period of time between two dates',
                 //     subPropOf: Constants.HasValue,
                 //     objectType: Constants.DateValue,
-                //     guiEle: Constants.SalsahGui + Constants.HashDelimiter + 'Date',
+                //     guiEle: Constants.GuiDatePicker',
```
mdelez, line 127: is this commented out code going to be needed in the future or can we remove it?

src/app/project/ontology/ontology.service.ts
```diff
@@ -79,4 +95,92 @@ export class OntologyService {
...
+                    // get property iri from another ontology
+                    this._cache.get('currentProjectOntologies').subscribe(
+                        (ontologies: ReadOntology[]) => {
+                            const onto = ontologies.find(i => i.id === baseOntoIri);
+                            superPropIri = onto.properties[subProp].subPropertyOf[0];
+                        }
+                    );
+                }
+
+                if (superPropIri) {
```
mdelez, line 121: Why does this if statement need to be here?

src/app/project/ontology/ontology.service.ts
```diff
@@ -79,4 +95,92 @@ export class OntologyService {
...
+                            ));
+                        } else {
+                            propType = (group.elements.find(i =>
+                                i.guiEle === property.guiElement && i.objectType === property.objectType
+                            ));
+                        }
+
+                    }
+                }
+                if (propType) {
```
mdelez, line 170: same as my comment above, maybe I'm missing something..

mdelez (changes requested):
Everything works well for me! I just have a few remarks/questions

src/app/project/ontology/default-data/default-properties.ts
kilchenmann: I thought there is or there will be a difference between a date and a date period. But in our case you can always create a date period from a "normal" date property. So, there's no additional date period property needed. I can remove this section.

src/app/project/ontology/ontology.service.ts
kilchenmann: As soon as the propType is defined (or found), we can exit the for-loop.

src/app/project/ontology/ontology.service.ts
kilchenmann: As soon as the superPropIri is defined (or found), we can exit the for-loop.

kilchenmann: Maybe it's not needed. I think I found this solution somewhere on stackoverflow 🙈

mdelez: I think it would work if you put the `break` on line 117, no?

kilchenmann: You mean like this:
Screen Shot 2022-02-01 at 10 10 38

Then I got an error

mdelez: Yeah that's what I meant, good to know it doesn't work like I thought it would 😅

mdelez (approved):

PR #645, opened 2022-01-31
feat(add-link-resource-button): Add Link Resource Button (DEV-404)

mdelez:
resolves DEV-404


src/app/workspace/resource/operations/create-link-resource/create-link-resource.component.html
```diff
@@ -0,0 +1,29 @@
...
+            </button>
+        </span>
+        <span class="fill-remaining-space"></span>
+        <span>
+            <button mat-raised-button type="submit" color="primary" class="form-submit" [disabled]="!propertiesForm.valid">
+                {{ 'appLabels.form.action.submit' | translate}}
+            </button>
+        </span>
+    </div>
+</form>
```
kilchenmann, line 29: Is the `create-link-resource.component` the same as in the second step of `resource-instance-form`? Does it make sense to duplicate the code?

mdelez: @kilchenmann they're quite similar but the `resource-instance-form` has a back button to go back to the first step and it redirects the user to the created resource after it's been created. `resource-instance-form` uses an array of `ResourceClassDefinition` whereas `create-link-resource` does not use an array. We could try to condense it all into one component but we might have a lot of ugly if statements to hide/show relevant parts of the component depending on what the user is doing. What are your thoughts?

kilchenmann: Thanks for your response. That's true. Sooner or later the resource-instance-form will anyway have only one form (without the first step) and the back button will not be needed anymore. I would appreciate it if we do not have too many of the same forms.

mdelez: Should I refactor in this PR or another one?

kilchenmann: What do you prefer? I think it should be resolved in this PR... or what will be the simplest solution?

mdelez: after speaking with André, we decided to do the refactor in another PR because the `create-resource-instance` form will change in the future

kilchenmann (approved):

PR #660, opened 2022-02-15
feat(resource): display permissions (DEV-454)

kilchenmann:
resolves DEV-454

### DIFF
--- a/src/app/workspace/resource/permission-info/permission-info.component.spec.ts
+++ b/src/app/workspace/resource/permission-info/permission-info.component.spec.ts
@@ -40,7 +40,7 @@ class TestHostComponent implements OnInit {
 
 }
 
-fdescribe('PermissionInfoComponent', () => {
+describe('PermissionInfoComponent', () => {
     let testHostComponent: TestHostComponent;
     let testHostFixture: ComponentFixture<TestHostComponent>;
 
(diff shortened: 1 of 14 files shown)


src/app/workspace/resource/permission-info/permission-info.component.ts
```diff
@@ -0,0 +1,184 @@
...
+        originY: 'bottom'
+    };
+
+    private _overlayPos: OverlayConnectionPosition = {
+        overlayX: 'end',
+        overlayY: 'top'
+    };
+
+    constructor(
+        // private _overlay: Overlay,
```
mdelez, line 93: this file could use some cleanup (removing commented out code)

src/app/workspace/resource/permission-info/permission-info.component.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    beforeEach(() => {
+        testHostFixture = TestBed.createComponent(TestHostComponent);
+        testHostComponent = testHostFixture.componentInstance;
+        testHostFixture.detectChanges();
+    });
+
+    it('should create', () => {
+        expect(testHostComponent.permissionInfoComponent).toBeTruthy();
```
mdelez, line 67: it might be nice to add a test to ensure the radio boxes are checked/unchecked correctly

mdelez (changes requested):
Adding the test isn't crucial but it may be nice to have if we're going to be working with permissions at a later point in time so we know that this new component still works

src/app/workspace/resource/permission-info/permission-info.component.ts
kilchenmann: oops...sure thing &rarr; resolved in fc86cd9b

src/app/workspace/resource/permission-info/permission-info.component.spec.ts
kilchenmann: resolved in 8a022666

mdelez (approved):
Thanks!!

PR #666, opened 2022-02-17
fix(resource): resource viewer is broken if fileRepresentation has restricted view (DEV-455)

kilchenmann:
resolves DEV-455

Resource viewer (and app) is broken if user doesn't have the necessary permission on still image file representation. In this case the representationsToDisplay fileValue is undefined. This PR resolves this issue.


mdelez (approved):
thanks!!

PR #667, opened 2022-02-18
fix(permission-info): better permission string handling (DEV-537)

kilchenmann:
resolves DEV-537


mdelez (approved):
thanks for adding the comments explaining how it works :)

PR #671, opened 2022-02-22
fix(permission-info): close permission info on scrolling (DEV-552)

kilchenmann:
The cdk-overlay-scrollstrategy was not working correctly. The `reposition` attribute was not useful. With this PR it will close the info box as soon the user scrolls (in Safari). There is one bad behaviour which I was not able to resolve. The the scroll event closes the info box, the user has to click twice on the permission button, because the `isOpen` status is still set to true. I couldn't find a good example to resolve this.


kilchenmann:
thank you!

PR #673, opened 2022-02-23
fix(error): display 500 server error when the api is not healthy (DEV-475)

kilchenmann:
resolves DEV-475


kilchenmann:
@mdelez wait with the review... these changes needs latest version of js-lib

mdelez (approved):
thanks!

PR #677, opened 2022-02-28
feat(archive-value): Download archive files with original filename (DEV-295)

mdelez:
resolves DEV-295


mdelez:
@subotic I've tested this with the main branch of DSP-API. Is there a release of DSP-API that implements the changes you made to SIPI?

PR #680, opened 2022-03-11
chore(ontology): adjust tooltip text in app when property is inherited (DEV-323)

kilchenmann:
resolves DEV-323


kilchenmann:
It's already implemented

PR #690, opened 2022-03-22
feat(still-image): IIIF URL copy button (DEV-524)

mdelez:
resolves DEV-524

### DIFF
--- a/src/app/workspace/resource/representation/still-image/still-image.component.html
+++ b/src/app/workspace/resource/representation/still-image/still-image.component.html
@@ -6,7 +6,15 @@
                 <span class="caption mat-caption">{{imageCaption}}</span>
             </div>
             <div class="bottom">
-                <a href="{{iiifUrl}}" target="_blank" class="iiif mat-caption">{{iiifUrl}}</a>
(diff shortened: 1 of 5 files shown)


kilchenmann (changes requested):
We should also use the notification service which pops up after clicking on the copy button. Same way as we did in the "Copy ARK URL".

mdelez:
> 25bc724

done in [25bc724](https://github.com/dasch-swiss/dsp-app/pull/690/commits/25bc7244848f2da86462294c4a08aad25c97848d)

mdelez:
> We should also use the notification service which pops up after clicking on the copy button. Same way as we did in the "Copy ARK URL".
> 
> Have a look at the `properties.component.html` where we have a click event `(click)="openSnackBar('ARK URL copied to clipboard!')"` and the corresponding `openSnackBar` method in `properties.component.ts`.

done in [25bc724](https://github.com/dasch-swiss/dsp-app/pull/690/commits/25bc7244848f2da86462294c4a08aad25c97848d)

kilchenmann (approved):

PR #703, opened 2022-04-06
fix(region): highlight info of clicked region (DEV-724)

kilchenmann:
resolves DEV-724


src/app/workspace/resource/resource.component.ts
```diff
@@ -584,12 +591,16 @@ export class ResourceComponent implements OnInit, OnChanges, OnDestroy {
...
+            region.scrollIntoView({
+                behavior: 'smooth',
+                block: 'center'
+            });
         }
 
     }
 
     updateRegions(iri: string) {
+        console.log('res component: updateRegions', iri);
```
mdelez, line 603: your turn to remove console logs :)

src/app/workspace/resource/representation/still-image/still-image.component.ts
```diff
@@ -684,11 +677,34 @@ export class StillImageComponent implements OnChanges, OnDestroy {
         loc.y = loc.y * aspectRatio;
 
         this._viewer.addOverlay({
-            element: elt,
-            location: loc
+            element: regEle,
+            location: loc,
+        });
+
+        const tracker = new OpenSeadragon.MouseTracker({
```
mdelez, line 684: I would add a comment explaining why this is necessary since I can see someone deleting this is the future because it doesn't appear to do anything, especially since the function is empty

mdelez (changes requested):

src/app/workspace/resource/resource.component.ts
kilchenmann: damn 🙈 
Removed in b9975183

src/app/workspace/resource/representation/still-image/still-image.component.ts
kilchenmann: ah yes, we talked about.
Added in 6a830998

mdelez (approved):
works great, thanks!

PR #722, opened 2022-05-03
refactor: clean up jsDocs and delete deprecated methods (DEV-741)

kilchenmann:
resolves DEV-741

### DIFF
--- a/src/app/workspace/search/search-panel/search-panel.component.ts
+++ b/src/app/workspace/search/search-panel/search-panel.component.ts
@@ -26,14 +26,6 @@ export class SearchPanelComponent implements OnInit {
     @Input() projectfilter?: boolean = false;
 
     /**
-     * @deprecated Use `limitToProject` instead
-     *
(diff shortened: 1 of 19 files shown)


src/app/workspace/search/search-panel/search-panel.component.ts
```diff
@@ -76,10 +68,7 @@ export class SearchPanelComponent implements OnInit {
     ) { }
 
     ngOnInit() {
```
mdelez, line 70: do we still need `ngOnInit` here?

mdelez (approved):
Thanks for this! Just a small remark

src/app/workspace/search/search-panel/search-panel.component.ts
kilchenmann: I think we have it everywhere, even when it's empty. Should I remove all empty angular lifecycle hooks?

PR #725, opened 2022-05-04
feat(resource): better progress indicator handling in resource viewer (DEV-638)

kilchenmann:
resolves DEV-638

Especially in case of a book, where the incoming file representations have to be loaded first, it makes sense to display the whole resource when everything is ready.

https://user-images.githubusercontent.com/4253580/166705598-b29e402b-4762-4528-9877-36e99d54ea26.mov


mdelez (approved):
Looks good :)

PR #727, opened 2022-05-06
refactor(message): use only one static error component (DEV-900)

kilchenmann:
resolves DEV-900


kilchenmann:
@mdelez would it make sense to rename the error-component into status-component, because it's not only used for errors. But we can discuss it on Monday.

mdelez:
> @mdelez would it make sense to rename the error-component into status-component, because it's not only used for errors. But we can discuss it on Monday.

maybe `app-status` would be more clear since "status" can be a bit ambiguous

kilchenmann:
@mdelez the PR should be ready for review now

src/app/main/status/status.component.scss
```diff
@@ -12,24 +12,27 @@
...
+    width: 480px;
+    margin: 0 auto;
 
     .image,
     .text {
       justify-self: stretch;
       align-self: stretch;
       box-sizing: border-box;
-      padding: 24px;
+      // padding: 24px;
```
mdelez, line 23: can this be removed?

src/app/user/user-form/user-form.component.ts
```diff
@@ -384,33 +381,35 @@ export class UserFormComponent implements OnInit, OnChanges {
...
+                        session.user.lang = this.userForm.controls['lang'].value;
                         localStorage.setItem('session', JSON.stringify(session));
                     }
 
                     this._cache.set(this.username, response);
 
-                    this.success = true;
-
+                    // this.loading = false;
+                    this._notification.openSnackBar('You have successfully updated user\'s profile data.');
```
mdelez, line 391: **the** user's profile data

mdelez (approved):
Thank you for this! A few remarks but then it can be merged.

src/app/user/user-form/user-form.component.ts
kilchenmann: resolved in a3fdd232

PR #733, opened 2022-05-11
feat(onto-editor) delete resource class and property comments (DEV-698)

mdelez:
resolves DEV-698


mdelez:
note to self: replace the local js-lib version with the new js-lib version once it's released

package-lock.json
```diff
@@ -72,6 +72,20 @@
...
+      "dependencies": {
+        "@babel/helper-compilation-targets": "^7.16.7",
+        "@types/jsonld": "^1.5.6",
+        "json2typescript": "1.4.1",
+        "jsonld": "^5.2.0"
+      },
+      "peerDependencies": {
+        "rxjs": "6.x"
+      }
+    },
```
kilchenmann, line 88: the yalc package shouldn't be here anymore

kilchenmann (approved):
I haven't tested it, but the code looks fine.

package-lock.json
mdelez: hmm, I'll have a look. I thought removing the yalc package and adding the real package would be enough to get rid of it in the lock file.

mdelez: I removed it manually in [fdb4118](https://github.com/dasch-swiss/dsp-app/pull/733/commits/fdb411808a93ab616d71ec4e3e7539ed73d81704)

PR #734, opened 2022-05-12
feat(hint): display hint about fulltext search (DEV-901)

kilchenmann:
resolves DEV-901


kilchenmann:
Screen Shot 2022-05-12 at 11 25 17

There's a new hint.component where we define tipps and tricks. In the first iteration it contains the hint for the search syntax and it can be expanded in the future. Maybe we can grab hints from a specific definition file maybe written in markdown.
In this PR the hint component is used in the search-panel and in the results page if none resource was found.

mdelez (approved):
Works well, thanks!

PR #737, opened 2022-05-13
feat(ontology): select link class from all ontologies (DEV-688)

kilchenmann:
resolves DEV-688


kilchenmann:
@mdelez you'll find the solution (how it should look like) already in the corresponding jira-subtask

mdelez (approved):
works well, thanks!

PR #738, opened 2022-05-13
refactor(project): resolve issues in lists and with status 204

kilchenmann:
resolves DEV-926 and resolves issue with required comment in list root node


mdelez (approved):
thanks for this!

PR #740, opened 2022-05-16
fix(ontology): unsupported property type was displayed wrong (DEV-936)

kilchenmann:
resolves DEV-936

### DIFF
--- a/src/app/project/ontology/ontology.service.ts
+++ b/src/app/project/ontology/ontology.service.ts
@@ -133,11 +133,6 @@ export class OntologyService {
     getDefaultPropType(property: ResourcePropertyDefinitionWithAllLanguages): Observable<DefaultProperty> {
         let propType: DefaultProperty;
 
-        if (!property.guiElement) {
-            // we don't know what element to use, so it's unsupported property
-            return of (DefaultProperties.unsupported);
-        }
-
         for (const group of this.defaultProperties) {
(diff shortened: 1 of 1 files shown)


src/app/project/ontology/ontology.service.ts
```diff
@@ -148,20 +143,23 @@ export class OntologyService {
...
+                        } else if (property.objectType === Constants.TextValue) {
+                            // if the property is of type text value, we have to check the gui element
+                            // to get the correct default prop params
                             propType = (group.elements.find(i =>
-                                i.objectType === property.objectType && i.subPropOf === Constants.SeqNum
+                                i.guiEle === property.guiElement && i.objectType === property.objectType
                             ));
                         } else {
+                            // in all other cases the gui-element resp. the subProp is not relevant
+                            // because the object type is unique
```
Vijeinath, line 160: Small observation: In line 139/140 you put the comment before the _if_ structure and here after the _if_ and _else_

Vijeinath (approved):
LGTM

src/app/project/ontology/ontology.service.ts
kilchenmann: Good point. It's always the question, where to put the comment. Normally we put it outside the if statement. But in this long list of `if`, `else if`, `else` I thought it makes sense to have it inside.

PR #741, opened 2022-05-16
fix(representation): resolve timeline issue when resizing window (DEV-819)

kilchenmann:
resolves DEV-819


kilchenmann:
Not the best solution (because we do not listen when the resize event starts, so the drag-drop-circle has a wrong position during the window resize resp. split area resize process), but at the end it works:

https://user-images.githubusercontent.com/4253580/169292091-a2868896-db7d-4ab0-b609-412e4d752b5c.mov

Vijeinath (approved):
Nothing to complain

PR #751, opened 2022-05-25
feat(text-file): add support for text file representations (DEV-920)

mdelez:
resolves DEV-920

### DIFF
--- a/src/app/workspace/resource/representation/upload/upload.component.ts
+++ b/src/app/workspace/resource/representation/upload/upload.component.ts
@@ -52,7 +52,7 @@ export class UploadComponent implements OnInit {
     supportedAudioTypes = ['audio/mpeg'];
     supportedVideoTypes = ['video/mp4'];
     supportedArchiveTypes = ['application/zip', 'application/x-tar', 'application/gzip'];
-    supportedTextTypes = ['application/csv', 'application/xml', 'text/plain', 'text/xml'];
+    supportedTextTypes = ['application/csv', 'application/xml', 'text/csv', 'text/plain', 'text/xml'];
 
     constructor(
         private _fb: FormBuilder,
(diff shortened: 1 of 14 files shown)


src/app/workspace/resource/representation/upload/upload.component.ts
```diff
@@ -50,6 +52,7 @@ export class UploadComponent implements OnInit {
     supportedAudioTypes = ['audio/mpeg'];
     supportedVideoTypes = ['video/mp4'];
     supportedArchiveTypes = ['application/zip', 'application/x-tar', 'application/gzip'];
+    supportedTextTypes = ['application/csv', 'application/xml', 'text/plain', 'text/xml'];
```
mdelez, line 55: Since we support both `application/xml` and `text/xml`, the upload component has the unfortunate consequence of listing `xml` twice. I could add a workaround in the html like @kilchenmann did for the audio files.
Screen Shot 2022-05-25 at 13 18 36

kilchenmann: This is a good question. I suggest to have it somehow hardcoded similar to the audio format.

kilchenmann: We should display format, that the user understands. There should be the file extension depending on the mimetype. The question is, what does `plain` contain? Nobody will understand the `plain` text format.

mdelez: That's a fair point. Sometimes the MIME types don't really match up with the file extensions. So far they've worked out in our favor but not in this case. I'll hardcode it for now and write a comment. I don't really see a point in creating a converter since we don't use these strings anywhere else to my knowledge.

kilchenmann: Yes, this is why I had to hardcode it in the audio upload.

kilchenmann (approved):
LGTM. Thanks 🚀

PR #755, opened 2022-05-31
feat(resource): create new resource instance directly from ontology's class (DEV-959)

kilchenmann:
resolves DEV-959

### DIFF
--- a/src/app/project/ontology/resource-class-info/resource-class-info.component.ts
+++ b/src/app/project/ontology/resource-class-info/resource-class-info.component.ts
@@ -10,6 +10,7 @@ import {
     KnoraApiConnection,
     PropertyDefinition,
     ReadOntology,
+    ReadProject,
     ResourceClassDefinitionWithAllLanguages,
     ResourcePropertyDefinitionWithAllLanguages,
     UpdateOntology,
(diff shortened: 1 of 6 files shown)


src/app/project/ontology/resource-class-info/resource-class-info.component.ts
```diff
@@ -514,4 +515,32 @@ OFFSET 0`;
...
+            width: '840px',
+            maxHeight: '80vh',
+            position: {
+                top: '112px'
+            },
+            data: { id: iri, mode: 'createResource', project: projectIri, title: label,  subtitle: 'Set the property values of the resource' },
+            disableClose: true
+        };
+
+        const dialogRef = this._dialog.open(
```
mdelez, line 538: If you don't have any future plans for anything happening after the dialog is closed, you don't need to assign this to a `const` and can remove the logic below for doing something after the dialog is closed. But it's fine if we keep it as well. Maybe we'll need it in the future.

mdelez (approved):
works well :) I left a comment but it's up to you if you want to keep or remove the code.

src/app/project/ontology/resource-class-info/resource-class-info.component.ts
kilchenmann: Good point. Thanks. But I think I'll keep it for the moment. Just copied from other components.

PR #764, opened 2022-06-13
fix(docs): resolve dead links to documentation

kilchenmann:
Since I've updated the docs.dasch.swiss page with versioning, we have dead links in DSP-App.


mdelez (approved):
thanks :)

PR #766, opened 2022-06-15
fix(link-value): list resource classes from all ontologies in project (DSP-1034)

mdelez:
resolves DEV-1034


kilchenmann (approved):
LGTM. Thanks

PR #769, opened 2022-06-22
docs: Update the user guide (DEV-1049)

flavens:
resolves DEV-1049


mdelez (approved):

flavens:
@mdelez I may have missed some things. Please, let me know!

flavens:
@mdelez concerning the screenshot with the classes Buch/Buchseite, my user account is in English and we always get the label in German (why?!) so I cannot change it now.

flavens:
> > @mdelez concerning the screenshot with the classes Buch/Buchseite, my user account is in English and we always get the label in German (why?!) so I cannot change it now.
> 
> hmm strange...is there an english label? Otherwise you could just make the resource again just for the screenshot. It's not a big issue but consistency is nice :)

The thing is: even when we have English and German labels, the German one is displayed by default. Like the list label order issue, you fixed... Should I report it? I did a test locally and German labels out!

PR #777, opened 2022-07-20
feat(overview): add overview and project tile components (DEV-984)

mdelez:
resolves DEV-984

### DIFF
--- a/src/app/project/project.component.ts
+++ b/src/app/project/project.component.ts
@@ -222,6 +222,11 @@ export class ProjectComponent implements OnInit {
         }
     }
 
+    // go to overview page
+    goToOverview() {
+        this._router.navigate(['/overview'], { relativeTo: this._route });
+    }
+
     /**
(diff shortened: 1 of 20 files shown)


src/app/project/project.component.ts
```diff
@@ -222,6 +222,11 @@ export class ProjectComponent implements OnInit {
         }
     }
 
+    // go to overview page
```
Vijeinath, line 225: Because this comment refers to the function its more common to do the "/** [...] */" comment

Vijeinath (approved):
LGTM! Just keep in mind that I have never implemented tests in angular.

EricSommerhalder (approved):
Perfect thanks

src/app/project/project.component.ts
mdelez: true true, changed in [2832fce](https://github.com/dasch-swiss/dsp-app/pull/777/commits/2832fce9f9eb0e95528ce0c04e934986f136a34b)

PR #783, opened 2022-07-29
fix(property-form): App changes "salsah-gui:List" to "salsah-gui:Pulldown" (DEV-1089)

mdelez:
resolves DEV-1089

The property form currently calls the `replaceGuiElement()` method any time a label or comment on a property is changed and since `salsah-gui:List` is no longer supported in the app as of https://github.com/dasch-swiss/dsp-app/pull/643, the `guiElement` of ListValue object types was always changed to `salsah-gui:Pulldown`. However, in the [API](https://docs.dasch.swiss/2022.07.02/DSP-API/03-apis/api-v2/ontology-information/#changing-the-gui-element-and-gui-attributes-of-a-property) (not specifically stated in the docs, see screenshot below), it's only possible to change the `guiElement` of a TextValue object type so we should only call `replaceGuiElement()` if the property is a TextValue and the `guiElement` is different from it's initial value.

screenshot from old Youtrack task:
Screen Shot 2022-07-29 at 10 44 33

### DIFF
--- a/src/app/project/ontology/property-form/property-form.component.ts
+++ b/src/app/project/ontology/property-form/property-form.component.ts
@@ -455,7 +455,12 @@ export class PropertyFormComponent implements OnInit {
                                 (propertyCommentResponse: ResourcePropertyDefinitionWithAllLanguages) => {
                                     this.lastModificationDate = propertyCommentResponse.lastModificationDate;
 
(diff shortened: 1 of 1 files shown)


Vijeinath (approved):
LGTM

PR #789, opened 2022-08-05
feat(document): changes to pdf viewer (DEV-1149)

Vijeinath:
resolves DEV-1149


mdelez (changes requested):
Works well, thanks! I just have a small request to add a call to `_getOriginalFilename()` in the `_replaceFile` method similar to how it's done in the archive component. This wasn't necessary before because we didn't allow the user to download the file but now we need to ensure the filename is updated when replacing the file.

mdelez (approved):

PR #793, opened 2022-08-09
fix(resource view): dsp app the resource viewer does not reload properly after deselecting a resource from the comparison viewer(Dev-1123)

domsteinbach:
resolves DEV-1123


mdelez (approved):
works well, thanks!

PR #794, opened 2022-08-09
fix(notification): Notifications only appearing for a split second (DEV-1093)

mdelez:
resolves DEV-1093


EricSommerhalder (approved):
I'm amazed how you kept overview with this complexity. Maybe in the future you can split these in two prs? LGTM though.

PR #796, opened 2022-08-11
feat(audio): Make changes to audio component (DEV-1148)

EricSommerhalder:
resolves DEV-1148


mdelez (approved):
works well, thanks :)

PR #797, opened 2022-08-12
style(file-representations): adjust styling of file representation components (DEV-1186)

mdelez:
resolves DEV-1186


mdelez:
This is a rough PR due to the need for `::ng-deep` to match the designs. If anyone wants to try to find a way around using `::ng-deep`, go ahead :)

EricSommerhalder (approved):
Works for me and looks good. As a general remark it seems weird/not very DRY to me to have so many very similar css files for each type of representation. Don't know a better solution though :)

PR #799, opened 2022-08-16
feat(ontology): ontology editor read mode (DEV-1183)

mdelez:
resolves DEV-1183


src/app/project/project.component.html
```diff
@@ -65,10 +65,11 @@
                         <mat-expansion-panel [expanded]="first">
                             <mat-expansion-panel-header>
                                 <div class="sidenav-panel-header"
-                                    (click)="projectAdmin ? $event.stopPropagation() : true">
+                                    (click)="projectMember ? $event.stopPropagation() : true">
```
Vijeinath, line 68: I didn't know that you can just pass a boolean instead of executing a function at a click event

Vijeinath (approved):
Looks good

PR #805, opened 2022-08-23
feat(list-value): show hierarchy of list node (DEV-1092)

mdelez:
resolves DEV-1092


EricSommerhalder (approved):
LGTM, works like intended

PR #808, opened 2022-08-26
refactor(upload): use file extensions for valid file types instead of MIME types (DEV-1203)

mdelez:
resolves DEV-1203


mdelez:
@EricSommerhalder @domsteinbach I changed how we check if a file type is allowed. It now checks the file extension instead of the MIME type. I would be grateful if you could also test this on your end and make sure all the file uploads work. You don't actually need to create a resource instance for each file, just check if the upload work within the upload component when you create a new resource instance.

Here are the test files:
[dsp-tools-test-data.zip](https://github.com/dasch-swiss/dsp-app/files/9443655/dsp-tools-test-data.zip)

src/app/workspace/resource/representation/upload/upload.component.ts
```diff
@@ -46,13 +46,13 @@ export class UploadComponent implements OnInit {
...
-    // todo: maybe we can use this list to display which file format is allowed to
-    supportedImageTypes = ['image/jpeg', 'image/jp2', 'image/tiff', 'image/tiff-fx', 'image/png'];
-    supportedDocumentTypes = ['application/pdf'];
-    supportedAudioTypes = ['audio/mpeg'];
-    supportedVideoTypes = ['video/mp4'];
-    supportedArchiveTypes = ['application/zip', 'application/x-tar', 'application/gzip'];
-    supportedTextTypes = ['application/csv', 'application/xml', 'text/csv', 'text/plain', 'text/xml'];
+
+    supportedAudioTypes = ['mp3', 'wav'];
+    supportedArchiveTypes = ['zip', 'tar', 'gz', 'gzip', '7z', 'tgz', 'z'];
```
domsteinbach, line 51: For all supported types: Form what the order of the types is deriving from? For my taste it would be nice to have them ordered alphabetically, hard coded and/or via .sort()'

domsteinbach (changes requested):
The ordering of the supported types is the only thing I like to address.

src/app/workspace/resource/representation/upload/upload.component.ts
mdelez: it's the order that's defined in the API so that I could make sure that I included everything easily but I can sort them alphabetically

https://github.com/dasch-swiss/dsp-api/blob/main/sipi/scripts/file_info.lua

mdelez: done in [635ae3a](https://github.com/dasch-swiss/dsp-app/pull/808/commits/635ae3a38c575c9b6ecc923fe4a096459eae349f)

EricSommerhalder (approved):
LGTM

PR #813, opened 2022-09-01
fix(resource): compound resource with restricted resources infinite loading (DEV-1114)

mdelez:
resolves DEV-1114

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -122,6 +122,22 @@
                             "vendorChunk": false,
                             "buildOptimizer": false,
                             "budgets": []
+                        },
+                        "staging-server": {
+                            "fileReplacements": [
+                                {
+                                    "replace": "src/environments/environment.ts",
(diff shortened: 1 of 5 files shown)


mdelez:
@EricSommerhalder it's a one-line fix. The other files are the setup to run the app using the staging server. I think it'll be useful to have the staging-server config in case we need to investigate other issues Lausanne may have.

To test, you can go to these two links while logged out:
**This resource will not exist after today as the staging db is reset every night**

Actual staging server:
https://admin.staging.dasch.swiss/resource/0118/pf0Ll0c4QeG8Oypb8_0R_A -> **endless loading**

Local instance connected to staging server via `npm run start-with-staging-server`: 
http://0.0.0.0:4200/resource/0118/pf0Ll0c4QeG8Oypb8_0R_A -> **you can see the resource without the image**

EricSommerhalder (approved):

PR #815, opened 2022-09-05
fix(int-value): fix int-value validation (DEV-1277)

mdelez:
resolves DEV-1277


src/app/workspace/resource/values/int-value/int-value.component.ts
```diff
@@ -105,7 +105,7 @@ export class IntValueComponent extends BaseValueDirective implements OnInit, OnC
 
         updatedIntValue.id = this.displayValue.id;
 
-        updatedIntValue.int = this.valueFormControl.value;
+        updatedIntValue.int = parseInt(this.valueFormControl.value, 10);
```
Vijeinath, line 108: look above

Vijeinath (approved):
LGTM

PR #818, opened 2022-09-06
fix(resource): fix properties sorting (DEV-1204)

mpro7:
resolves DEV-1204

### DIFF
--- a/src/app/workspace/resource/resource.component.ts
+++ b/src/app/workspace/resource/resource.component.ts
@@ -1,5 +1,4 @@
 /* eslint-disable @typescript-eslint/member-ordering */
-import { A } from '@angular/cdk/keycodes';
 import { Component, Inject, Input, OnChanges, OnDestroy, OnInit, SimpleChange, ViewChild } from '@angular/core';
 import { MatTabChangeEvent } from '@angular/material/tabs';
 import { Title } from '@angular/platform-browser';
@@ -31,7 +30,6 @@ import {
     SystemPropertyDefinition
 } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
-import { delay } from 'rxjs/operators';
 import { DspApiConnectionToken } from 'src/app/main/declarations/dsp-api-tokens';
 import { ErrorHandlerService } from 'src/app/main/services/error-handler.service';
 import { NotificationService } from 'src/app/main/services/notification.service';
@@ -321,11 +319,8 @@ export class ResourceComponent implements OnInit, OnChanges, OnDestroy {
                             this.requestIncomingResources(this.resource);
                         }
 
-                        setTimeout(() => {
-                            // gather resource property information
-                            res.resProps = this.initProps(response);
-                        }, 2000);
-
+                        // gather resource property information
+                        res.resProps = this.initProps(response);
 
                         // gather system property information
                         res.systemProps = this.resource.res.entityInfo.getPropertyDefinitionsByType(SystemPropertyDefinition);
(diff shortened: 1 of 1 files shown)


src/app/workspace/resource/resource.component.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/member-ordering */
+import { A } from '@angular/cdk/keycodes';
```
mdelez, line 2: can be removed, probably an auto-import

src/app/workspace/resource/resource.component.ts
```diff
@@ -30,6 +31,7 @@ import {
     SystemPropertyDefinition
 } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
+import { delay } from 'rxjs/operators';
```
mdelez, line 34: can be removed, probably an auto-import

src/app/workspace/resource/resource.component.ts
```diff
@@ -319,8 +321,11 @@ export class ResourceComponent implements OnInit, OnChanges, OnDestroy {
                             this.requestIncomingResources(this.resource);
                         }
 
-                        // gather resource property information
-                        res.resProps = this.initProps(response);
+                        setTimeout(() => {
```
mdelez, line 324: Why does this need to be wrapped in a `setTimeout`?

mdelez (changes requested):

src/app/workspace/resource/resource.component.ts
mpro7: Sorry, all of these you've find are leftovers I forgot to remove. Haven't work with TS lately and my IDE is not removing unused imports though.

mdelez:
@mpro7 The `if` condition in the second `sort` seems to never be met so I still have the issue on my end. Is it working for you?

mpro7:
> @mpro7 The `if` condition in the second `sort` seems to never be met so I still have the issue on my end. Is it working for you?

It's working for me. Why do you say the condition it's never to be met?

mdelez:
> > @mpro7 The `if` condition in the second `sort` seems to never be met so I still have the issue on my end. Is it working for you?
> 
> It's working for me. Why do you say the condition it's never to be met?

The account I'm using (root) has german set as its language so the label is "hat Standoff Link zu" which is the reason why the `if` statement is never met. Also, this would mean that if anyone were to put "has" in their property label (which is pretty common unfortunately, especially for Lausanne) it would also get sorted to the bottom. 

The biggest issue I see is that these default props do not have a `guiOrder`. I took a look at a more complex resource to see what else we can sort by. I think we're safe to either sort all properties that have a `undefined` `guiOrder` to the bottom **or** we can sort by `isEditable` but I'm not certain under what conditions this boolean is set.

mpro7:
> > > @mpro7 The `if` condition in the second `sort` seems to never be met so I still have the issue on my end. Is it working for you?
> > 
> > 
> > It's working for me. Why do you say the condition it's never to be met?
> 
> The account I'm using (root) has german set as its language so the label is "hat Standoff Link zu" which is the reason why the `if` statement is never met. Also, this would mean that if anyone were to put "has" in their property label (which is pretty common unfortunately, especially for Lausanne) it would also get sorted to the bottom.
> 
> The biggest issue I see is that these default props do not have a `guiOrder`. I took a look at a more complex resource to see what else we can sort by. I think we're safe to either sort all properties that have a `undefined` `guiOrder` to the bottom **or** we can sort by `isEditable` but I'm not certain under what conditions this boolean is set.

Hmm, the language factor I haven't took into the consideration. That's why the property labels should be unified and in English only ;)

Anyways, I will try to change the condition and see how if it works with `undefined`. It seems to be the most reasonable.

mdelez (approved):
works well, thanks :)

PR #825, opened 2022-09-15
feat(lists): read mode project member (DEV-1343)

Vijeinath:
resolves DEV-1343
The lists (within the list tab and on the sidenav) are visible for project member but all the CRUD operations regarding lists were hidden.

### DIFF
--- a/src/app/project/list/list-item-form/list-item-form.component.ts
+++ b/src/app/project/list/list-item-form/list-item-form.component.ts
@@ -98,9 +98,6 @@ export class ListItemFormComponent implements OnInit {
     sysAdmin = false;
     projectAdmin = false;
 
-    // project data
-    project: ReadProject;
-
     loading: boolean;
 
     initComponent: boolean;
@@ -127,10 +124,9 @@ export class ListItemFormComponent implements OnInit {
         // get the project data from cache
         this._cache.get(this.projectCode).subscribe(
             (response: ReadProject) => {
-                this.project = response;
 
                 // is logged-in user projectAdmin?
-                this.projectAdmin = this.sysAdmin ? this.sysAdmin : this.session.user.projectAdmin.some(e => e === this.project.id);
+                this.projectAdmin = this.sysAdmin ? this.sysAdmin : this.session.user.projectAdmin.some(e => e === response.id);
             },
             (error: ApiResponseError) => {
                 this._errorHandler.showMessage(error);
--- a/src/app/project/list/list-item-form/list-item-form.component.ts
+++ b/src/app/project/list/list-item-form/list-item-form.component.ts
@@ -11,11 +11,14 @@ import {
     ListInfoResponse,
     ListNode,
     ListNodeInfoResponse,
+    ReadProject,
     StringLiteral
 } from '@dasch-swiss/dsp-js';
 import { DspApiConnectionToken } from 'src/app/main/declarations/dsp-api-tokens';
 import { DialogComponent } from 'src/app/main/dialog/dialog.component';
 import { ErrorHandlerService } from 'src/app/main/services/error-handler.service';
+import { Session, SessionService } from '../../../main/services/session.service';
+import { CacheService } from '../../../main/cache/cache.service';
 
 export class ListNodeOperation {
     operation: 'create' | 'insert' | 'update' | 'delete' | 'reposition';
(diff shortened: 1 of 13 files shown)


src/app/project/list/list-item/list-item.component.html
```diff
@@ -2,28 +2,28 @@
     <div class="list-node" *ngFor="let node of list; let first = first; let last = last;">
 
         <!-- button to expand / close node -->
-        <button type="button" mat-icon-button (click)="toggleChildren(node.id)" class="">
+        <button *ngIf="node.children.length !== 0" type="button" mat-icon-button (click)="toggleChildren(node.id)">
```
mdelez, line 5: This causes the arrow to be hidden even for project admins. The arrow needs to be shown for project admins so that they can add child node even if one doesn't already exist.

mdelez (changes requested):
In addition to the other comments, the indentation needs to be worked on a bit. I think it's tricky because you have to account for adding extra space if the expand/collapse button is hidden.

Screen Shot 2022-09-16 at 15 20 48
Screen Shot 2022-09-16 at 15 21 27

src/app/project/list/list-item/list-item.component.html
Vijeinath: Done in https://github.com/dasch-swiss/dsp-app/pull/825/commits/cd871506a5efe3641da19242f9ba711f17d10a68

src/app/project/list/list-item/list-item.component.html
```diff
@@ -2,28 +2,28 @@
...
             <mat-icon class="mat-icon-rtl-mirror">
                 {{showChildren(node.id) ? 'expand_more' : 'chevron_right'}}
             </mat-icon>
         </button>
 
+        <div class="list-item-indentation" *ngIf="node.children.length === 0"></div>
```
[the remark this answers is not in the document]
Vijeinath: I tried with the class condition but couldn't do it. So I changed the whole situation by adding a div around the button with a certain size -> https://github.com/dasch-swiss/dsp-app/pull/825/commits/cd871506a5efe3641da19242f9ba711f17d10a68

src/app/project/list/list-item-form/list-item-form.component.ts
```diff
@@ -90,6 +93,14 @@ export class ListItemFormComponent implements OnInit {
 
     @Output() refreshParent: EventEmitter<ListNodeOperation> = new EventEmitter<ListNodeOperation>();
 
+    // permissions of logged-in user
+    session: Session;
+    sysAdmin = false;
+    projectAdmin = false;
+
+    // project data
+    project: ReadProject;
```
mdelez, line 102: I don't think this variable is needed, you can use a `const` to temporarily store the project you get from the cache

src/app/project/list/list-item-form/list-item-form.component.ts
```diff
@@ -101,10 +112,30 @@ export class ListItemFormComponent implements OnInit {
...
+        // get information about the logged-in user
+        this.session = this._session.getSession();
+
+        // is the logged-in user system admin?
+        this.sysAdmin = this.session.user.sysAdmin;
+
+        // get the project data from cache
+        this._cache.get(this.projectCode).subscribe(
+            (response: ReadProject) => {
+                this.project = response;
```
mdelez, line 130: You can use a `const` here instead

src/app/project/list/list-item/list-item.component.ts
```diff
@@ -36,14 +39,43 @@ export class ListItemComponent implements OnInit {
 
     @Output() refreshChildren: EventEmitter<ListNode[]> = new EventEmitter<ListNode[]>();
 
+    // permissions of logged-in user
+    session: Session;
+    sysAdmin = false;
+    projectAdmin = false;
+
+    // project data
+    project: ReadProject;
```
mdelez, line 48: same thing here as in the `list-item-form` component

src/app/project/project.component.ts
```diff
@@ -26,8 +26,10 @@ import { OntologyService } from './ontology/ontology.service';
     styleUrls: ['./project.component.scss']
 })
 export class ProjectComponent implements OnInit {
+    readonly TAB_DATA_MODEL = 3;
```
mdelez, line 29: nice!

mdelez (changes requested):
just a small thing and then I will approve :)

src/app/project/project.component.ts
Vijeinath: Thanks

src/app/project/list/list-item-form/list-item-form.component.ts
Vijeinath: You are right, it is not needed elsewhere

src/app/project/list/list-item/list-item.component.ts
Vijeinath: https://github.com/dasch-swiss/dsp-app/pull/825/commits/ff8b5a85e3d05f019d874791b5de77a05edb7572

mdelez (approved):
works well, thanks!

PR #830, opened 2022-09-21
chore: upgrade Angular and Material to v14 (DEV-1322)

mpro7:
resolves DEV-1322


.eslintrc.js
```diff
@@ -47,7 +47,6 @@ module.exports = {
                 "accessibility": "explicit"
             }
         ],
-        "@typescript-eslint/indent": "error",
```
mdelez, line 50: does this need to be removed?

mdelez (changes requested):
Are we able to upgrade core-js?

Screen Shot 2022-09-22 at 10 04 24

On the homepage, the project icons are cutoff now but this can be fixed by adding a height and width of 40px to the css for the icon in `project-tile` component

.eslintrc.js
mpro7: Well, I had issues with this rule which I couldn't manage to fix w/o turning it off. IMO all these rules should be revised. I'm not sure how happy you are with them, but I dislike some of. However this should be separate task.

About `core-js`. I'll try, but if this triggers another issue with peer deps will adjust the css. There is a separate task defined for deps maintanace.

mdelez: sorry, my comment was bit messy. The `core-js` issue doesn't have anything to do with the css of the icon, it was a separate remark.

mpro7:
@mdelez updated `js-core` to the newest version anyway and fixed the icon. I just see now that my VSC seems to be implementing different formatting rules. Should I bring back the formatting in that SCSS file?

mdelez (approved):

PR #832, opened 2022-09-22
fix: Ontologies displayed twice (DEV-1325)

EricSommerhalder:
resolves DEV-1325

### DIFF
--- a/src/app/project/ontology/ontology.component.ts
+++ b/src/app/project/ontology/ontology.component.ts
@@ -1,3 +1,4 @@
+import { O } from '@angular/cdk/keycodes';
 import { Component, HostListener, Inject, OnInit, ViewChild, ViewContainerRef } from '@angular/core';
 import { FormBuilder, FormControl, FormGroup } from '@angular/forms';
 import { MatDialog, MatDialogConfig } from '@angular/material/dialog';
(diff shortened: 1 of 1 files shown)


src/app/project/ontology/ontology.component.ts
```diff
@@ -382,7 +383,26 @@ export class OntologyComponent implements OnInit {
             (ontologies: ReadOntology[]) => {
                 // update current list of project ontologies
                 ontologies[ontologies.findIndex(onto => onto.id === ontology.id)] = ontology;
-                this._cache.set('currentProjectOntologies', ontologies);
+                // avoid duplicates
+                const temp: ReadOntology[] = [];
```
mdelez, line 387: I would rename these temp variables to something a bit more meaningful

src/app/project/ontology/ontology.component.ts
```diff
@@ -1,3 +1,4 @@
+import { O } from '@angular/cdk/keycodes';
```
mdelez, line 1: auto-import FTW?

mdelez (changes requested):

src/app/project/ontology/ontology.component.ts
EricSommerhalder: got to love it

mdelez (approved):
works well, thanks :)

PR #849, opened 2022-10-10
refactor(project workspace): Remove grid view from the project workspace (DEV-1318)

domsteinbach:
resolves DEV-1318


mdelez (changes requested):
The `resource-grid` component is mocked in the `list-view` component spec file. This should also be removed even though it's not causing an issue.

PR #853, opened 2022-10-12
refactor: replace substr with substring (DEV-1409)

mdelez:
resolves DEV-1409


mpro7 (approved):
Approved, unless these are not all occurrences ;)

mdelez:
> Approved, unless these are not all occurrences ;)

They are :)

PR #857, opened 2022-10-21
fix(ontology): list of class properties sometimes not loading correctly (DEV-1435)

mdelez:
resolves DEV-1435


Vijeinath (approved):
Worked for me!

PR #860, opened 2022-10-27
feat(link-value, search-link-value): add debouncing to search requests (DEV-1463)

mdelez:
resolves DEV-1463

### DIFF
--- a/src/app/workspace/resource/values/link-value/link-value.component.spec.ts
+++ b/src/app/workspace/resource/values/link-value/link-value.component.spec.ts
@@ -22,6 +22,8 @@ import { of } from 'rxjs';
 import { DspApiConnectionToken } from 'src/app/main/declarations/dsp-api-tokens';
 import { LinkValueComponent } from './link-value.component';
 
+const DEBOUNCING_VALUE = 400;
+
 /**
  * test host component to simulate parent component.
  */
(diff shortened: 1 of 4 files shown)


domsteinbach (approved):
LGTM. We could even go further and don't send a request at all if a letter is added and the latest letter equals the label of the first resource. In that case we need only to filter the already existing result set, i.e. the resources.

```
            if (this.resources[0]?.label && this.resources[0].label[searchTerm.length-1]
                && this.resources[0].label[searchTerm.length-1] === searchTerm[searchTerm.length-1]) {
                this.resources = this.resources.filter(
                    r => r.label[searchTerm.length-1] === searchTerm[-1] );
            } else { // send request
```

domsteinbach:
>
> Where do you imagine this code snippet to be placed? Inside the `searchByLabel` method?
> 

I thought so, yes. That way the user experience on typing & receiving updated menu entries is consistent. 

But we need the information on whether a letter was added or removed from the searchTerm. If one gets removed we need to make a new request (or store the old results ...) So either the information added/removed or the former length must be available.

AND: the snipped must compare & filter for the whole substring until searchTerm.length-1, not only the nth letter. Like this:

```
if ( letterHasBeenAdded && this.resources[0]?.label && this.resources[0].label[searchTerm.length-1]
                && this.resources[0].label[0:searchTerm.length-1] === searchTerm ) {
                this.resources = this.resources.filter( r => r.label[0:searchTerm.length-1] === searchTerm);
            } else { // send new request
```

Otherwise if a user is entering sth. like "Bart", receiving some resources, then changing lightning quick to "Bert", it could get tricked and will filter the old "Bart"-results" with "Bert".

PR #874, opened 2022-11-23
refactor(text-value-as-string): no longer account for richText gui attribute (DEV-1471)

mdelez:
resolves DEV-1471


BalduinLandolt (approved):
I have a hard time giving reasonable feedback on this. But from what I understand, it should be good.

I know there is more enough to refactor in the App. But imho the whole "text-value-as-string/html/xml" logic is far from ideal because it makes the app depend on what exactly the API returns, rather than trying to capture the actual logic of it. If you ever want to tackle that, I may have some ideas.

mdelez:
> I have a hard time giving reasonable feedback on this. But from what I understand, it should be good.
> 
> I know there is more enough to refactor in the App. But imho the whole "text-value-as-string/html/xml" logic is far from ideal because it makes the app depend on what exactly the API returns, rather than trying to capture the actual logic of it. If you ever want to tackle that, I may have some ideas.

yes, I also don't like how it's implemented but that's a task for future me 🙃

PR #879, opened 2022-12-02
Fix off by one error during calculation of number of lines in the matrix files

seakayone:
The number of lines should always be the next bigger integer not the closest smaller integer

resolves DEV-1513


EricSommerhalder:
The fix looks good, however we need to wait for Mike's branch to be merged to reproduce the error and see if the fix works.

EricSommerhalder (approved):
Perfect thank you!

PR #881, opened 2022-12-12
feat (still image component): view stays within image borders, instant drag, no delay, double click and flick enabled

domsteinbach:
Resolves DEV-1554.

Introduce new features:

- the viewers focus is limited to the image borders (no more zooming into black background). On zooming in and out the view got stuck sometimes and cut the image. The view now sticks to the images borders and centers back on zooming out.
- improve drag behaviour: direct/instant. No delay, no slow animations
- double click zooms in
- flick gestures are possible to drag the image


Vijeinath (approved):
LGTM

PR #887, opened 2023-01-05
fix: fix gravsearch results infinite loading (DEV-1541)

mpro7:
resolves DEV-1541

This is just small enhancement to distinguish between different states of no resources to display. The snack bar shows more precise info for the user that there is either nothing to display or he doesn't have the permission to see the resources.

To test it, must run JS-LIB in development mode via yalc from [this branch](https://github.com/dasch-swiss/dsp-js-lib/pull/480), or mian after the merge.


Vijeinath (approved):
LGTM

PR #892, opened 2023-01-16
fix(project): Prevent adding new data model for deactivated projects (DEV-1582)

EricSommerhalder:
resolves DEV-1582


mdelez (approved):
works well, thanks :)

PR #900, opened 2023-01-23
style(ontology editor): restyle ontology editor for cardinalities (DEV-1614)

domsteinbach:
resolves DEV-1614


mdelez (changes requested):
Thank you for this but the title still needs some adjustment. It should be `style` instead of `styling`. The docs make this a bit confusing. `styling` is the Github label you can optionally assign to the PR. Also, there cannot be a space between the type and the scope so it should be `style(ontology editor)`

mdelez (approved):

PR #904, opened 2023-01-30
feat(project): add button to collapse/expand sidenav (DEV-1535)

mdelez:
resolves DEV-1535


mdelez:
@EricSommerhalder @Vijeinath I added the `@HostListener` just for fun but I understand if we don't want it.

Vijeinath (commented):
Works well. One thing: According to our designs on figma the circle color is also blue.
Bildschirm­foto 2023-01-30 um 15 07 37

PR #912, opened 2023-02-07
refactor(resource-instance-form): refactor resource instance form (DEV-1460)

mdelez:
resolves DEV-1460

### DIFF
--- a/src/app/project/beta/ontology-classes/ontology-class-item/ontology-class-item.component.ts
+++ b/src/app/project/beta/ontology-classes/ontology-class-item/ontology-class-item.component.ts
@@ -1,4 +1,4 @@
-import { Component, Inject, Input, OnInit } from '@angular/core';
+import { Component, Inject, Input, OnDestroy, OnInit } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { ClassDefinition, KnoraApiConnection, CountQueryResponse, ApiResponseError, Constants } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
@@ -12,7 +12,7 @@ import { OntologyService } from 'src/app/project/ontology/ontology.service';
     templateUrl: './ontology-class-item.component.html',
     styleUrls: ['./ontology-class-item.component.scss']
 })
-export class OntologyClassItemComponent implements OnInit {
+export class OntologyClassItemComponent implements OnInit, OnDestroy {
     readonly MAX_LABEL_CHAR = 25;
 
     @Input() resClass: ClassDefinition;
@@ -72,6 +72,12 @@ export class OntologyClassItemComponent implements OnInit {
                 this._getSearchCount();
             }
         ));
+    }
+
+    ngOnDestroy(): void {
+        if (this.componentCommsSubscriptions !== undefined) {
+            this.componentCommsSubscriptions.forEach(sub => sub.unsubscribe());
+        }
     }
 
     open(route: string) {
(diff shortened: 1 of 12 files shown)


src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
```diff
@@ -96,96 +69,102 @@ export class ResourceInstanceFormComponent implements OnInit, OnChanges, OnDestr
...
         private _route: ActivatedRoute,
         private _router: Router,
-        private _sortingService: SortingService
-    ) {
-        // get feature toggle information if url contains beta
-        if(this._route.parent) {
-            this.beta = (this._route.parent.snapshot.url[0].path === 'beta');
-        }
-    }
+        private _componentCommsService: ComponentCommunicationEventService,
```
Vijeinath, line 79: Shouldn't be "_componentCommService" instead of "_componentComm`s`Service"?

src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
```diff
@@ -96,96 +69,102 @@ export class ResourceInstanceFormComponent implements OnInit, OnChanges, OnDestr
...
+                            this.hasFileValue = 'audio';
+                        } else if (onto.properties[Constants.HasMovingImageFileValue]) {
+                            this.hasFileValue = 'movingImage';
+                        } else if (onto.properties[Constants.HasArchiveFileValue]) {
+                            this.hasFileValue = 'archive';
+                        } else if (onto.properties[Constants.HasTextFileValue]) {
+                            this.hasFileValue = 'text';
+                        } else {
+                            this.hasFileValue = undefined;
+                        }
```
Vijeinath, line 147: You could use switch case in that case

src/app/project/beta/ontology-classes/ontology-class-item/ontology-class-item.component.ts
```diff
@@ -66,6 +66,18 @@ export class OntologyClassItemComponent implements OnInit {
...
+
+        this.componentCommsSubscriptions.push(this._componentCommsService.on(
+            Events.resourceCreated, () => {
+                this._getSearchCount();
+            }
+        ));
+    }
+
+    ngOnDestroy(): void {
+        if (this.componentCommsSubscriptions !== undefined) {
```
Vijeinath, line 78: Is it necessary to check for undefined if you assign an empty array at the declaration?

Vijeinath (approved):
Works well. I've just added small things that are not severe

src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
mdelez: we have this code in a few places in the app so if we want to change it to a switch case, we should change it everywhere but then that kind of exceeds the goal of this PR

src/app/workspace/resource/resource-instance-form/resource-instance-form.component.ts
mdelez: I think it's fine. Technically the service should have an `s` at the end of `Communication`.

src/app/project/beta/ontology-classes/ontology-class-item/ontology-class-item.component.ts
mdelez: No, it's not necessary. I've removed it in [b0d2b01](https://github.com/dasch-swiss/dsp-app/pull/912/commits/b0d2b01577036aca6506a693129df69eb2fe94c8)

PR #923, opened 2023-02-17
chore: bump js-lib to 8.0.0

mpro7:


mdelez (approved):
will break the app but #918 will fix that

PR #932, opened 2023-02-28
chore(material): update material icons (DEV-1778)

mdelez:
resolves DEV-1778

google no longer provides .eot, .woff, or .woff2 files


Vijeinath (approved):
LGTM

PR #934, opened 2023-02-28
feat(main menu): changes to the main menu (DEV-1451)

domsteinbach:
resolves DEV-1451


mdelez (changes requested):
Screenshot 2023-03-10 at 13 56 29

There's some top and bottom padding that should be removed and the menu item is highlighted if you are on the respective route which should be removed. This may also overwrite what @Vijeinath has done in #944 but he will have to confirm that.

domsteinbach:
> Screenshot 2023-03-10 at 13 56 29
> 
> There's some top and bottom padding that should be removed and the menu item is highlighted if you are on the respective route which should be removed. This may also overwrite what @Vijeinath has done in #944 but he will have to confirm that.

I am very aware of the top and bottom padding. I could change it locally for this menu only - or change it for every menu within the app. The first option is imho wrong. The second option is out of the scope of this PR.

mdelez (approved):

PR #937, opened 2023-03-07
fix(lists): error messages only appear after focus (DEV-1692)

EricSommerhalder:
resolves DEV-1692


mdelez (approved):
works well, messi viel mal

PR #942, opened 2023-03-08
fix(cardinalities): enable the required toggle programmatically (DEV-1790)

domsteinbach:
…for newly assigned properties

resolves DEV-1790

### DIFF
--- a/src/app/project/ontology/property-form/property-form.component.ts
+++ b/src/app/project/ontology/property-form/property-form.component.ts
@@ -834,17 +834,17 @@ export class PropertyFormComponent implements OnInit {
         }
 
         if (this.resClassIri && !this.propertyInfo.propDef) {
-            // in context of class and with an existing property
+            // in context of class and without an existing property
             this.editMode = 'assignNewProperty';
         }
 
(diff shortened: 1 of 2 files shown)


src/app/project/ontology/property-form/property-form.component.ts
```diff
@@ -817,4 +820,32 @@ export class PropertyFormComponent implements OnInit {
...
+            // no editMode - just cards
+            return;
+        }
+        if (this.resClassIri && !!this.propertyInfo.propDef) {
+            // in context of class and with an existing property
+            this.editMode = 'assignExistingProperty';
+        }
+
+        if (this.resClassIri && !this.propertyInfo.propDef) {
+            // in context of class and with an existing property
```
mdelez, line 837: this should be "no existing property" right?

mdelez (changes requested):

src/app/project/ontology/property-form/property-form.component.ts
domsteinbach: Indeed a wrong comment.

domsteinbach:
Comment is changed. Tx!

mdelez (approved):

PR #950, opened 2023-03-15
chore: remove beta toggle (DEV-1449)

mdelez:
resolves DEV-1449


mdelez:
@Vijeinath @EricSommerhalder this is kind of a big PR but basically I just removed the beta feature toggle from the whole app and I was able to refactor some components now that we don't have to account for the old way anymore. I would be grateful if you could mess around in the app and make sure everything appears to work. I.E. create, edit, delete a resource class, property, resource class instance, and a list

Vijeinath (approved):
I focussed my attention mainly on testing the app (creating, editing, deleting) and it seems to work

PR #951, opened 2023-03-15
docs: Update the user guide regarding the full-text search section

flavens:
resolves DEV-1837


flavens:
@gautschr I did what I could but I think you could give me some more hints or examples to add or correct me.

gautschr (approved):
Fine, thanks!

PR #952, opened 2023-03-15
feat(resource): Restricted view message (DEV-1456)

EricSommerhalder:
resolves DEV-1456


src/app/workspace/resource/resource.component.html
```diff
@@ -1,3 +1,8 @@
+<div class="restricted-message" *ngIf="resource && resource.res && resource.res.userHasPermission=='RV' && showRestrictedMessage">
```
mdelez, line 1: you can use the safe navigation operator `?` to make this a bit cleaner and the equality check should be stricter by ensuring the type as well with `===`

`*ngIf="resource?.res?.userHasPermission === 'RV' && showRestrictedMessage"`

mdelez (changes requested):

mdelez (approved):

PR #967, opened 2023-03-22
build: add missing cache control headers when serving the app

subotic:
resolves DEV-1949


Dockerfile
```diff
@@ -1,9 +1,19 @@
...
-RUN rm -rf /public/*
+# set default internal port for the server
+# can be overriden at runsite
+ENV NGINX_PORT 4200
+EXPOSE ${NGINX_PORT}
 
-COPY apps/dsp-app /public
+# Copy nginx configuration
+COPY ./nginx/default.conf.template /etc/nginx/templates/
+# COPY ./nginx/nginx.conf /etc/nginx/
```
mpro7, line 12: Is this commented out line necessary?

subotic: yes, good catch

subotic: oh man, now I need to try it out again on the DEV server first.

nginx/nginx-security-headers.conf
```diff
@@ -0,0 +1,7 @@
+# add_header Strict-Transport-Security "max-age=31449600; includeSubDomains" always;
+# add_header Content-Security-Policy "object-src 'none'; script-src 'self'; script-src-elem 'self'; base-uri 'self'; require-trusted-types-for 'script'" always;
```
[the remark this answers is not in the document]
subotic: They are recommended settings but they don't work with our setup. I think that the docker image is also used without SSL. The second line messes up the Fathom script.

nginx/default.conf.template
```diff
@@ -0,0 +1,31 @@
...
+      add_header Cache-Control 'max-age=31449600'; # one year
+      include /etc/nginx/security-headers.conf;
+    }
+
+    location / {
+      try_files $uri$args $uri$args/ /index.html;
+
+      add_header Cache-Control 'max-age=86400'; # one day
+      include /etc/nginx/security-headers.conf;
+    }
```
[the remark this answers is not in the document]
subotic: everything else.

mpro7 (approved):

PR #972, opened 2023-03-28
feat(submit cardinnalities): disable double click

domsteinbach:
Disable submit button after click and show progress spinner instead to prevent a double click and a double request.

resolves DEV-1932

### DIFF
--- a/apps/dsp-app/src/app/project/ontology/property-form/property-form.component.ts
+++ b/apps/dsp-app/src/app/project/ontology/property-form/property-form.component.ts
@@ -836,6 +836,7 @@ export class PropertyFormComponent implements OnInit {
                     this.error = true;
                     this.loading = false;
                     this._errorHandler.showMessage(error);
+                    this.closeDialog.emit();
                 }
             );
     }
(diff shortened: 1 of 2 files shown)


apps/dsp-app/src/app/project/ontology/property-form/property-form.component.ts
```diff
@@ -828,6 +829,7 @@ export class PropertyFormComponent implements OnInit {
                     this.lastModificationDate = res.lastModificationDate;
                     // close the dialog box
                     this.loading = false;
+                    this.submittingChange = false;
                     this.closeDialog.emit();
                 },
                 (error: ApiResponseError) => {
```
Vijeinath, line 835: I think you should also set the submittingChange to false when you get in to the error body

Vijeinath (approved):

apps/dsp-app/src/app/project/ontology/property-form/property-form.component.ts
domsteinbach: Good point as it is not submitting anymore. But if there is an error we don't want a user to retry the same request which resulted in an error (retries should be introduced automatically).

So I prefer to close the whole dialog and destroy the whole property component. I'll change that.

PR #990, opened 2023-04-11
refactor (property-info-component): Split up property info component

domsteinbach:
resolves DEV-1868


mdelez (approved):
works well, thanks

PR #1014, opened 2023-04-19
fix: app breaks when on /account route

mdelez:
resolves DEV-2065


domsteinbach (approved):
I removed the cacheService usage for the user form component. LGTM

PR #1021, opened 2023-04-20
style: fix buttons on start page (more or less)

subotic:
resolves DEV-2064


apps/dsp-app/src/styles.scss
```diff
@@ -1,4 +1,282 @@
...
+// The following styles will override material design!
+
+// jdnconvertiblecalendar main style
+.mat-datepicker-content {
+  .mat-calendar {
+    height: auto !important;
+  }
+}
+
+/* TODO(mdc-migration): The following rule targets internal classes of card that may no longer apply for the MDC version.*/
```
mdelez, line 92: You should check if this is doing something still. If it is, you can delete the TODO comment, otherwise delete the css.

mdelez (approved):
Screenshot 2023-04-21 at 10 04 51

Another small thing. The button is smaller than the image (kinda hard to see in the screenshot). The button height used to be the height of the header.

apps/dsp-app/src/styles.scss
subotic: This has to do with grid. Is grid already migrated?

subotic:
> Screenshot 2023-04-21 at 10 04 51
> 
> Another small thing. The button is smaller than the image (kinda hard to see in the screenshot). The button height used to be the height of the header.

I will create another issue for this.

PR #1031, opened 2023-05-02
refactor: overview and project tile refactor

mdelez:
resolves DEV-2087


Vijeinath (approved):
It works! 
One thing that could be handy is the use of String Literal Types:
`type TileLinks = 'workspace' | 'settings';`

PR #1033, opened 2023-05-03
fix(resource viewer): images flickering or not loading

domsteinbach:
resolves DEV-1942
resolves DEV-1952


mdelez (approved):
the resource component needs to be refactored for sure but it's fine for now

PR #1034, opened 2023-05-03
feat(boolean values): implement removal of boolean values

domsteinbach:
resolves DEV-2120

### DIFF
--- a/apps/dsp-app/src/app/workspace/resource/values/boolean-value/boolean-value.component.ts
+++ b/apps/dsp-app/src/app/workspace/resource/values/boolean-value/boolean-value.component.ts
@@ -5,7 +5,7 @@ import {
     OnChanges,
     OnDestroy,
     OnInit,
-    SimpleChanges,
+    SimpleChanges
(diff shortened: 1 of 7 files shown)


apps/dsp-app/src/app/workspace/resource/values/boolean-value/boolean-value.component.ts
```diff
@@ -4,7 +4,7 @@ import {
     Input,
     OnChanges,
     OnDestroy,
-    OnInit,
+    OnInit
```
Vijeinath, line 7: The idea of the comma is: In case you import a new class you only change one line

domsteinbach: Yup, that's no secret. It is a question of style and preference: Some hate redundant characters - some like to set it in advance, so they don't have to set it if they add a new class, but set it there again in advance  ;-)

Vijeinath (approved):
Good

PR #1045, opened 2023-05-15
fix(app): top-level-domain with 63 chars

Vijeinath:
resolves DEV-2096


EricSommerhalder (approved):
Thank you! please try to keep pr's shorter in the future

domsteinbach (commented):
lgtm ;-)

PR #1047, opened 2023-05-15
fix(app): Full text search shows next 25 results

Vijeinath:
resolves DEV-2166

Should also not break DEV-2076


domsteinbach (approved):
lgtm

PR #1053, opened 2023-05-22
fix(boolean property comment):  Comment field is always enabled

domsteinbach:
For boolean properties(no invalid state for booleans), the comment field is always enabled.

resolves DEV-2167


Vijeinath (approved):
LGTM

PR #1057, opened 2023-05-22
fix(resource-list): set a fix width for the element within mat-list-item

domsteinbach:
resolves DEV-2179


mdelez (approved):
works well, thanks

PR #1082, opened 2023-06-07
chore: angular mdc migration menu

Vijeinath:
resolves DEV-2194


apps/dsp-app/src/app/workspace/resource/representation/document/document.component.scss
```diff
@@ -77,11 +77,3 @@ $osd-height: 460px;
     }
   }
 }
-
```
mdelez, line 80: I remember this being a specific request from Rebecca so maybe you want to check with her?

mdelez (approved):

apps/dsp-app/src/app/workspace/resource/representation/document/document.component.scss
Vijeinath: I've only managed to change the background but not the font color. Right now I will merge this version. But I create an style Issue and put it to the backlog

PR #1099, opened 2023-06-13
fix(freezing): app freeze if sipi does not respond (DEV-1562)

domsteinbach:
… freeze

resolves DEV-1562


apps/dsp-app/src/app/workspace/resource/representation/video/video.component.ts
```diff
@@ -39,6 +39,8 @@ import { PointerValue } from '../av-timeline/av-timeline.component';
 import { FileRepresentation } from '../file-representation';
 import { RepresentationService } from '../representation.service';
 import { NotificationService } from '../../../../main/services/notification.service';
+import { fileUrl } from 'ng-packagr/lib/ng-package/nodes';
```
mdelez, line 42: I'm assuming this was an erroneous auto import

domsteinbach: Indeed. Thanks!

PR #1110, opened 2023-06-22
fix(freeze): Remove/replace freezing request

domsteinbach:
resolves DEV-1562

### DIFF
--- a/apps/dsp-app/src/app/workspace/resource/representation/document/document.component.html
+++ b/apps/dsp-app/src/app/workspace/resource/representation/document/document.component.html
@@ -10,7 +10,7 @@
 
     <pdf-viewer
         class="pdf-viewer"
-        [src]= "{ url: src.fileValue.fileUrl, withCredentials: true }"
+        [src]= "{ url: src.fileValue.fileUrl,  withCredentials: true }"
         [original-size]="false"
         [autoresize]="true"
         [show-all]="true"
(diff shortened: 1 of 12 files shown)


apps/dsp-app/src/app/workspace/resource/representation/representation.service.ts
```diff
@@ -78,6 +57,20 @@ export class RepresentationService {
         );
     }
 
+    /**
+     * return the jwt token from the session to authenticate
+     * @return the token
+     */
+    public getTokenFromLocalStorage(): string {
```
mdelez, line 64: This is the same method that you have below but is private. Should it be public or private?

apps/dsp-app/src/app/workspace/resource/representation/document/document.component.html
```diff
@@ -10,7 +10,7 @@
 
     <pdf-viewer
         class="pdf-viewer"
-        [src]= "{ url: src.fileValue.fileUrl, withCredentials: true }"
+        [src]= "{ url: src.fileValue.fileUrl,  withCredentials: true }"
```
mdelez, line 13: does `withCredentials` have a fear of commas? :)

mdelez (approved):

apps/dsp-app/src/app/workspace/resource/representation/document/document.component.html
domsteinbach: I just needed some space ;-)

apps/dsp-app/src/app/workspace/resource/representation/representation.service.ts
domsteinbach: Damn, was due to merging. (I changed it to public in order the pdf viewer could possibly also send the token.)

PR #1125, opened 2023-06-30
fix(app): navigate to home after logout

Vijeinath:
resolves DEV-2243


apps/dsp-app/src/app/main/services/authentication.service.ts
```diff
@@ -31,8 +33,11 @@ export class AuthenticationService {
                 // destroy application state
                 this._applicationStateService.destroy();
 
-                // reload the page
-                window.location.reload();
+                // redirect to home and reload that page
+                this._router.navigate([''])
+                    .then(() => {
+                        window.location.reload();
```
mdelez, line 39: I think the reload is only necessary when you are on the overview page already when you logout. Maybe we can add some additional logic to check for this to reduce the number of full page reloads the app has to do.

mdelez (approved):

PR #1156, opened 2023-07-20
fix(ontology): Ontology cache is not updated when adding property to resource class

mdelez:
resolves DEV-2483


domsteinbach (approved):
LGTM

PR #1164, opened 2023-07-25
fix(geonames search results): filter out properties instead of error

domsteinbach:
resolves DEV-2467


mdelez (approved):
good work, thanks

PR #1172, opened 2023-08-02
chore: move date-picker out into a library

mdelez:
resolves DEV-2502


apps/dsp-app/src/app/workspace/search/advanced-search/resource-and-property-selection/search-select-property/specify-property-value/search-date-value/search-date-value.component.ts
```diff
@@ -1,12 +1,10 @@
 import { Component, Inject, Input, OnDestroy, OnInit } from '@angular/core';
-import {
-    UntypedFormBuilder,
+import {UntypedFormBuilder,
```
Vijeinath, line 2: What is here the code convention for multiple imports?

mdelez: Basically whatever the formatter chooses when you right click > format document

libs/vre/shared/app-date-picker/src/lib/app-date-picker/app-date-picker.component.ts
```diff
@@ -509,13 +646,61 @@ export class DatePickerComponent
...
+            );
+            return calDate.daysInMonth(date);
+        } else if (calendar === 'ISLAMIC') {
+            const calDate = new IslamicCalendarDate(
+                new CalendarPeriod(date, date)
+            );
+            return calDate.daysInMonth(date);
+        } else {
+            throw Error('Unknown calendar ' + calendar);
+        }
```
Vijeinath, line 701: Not mandatory but in case there are more than two if/ else if lines you can use switch case

Vijeinath (approved):
LGTM

libs/vre/shared/app-date-picker/src/lib/app-date-picker/app-date-picker.component.ts
mdelez: I agree but I didn't want to change too much of the code. I'd rather refactor it in a separate PR.

PR #1186, opened 2023-08-17
fix (type errors): fix type errors after dialog closed

domsteinbach:
resolves DEV-2542
Also:
- remove unnecessary subscription
- remove unnecessary refreshes when there are no changes


Vijeinath (approved):
LGTM

PR #1195, opened 2023-08-30
Feature/dev 2221 adapt dsp app to text cluttering v2

domsteinbach:
resolves DEV-2221


derschnee68:
no longer maintained, close.

PR #1222, opened 2023-10-12
chore: Reconfigure all "zio" endpoints to use the original address

seakayone:
The dsp-api will not expose "ziofied" endpoints on a seperate port/route anymore. We will migrate to using zio transparent to all of our clients. All existing project endpoints should therefore be used on the original address.

This PR removes all configuration for routes using the now deprecated endpoint on its own port (`5555` locally) or path (`..zio/` on the loadbalancer).

Out of scope for this PR is removing custom code which allowed for changing the route programatically. Remove the app's unused code is done in [DEV-2808](https://linear.app/dasch/issue/DEV-2808).

Related changes:
* ops-deploy: https://github.com/dasch-swiss/ops-deploy/pull/531


BalduinLandolt (commented):
Looks reasonable to me, but I don't know the App well enough to say if that suffices or if more needs to be done

SamuelBoerlin (commented):
I also can't comment on whether this is good to go or whether some more changes are required in APP. But I approve of the removal of the special ZIO routes.

mdelez (approved):
Yeah this will do the trick but we need to be cautious and merge the [API PR](https://github.com/dasch-swiss/dsp-api/pull/2879) first. Otherwise this PR will break the app on dev.

mpro7:
This PR should be marked as breaking change here too.

PR #1229, opened 2023-10-17
fix(advanced-search): generate correct gravsearch for linked resources with property values

mdelez:
resolves DEV-2839


libs/vre/advanced-search/src/lib/data-access/gravsearch-service/gravsearch.service.ts
```diff
@@ -119,17 +119,34 @@ export class GravsearchService {
                     ' .';
                 whereString = constructString;
             }
+            // if search value is an array that means that it's a linked resource with child properties
             if (Array.isArray(property.searchValue)) {
                 property.searchValue.forEach((value, i) => {
                     if (value.selectedProperty?.objectType !== ResourceLabel) {
```
irmastnt, line 125: Probably worth creating static method which updates search strings or private method which returns search string object

irmastnt (approved):

PR #1240, opened 2023-10-25
refactor(routing): cleanup unused routes

domsteinbach:
resolves DEV-2875


libs/vre/shared/app-config/src/lib/app-config/app-constants.ts
```diff
@@ -1,8 +1,8 @@
 export class RouteConstants {
     static readonly home = '';
     static readonly help = 'help';
-    static readonly login = 'login';
-    static readonly logout = 'logout';
+    // static readonly login = 'login';
```
mdelez, line 4: you can delete all these commented out constants, no?

PR #1251, opened 2023-10-30
refactor(date picker): implement date picker as standalone component

domsteinbach:
resolves DEV-2688


libs/vre/advanced-search/src/lib/ui/property-form/property-form-value/property-form-value.component.ts
```diff
@@ -29,7 +28,6 @@ import { PropertyFormItem } from '../../../data-access/advanced-search-store/adv
         MatSelectModule,
         FormsModule,
         ReactiveFormsModule,
-        AppDatePickerModule,
```
mdelez, line 32: Please add the import for the `AppDatePickerComponent` here

mdelez (approved):

PR #1287, opened 2023-11-29
fix (user form): create new user form is not displayed properly

domsteinbach:
resolves DEV-3032

Quick and dirty as the whole user form will be refactored next week. Actually fixes three issues:

- create new user now passes a user:User to the password form. (needed) and displays the password correctly
- edit user gets the user by the username and populates the form correctly
- passing the search term in AddUserComponent as username lead to a 404 as the userFormComponent gets the user by that username


apps/dsp-app/src/app/user/user-form/user-form.component.ts
```diff
@@ -224,6 +198,63 @@ export class UserFormComponent implements OnInit, OnChanges {
...
+        this.title = this.user.username;
+        this.subtitle = "'appLabels.form.user.title.edit' | translate";
+        this.loadingData = !this.buildForm(this.user);
+    }
+
+    editExistingUser() {
+        this.title = this.name;
+        this.subtitle = "'appLabels.form.user.title.edit' | translate";
+
+        this._dspApiConnection.admin.usersEndpoint
```
irmastnt, line 211: I think you should be able to get user data from the state. If not, dispatch action which loads user data and then you should be able to get it from the state. This way we get rid of unnecessary request to the server.

apps/dsp-app/src/app/user/user-form/user-form.component.ts
```diff
@@ -224,6 +198,63 @@ export class UserFormComponent implements OnInit, OnChanges {
...
+                (error: ApiResponseError) => {
+                    this._errorHandler.showMessage(error);
+                }
+            );
+    }
+
+    defineNewUser() {
+
+        // get existing users to avoid same usernames and email addresses
+        this.allUsers$
```
irmastnt, line 228: this subscription is not needed, just pre set new user data

apps/dsp-app/src/app/user/user-form/user-form.component.ts
```diff
@@ -224,6 +198,63 @@ export class UserFormComponent implements OnInit, OnChanges {
...
+            .pipe(take(1))
+            .subscribe((allUsers) => {
+                for (const user of allUsers) {
+                    // email address of the user should be unique.
+                    // therefore we create a list of existing email addresses to avoid multiple use of user names
+                    this.existingEmails.push(
```
irmastnt, line 234: existingEmails should be observable which observes existing user emails from the state

apps/dsp-app/src/app/user/user-form/user-form.component.ts
```diff
@@ -224,6 +198,63 @@ export class UserFormComponent implements OnInit, OnChanges {
...
+                        new RegExp('(?:^|W)' + user.email.toLowerCase() + '(?:$|W)'));
+                    // username should also be unique.
+                    // therefore we create a list of existingUsernames to avoid multiple use of user names
+                    this.existingUsernames.push(
```
irmastnt, line 238: same story like with existingemails, it should be observable

irmastnt (changes requested):

apps/dsp-app/src/app/user/user-form/user-form.component.ts
domsteinbach: Most certainly. That's exactly what is planned in the projects issue concerning the user form next week, as I wrote:. We refactor the whole component. This PR here is fixing some bugs. It is reintroducing what was there and only meant to fix, what is broken by now.

However, here we can not get the user state for this, as in this case the logged in user is editing another users account and not its own. We can not get that user via an action either, then the user which is edited will be displayed as logged in user. The goal is to not pass a user as input at all, but this would entail a proper refactoring like I am doing next week - not a bug fix. The requirements of the new user form are now defined since this morning.

If it can wait until next week I am fine.

apps/dsp-app/src/app/user/user-form/user-form.component.ts
domsteinbach: Very much agree. Let's go for that next week.

apps/dsp-app/src/app/user/user-form/user-form.component.ts
domsteinbach: same

domsteinbach:
Please decide about these fixes. I rewrite those components next week anyway.

PR #1317, opened 2023-12-18
fix: class instance adding and preview. 

irmastnt:
fixed class instance adding and preview. Added ontology class instance guard.


apps/dsp-app/src/app/app-routing.module.ts
```diff
@@ -28,6 +28,7 @@ import { ResultsComponent } from './workspace/results/results.component';
 import { AdvancedSearchContainerComponent } from './workspace/search/advanced-search/advanced-search-container.component';
 import { ProjectFormComponent } from "@dsp-app/src/app/project/project-form/project-form.component";
```
derschnee68, line 29: use relative import ?

apps/dsp-app/src/app/main/guard/ontology-class-instance.guard.ts
```diff
@@ -0,0 +1,49 @@
...
+import { RouteConstants } from '@dasch-swiss/vre/shared/app-config';
+import { Select } from '@ngxs/store';
+import { UserSelectors } from '@dasch-swiss/vre/shared/app-state';
+import { StoredProject } from '@dasch-swiss/dsp-js';
+
+
+@Injectable({
+    providedIn: 'root'
+})
+export class AddOntologyClassInstanceGuard implements CanActivate {
```
derschnee68, line 16: Rename file from ontology-class-instance.guard.ts to add-ontology-class-instance.guard.ts or the opposite with the class

derschnee68 (changes requested):

apps/dsp-app/src/app/app-routing.module.ts
derschnee68: like ./project/...

derschnee68 (approved):

PR #1325, opened 2023-12-19
refactor: add rule nx enforce-module-boundaries

derschnee68:
resolves DEV-

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,32 +20,6 @@
     {
       "files": [
         "*.ts",
-        "*.tsx",
-        "*.js",
-        "*.jsx"
-      ],
-      "rules": {
-        "@nrwl/nx/enforce-module-boundaries": [
-          "warn",
-          // TODO set to error later
-          {
-            "enforceBuildableLibDependency": true,
-            "allow": [],
-            "depConstraints": [
-              {
-                "sourceTag": "*",
-                "onlyDependOnLibsWithTags": [
-                  "*"
-                ]
-              }
-            ]
-          }
-        ]
-      }
-    },
-    {
-      "files": [
-        "*.ts",
         "*.tsx"
       ],
       "excludedFiles": [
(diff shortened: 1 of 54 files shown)


irmastnt (approved):
I am not a fan of relative paths, hope it will not give as problems with webpack in the future. Merge it according to the best of your knowledge.

BTW, tried to ask chat GPT, this is what I've got:

In an Nx workspace, enforcing module boundaries is a helpful practice for maintaining a clean and scalable codebase. However, there might be cases where you need to use absolute paths without triggering the Nx rule `enforce-module-boundaries`. To achieve this, you can follow these steps:

### 1. Configure `tsconfig.base.json`:

In your Nx workspace, you have a `tsconfig.base.json` file. Configure the `compilerOptions` section to include the `"baseUrl"` and `"paths"` options. Here's an example:

```json
{
  "compilerOptions": {
    // other options...

    "baseUrl": "./src",
    "paths": {
      "@your-app/*": ["app/*"]
      // Add more path aliases as needed
    }
  }
}
```

This configuration sets up a base URL for your TypeScript compiler and defines a path alias (`@your-app/*`) that maps to the `app/*` directory. Adjust the alias and paths according to your project structure.

### 2. Use the Aliases in Imports:

Now, you can use these aliases in your import statements. For example:

```typescript
// Before
// import { SomeComponent } from '../../../path-to-component';

// After
import { SomeComponent } from '@your-app/some-component-path';
```

### 3. Configure Nx Workspace Rules:

In your `nx.json` file, configure the `implicitDependencies` section to explicitly define module boundaries:

```json
{
  // other options...
  
  "projects": {
    "your-app": {
      // other options...

      "implicitDependencies": {
        // Define module boundaries explicitly
        "tsconfig.base.json": ["!**"]
      }
    }
  }
}
```

This configuration tells Nx not to enforce module boundaries based on `tsconfig.base.json`. Adjust the project name (`your-app`) and the rules according to your project structure.

### 4. Run Nx Commands:

Now, you should be able to use absolute paths without triggering the `enforce-module-boundaries` rule. When running Nx commands, such as `nx lint` or `nx build`, Nx will respect the configured module boundaries.

Keep in mind that bypassing module boundaries should be done cautiously, as enforcing them helps maintain a well-organized codebase. Only use this approach when necessary, and make sure to communicate any deviations from the standard practices with your team.

PR #1331, opened 2024-01-05
feat: reloads state if browser refreshed manually by the user

irmastnt:
resolves DEV-


derschnee68:
Needs discussions about this PR, as feature seems opposed to Ivan's message (dec 13)
"Refreshing the page or opening a new tab should simply fetch the data from the backend. The only thing that needs to be persisted locally is the JWT token and even then, the first thing would be to check if it is still valid."

In my opinion, the only thing we should do is:
AppComponent -> onInit() {
clearState()
}

PR #1345, opened 2024-01-15
fix: add tooltip on non accessible list

derschnee68:
Also refactor some parts of the component

### DIFF
--- a/apps/dsp-app/src/app/project/data-models/data-models.component.html
+++ b/apps/dsp-app/src/app/project/data-models/data-models.component.html
@@ -21,39 +21,21 @@
   </div>
   <div *ngIf="ontologiesMetadata$ | async" class="projectOntos">
     <div class="list" [class.top-padding]="isAdmin$ | async">
-      <ng-container *ngIf="(isLoggedIn$ | async) === true">
-        <div
-          class="list-item ontos"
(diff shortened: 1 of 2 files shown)


apps/dsp-app/src/app/project/data-models/data-models.component.html
```diff
@@ -10,50 +10,37 @@
     </mat-icon>
   </div>
   <div *ngIf="isAdmin$ | async" class="action-buttons">
-    <button color="primary" mat-raised-button class="create" (click)="open('add-ontology')">
+    <a color="primary" mat-raised-button class="create" [routerLink]="['..', 'add-ontology']">
```
irmastnt, line 13: use route constants, in this way we can track where this route is used in whole app

apps/dsp-app/src/app/project/data-models/data-models.component.html
```diff
@@ -67,19 +54,25 @@
     </mat-icon>
   </div>
   <div *ngIf="isAdmin$ | async" class="action-buttons">
-    <button color="primary" mat-raised-button class="create" (click)="open('add-list')">
+    <a color="primary" mat-raised-button class="create" [routerLink]="['..','add-list']">
```
irmastnt, line 57: route constants

irmastnt (changes requested):

irmastnt (approved):

PR #1346, opened 2024-01-15
fix(dsp-app): ontology loading, ontology clases refresh, reloads proj…

irmastnt:
…ogout

resolves DEV-

### DIFF
--- a/libs/vre/shared/app-session/src/lib/auth.service.ts
+++ b/libs/vre/shared/app-session/src/lib/auth.service.ts
@@ -6,7 +6,6 @@ import {
   ClearListsAction,
   ClearOntologiesAction,
   ClearProjectsAction,
-  LoadProjectsAction,
   LogUserOutAction,
 } from '@dasch-swiss/vre/shared/app-state';
 import { Actions, Store, ofActionSuccessful } from '@ngxs/store';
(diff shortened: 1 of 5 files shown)


libs/vre/shared/app-session/src/lib/auth.service.ts
```diff
@@ -176,14 +178,20 @@ export class AuthService {
 
   doLogoutUser() {
     this.removeTokens();
+    this._actions
+      .pipe(ofActionSuccessful(ClearProjectsAction))
+      .pipe(take(1))
+      .subscribe(() => {
+        this.store.dispatch(new LoadProjectsAction());
```
derschnee68, line 185: Why do you load projects when user logs out ?

libs/vre/shared/app-session/src/lib/auth.service.ts
```diff
@@ -176,14 +178,20 @@ export class AuthService {
...
+        this.router.navigate([RouteConstants.home], { replaceUrl: true });
+      });
     this.store.dispatch([
       new LogUserOutAction(),
```
derschnee68, line 189: Why do we not call this line anymore ?

derschnee68 (approved):

PR #1358, opened 2024-01-17
feat(dsp-app): introduced ontology class item count from state

irmastnt:
resolves DEV-


apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
```diff
@@ -35,15 +34,20 @@ import {
...
   @Input() resClass: ClassDefinition;
 
   @Input() projectMember: boolean;
 
   @ViewChild('resClassLabel') resClassLabel: ElementRef;
 
-  gravsearch: string;
-
-  results: number;
+  get results$(): Observable<number> {
```
derschnee68, line 45: dont use get with observables

libs/vre/shared/app-state/src/lib/ontologies/ontologies.state.ts
```diff
@@ -146,10 +147,22 @@ export class OntologiesState {
               this._actions$
                 .pipe(ofActionSuccessful(LoadOntologyAction))
                 .pipe(take(1))
-                .subscribe(() =>
+                .subscribe(() => {
```
derschnee68, line 150: subscribe into subscribe

derschnee68 (changes requested):

derschnee68 (approved):

PR #1372, opened 2024-01-23
fix(still image): User without permission can draw a region

domsteinbach:
and press the replace image button
resolves DEV-3200

### DIFF
--- a/apps/dsp-app/src/app/workspace/resource/resource.component.html
+++ b/apps/dsp-app/src/app/workspace/resource/resource.component.html
@@ -23,6 +23,7 @@
         [currentTab]="selectedTabLabel"
         [parentResource]="incomingResource ? incomingResource.res : resource.res"
         [activateRegion]="selectedRegion"
+        [adminPermissions]="isAdmin$ | async"
         (loaded)="representationLoaded($event)"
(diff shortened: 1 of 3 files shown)


apps/dsp-app/src/app/workspace/resource/resource.component.html
```diff
@@ -23,6 +23,7 @@
         [currentTab]="selectedTabLabel"
         [parentResource]="incomingResource ? incomingResource.res : resource.res"
         [activateRegion]="selectedRegion"
+        [adminPermissions]="isAdmin$ | async"
```
derschnee68, line 26: You don't need to add an input adminPermission if you use the admin selector. You can add it straight to the component ?

derschnee68 (changes requested):

apps/dsp-app/src/app/workspace/resource/resource.component.html
domsteinbach: There is no equivalent selector yet as far as I can see (it is not sysadmin, it is sysadmin or project admin/member of the current project). So I went for a short streamlined fix reusing what is already there instead of refactoring and implementing new stuff as well.

derschnee68 (approved):

PR #1400, opened 2024-01-29
fix: display error if it is not an API error

derschnee68:
resolves DEV-


libs/vre/shared/app-error-handler/src/lib/app-error-handler.ts
```diff
@@ -17,6 +17,8 @@ export class AppErrorHandler implements ErrorHandler {
     } else if (error instanceof HttpErrorResponse) {
       // ApiServices
       this.handleHttpError(error, error.url);
+    } else {
```
irmastnt, line 20: remove this

libs/vre/shared/app-string-literal/src/lib/multi-language-form.service.ts
```diff
@@ -53,7 +53,12 @@ export class MultiLanguageFormService {
     return this.formArray.controls.find(control => control.value.language === lang && control.value.value !== '');
   }
 
-  changeLanguage(languageIndex: number) {
+  changeLanguage(languageIndex: number, deleteCurrentIfEmpty = true) {
+    console.log(this.selectedFormControl.value);
```
irmastnt, line 57: one more..

irmastnt (changes requested):

PR #1412, opened 2024-01-30
fix: no faro in dev

derschnee68:
resolves DEV-


libs/vre/shared/app-analytics/src/lib/grafana-faro.service.ts
```diff
@@ -9,6 +9,10 @@ export class GrafanaFaroService {
   constructor(private readonly _appConfig: AppConfigService) {}
 
   setup() {
+    if (this._appConfig.dspInstrumentationConfig.environment === 'local-dev') {
```
irmastnt, line 12: introduce environment constants

irmastnt (changes requested):

irmastnt (approved):

PR #1443, opened 2024-02-07
fix: styling issue in project membership

derschnee68:


irmastnt (changes requested):
image

- Make select list full width
- Distinguish project items(or project list) style from text label style

irmastnt (approved):

PR #1475, opened 2024-02-20
test: e2e data model tests (DEV-2379, DEV-2341, DEV-2339, DEV-2338)

irmastnt:
resolves DEV-2379, DEV-2341, DEV-2339, DEV-2338


apps/dsp-app/cypress/support/commands.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference types="cypress" />
```
derschnee68, line 1: why do we have interfaces in this file ?

apps/dsp-app/cypress/support/helpers/constants.ts
```diff
@@ -0,0 +1,36 @@
+export enum ClassType {
```
derschnee68, line 1: why are there enums in a constants.ts file ? (should be only constants)

apps/dsp-app/cypress/support/helpers/constants.ts
```diff
@@ -0,0 +1,36 @@
+export enum ClassType {
+  Resource = 'Resource',
+  StillImageRepresentation = 'StillImageRepresentation',
+  MovingImageRepresentation = 'MovingImageRepresentation',
+  AudioRepresentation = 'AudioRepresentation',
+  TextRepresentation = 'TextRepresentation',
+  DocumentRepresentation = 'DocumentRepresentation',
+  ArchiveRepresentation = 'ArchiveRepresentation',
+}
+
```
derschnee68, line 10: can't we import those enums / interfaces ? Don't they exist somewhere in the project ?

apps/dsp-app/cypress.config.ts
```diff
@@ -5,6 +5,9 @@ export default defineConfig({
   e2e: {
     specPattern: 'cypress/**/**/**/*.cy.ts',
     excludeSpecPattern: ['*.spec.js', '*.spec.ts'],
+    viewportHeight: 768,
```
derschnee68, line 8: why do we setup this ? Can't we have it flexible to have multiple test with different sizes ?

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
```diff
@@ -0,0 +1,61 @@
+import { faker } from '@faker-js/faker';
+import ProjectPage from '../../support/pages/project-page';
+
+describe('Data Model', () => {
```
derschnee68, line 4: Shouldn't we call this ontologies ?

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+    cy.visit(`/project/${projectPage.projectUuid}/data-models`);
+    cy.url().should('match', new RegExp(`project\/${projectPage.projectUuid}\/data-models`));
+
+    cy.get('[data-cy=data-models-container]')
+      .find('[data-cy=create-button]')
+      .scrollIntoView()
+      .should('be.visible')
+      .click();
+
+    cy.get('[data-cy=name-input]').type(data.name).wait(200);
```
derschnee68, line 28: why do you add wait() ? it is a bad pattern

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+    cy.get('[data-cy=label-input]').clear().type(data.label).wait(200);
+    cy.get('[data-cy=comment-textarea]').type(data.comment).wait(200);
+    cy.get('[data-cy=submit-button]').click();
+
+    cy.wait('@submitRequest').wait(5000);
+    cy.url().should('match', new RegExp(`project/${projectPage.projectUuid}/ontology/${data.name}/editor/classes`));
```
derschnee68, line 34: that is not really a regex, use should('include')

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    cy.createOntology(projectPage.projectUuid).then(ontology => {
+      cy.get('[data-cy=edit-ontology-button]').scrollIntoView().should('be.visible').click();
+      cy.get('[data-cy=label-input]').clear().type(data.label).wait(200);
+      cy.get('[data-cy=comment-textarea]').type(data.comment).wait(200);
+      cy.get('[data-cy=submit-button]').click();
+
+      cy.wait('@updateRequest').wait(5000);
+      cy.url().should(
+        'match',
```
derschnee68, line 55: includer

apps/dsp-app/cypress/e2e/system-admin/data-model-class.cy.ts
```diff
@@ -0,0 +1,135 @@
+import { faker } from '@faker-js/faker';
+import { ClassType, PropertyType } from '../../support/helpers/constants';
+import ProjectPage from '../../support/pages/project-page';
+
+describe('Data Model Class', () => {
```
derschnee68, line 5: isn't that called resource class ? Maybe I'm mistaken

apps/dsp-app/cypress/support/commands/data-model-class-command.ts
```diff
@@ -0,0 +1,29 @@
+import { faker } from '@faker-js/faker';
```
derschnee68, line 1: why do we have this command ? Can't we do it with an api request ?

apps/dsp-app/cypress/support/e2e.ts
```diff
@@ -1,5 +1,15 @@
-import './commands/login';
 import { UserProfiles } from '../models/user-profiles';
+import './commands/data-model-class-command';
+import './commands/login';
+import './commands/ontology-command';
+
+const resizeObserverLoopErrRe = /^[^(ResizeObserver loop limit exceeded)]/;
```
derschnee68, line 6: what is this for?

apps/dsp-app/cypress/support/commands/ontology-command.ts
```diff
@@ -0,0 +1,27 @@
+import { faker } from '@faker-js/faker';
```
derschnee68, line 1: can't we add it with api?

apps/dsp-app/cypress/support/pages/project-page.ts
```diff
@@ -0,0 +1,34 @@
+import { ProjectADM, ProjectOperationResponseADM } from '../../../../../libs/vre/open-api/src';
+
+class ProjectPage {
+  private payload = {
```
derschnee68, line 4: payload is only used in request Project, add it as a constant of requestProject()

derschnee68 (changes requested):
Please also remove all wait()

apps/dsp-app/cypress/support/commands.ts
irmastnt: those are used in the commands, but I can move it to separate file, how do you prefer?

apps/dsp-app/cypress/support/helpers/constants.ts
irmastnt: I prefer to have enums and constants in the same file, it will be used just for e2e. So far there is no constants yet.

apps/dsp-app/cypress/support/helpers/constants.ts
irmastnt: I think those comes from the backend. Anyhow, its ok to have strings unrelated with API in e2e tests.

apps/dsp-app/cypress.config.ts
irmastnt: I dont like default cypress size, its even smaller than this and not targeted to our users, also easier to see in the runner. I believe it can be overridden or pass different configuration files for multiple resolution testing.

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
irmastnt: Yeah.. its a bit messed up all over the project. routes, api and some labels called - data model, some other labels called ontology. Are you sure I need to rename it?

apps/dsp-app/cypress/support/commands/data-model-class-command.ts
irmastnt: I think request is not the same. e2e should simulate user behaviour from start to end.

apps/dsp-app/cypress/support/e2e.ts
irmastnt: I tried to get rid of those uncaught exceptions which flooding the console. Copied it from somewhere.

apps/dsp-app/cypress/support/commands/ontology-command.ts
irmastnt: same as with data-model-class-command. Should simulate user.

irmastnt:
> Also, all the e2e titles don't look great to me: Data Model Class -> should cancel to create a data model class
> 
> This has no meaning to me. Please make something more readable for every new tests

There may be no perfection, but I would be happy, if better name is proposed.

irmastnt:
> Please also remove all wait()

Due to multiple loaders and many requests is hard to manage without it. I think we can get rid of it over time when we improve this app. Some waits is waiting for UI components to be rendered on the client side, it takes milliseconds.

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
irmastnt: https://github.com/dasch-swiss/dsp-das/pull/1475#issuecomment-1959179537

apps/dsp-app/cypress/support/commands.ts
derschnee68: yes please!

apps/dsp-app/cypress/e2e/system-admin/data-model.cy.ts
derschnee68: I think ontologies seems better, it's everywehere in the app and everybody talks about ontology

derschnee68 (approved):

PR #1489, opened 2024-02-29
refactor: type multi language controls

derschnee68:
resolves DEV-3343


libs/vre/shared/app-string-literal/src/index.ts
irmastnt, line 1: Move multi language form components out from string literal

libs/vre/shared/app-string-literal/src/lib/multi-language-form.service.ts
```diff
@@ -1,43 +1,42 @@
...
-      this.formArray.controls.find(
-        control => control.value.language === this.availableLanguages[this.selectedLanguageIndex]
-      ) as FormControl
-    ) // TODO as FormControl = bad pattern, should be refactored, also form should be with Typed values
-      .get('value') as FormControl;
+  get selectedLanguageControl() {
+    return this.selectedFormControl.controls.value;
   }
 
   get invalidErrors() {
```
irmastnt, line 32: check this getter usages. undefined is not expected in some places

irmastnt (changes requested):

libs/vre/shared/app-string-literal/src/lib/multi-language-form.service.ts
derschnee68: it is intended to be undefined, but the check is done before

libs/vre/shared/app-string-literal/src/index.ts
derschnee68: it is actually the string-literal component that will be removed and folder will be renamed as soon as the last refactor is done (next PR)

PR #1494, opened 2024-02-29
refactor: forms using multi language inputs should be typed

derschnee68:
resolves DEV-3355


apps/dsp-app/src/app/project/ontology/edit-resource-class-dialog/edit-resource-class-dialog.component.ts
irmastnt, line 1: pre-filling data is missing

irmastnt (changes requested):

irmastnt (approved):

PR #1525, opened 2024-03-19
fix: resource attached project retrieving (#DEV-3424)

irmastnt:
resolves DEV-3424

### DIFF
--- a/apps/dsp-app/src/app/workspace/resource/properties/properties.component.ts
+++ b/apps/dsp-app/src/app/workspace/resource/properties/properties.component.ts
@@ -32,6 +32,7 @@ import {
   ReadUser,
   ReadValue,
   ResourcePropertyDefinition,
+  StoredProject,
   UpdateResourceMetadata,
   UpdateResourceMetadataResponse,
 } from '@dasch-swiss/dsp-js';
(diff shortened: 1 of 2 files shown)


apps/dsp-app/src/app/workspace/resource/properties/properties.component.ts
```diff
@@ -567,6 +553,34 @@ export class PropertiesComponent implements OnInit, OnChanges, OnDestroy {
...
+  private _getResourceAttachedData() {
+    this._store.dispatch([
+      new GetAttachedUserAction(this.resource.res.id, this.resource.res.attachedToUser),
+      new GetAttachedProjectAction(this.resource.res.id, this.resource.res.attachedToProject),
+    ]);
+
+    this.project = this._store
+      .selectSnapshot(ProjectsSelectors.allProjects)
+      .find(p => p.id === this.resource.res.attachedToProject);
+    this._actions$
```
domsteinbach, line 565: This seems obsolete and wrong. Why is this project set in two different ways once to a ReadProject and then to a StoredProject?

domsteinbach (approved):

apps/dsp-app/src/app/workspace/resource/properties/properties.component.ts
irmastnt: project is provided from multiple endpoints. One of them has StoredProject, other has ReadProject which extends StoredProject. It would be nice, if API can provide single type.

PR #1551, opened 2024-04-12
fix: color picker style (DEV-3476)

irmastnt:
resolves DEV-3476


domsteinbach (approved):
I suppose we want to refactor the color picker at one point and get rid of all ng deep.

PR #1556, opened 2024-04-19
test(docs): add markdownlint and mkdocs CI checks

jnussbaum:
check every PR if the docs comply with mkdocs and markdownlint

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,17 +17,13 @@ find-ignored-tests: ## find all ignored tests (e.g. fdescribe)
 # Documentation targets
 #################################
 
-.PHONY: docs-publish
-docs-publish: ## build and publish docs to github Pages
-	mkdocs gh-deploy
-
 .PHONY: docs-build
 docs-build: ## build docs into the local 'site' folder
-	mkdocs build
+	mkdocs build --strict
 
 .PHONY: docs-serve
 docs-serve: ## serve docs for local viewing
-	mkdocs serve
+	mkdocs serve --strict
 
 .PHONY: docs-install-requirements
 docs-install-requirements: ## install requirements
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,17 +17,13 @@ find-ignored-tests: ## find all ignored tests (e.g. fdescribe)
 # Documentation targets
 #################################
 
-.PHONY: docs-publish
-docs-publish: ## build and publish docs to github Pages
-	mkdocs gh-deploy
-
```
jnussbaum, line 23: We don't publish the docs of DSP-APP separately any more. So this command isn't used any more.

Makefile
```diff
@@ -17,17 +17,13 @@ find-ignored-tests: ## find all ignored tests (e.g. fdescribe)
...
 .PHONY: docs-build
 docs-build: ## build docs into the local 'site' folder
-	mkdocs build
+	mkdocs build --strict
```
jnussbaum, line 22: The `--strict` flag is necessary, so that warnings are not silently ignored. Often, those warnings later create problems when I want to publish the docs to docs.dasch.swiss. So it is better to catch them early, and let the CI fail if problematic markdown files are checked in.

docs/user-guide/publication.md
jnussbaum, line 1: This file is empty, and is not referenced in the documentation. So it doesn't make sense to leave it here.

docs/user-guide/user.md
```diff
@@ -2,7 +2,7 @@
 
 ## Your user profile and projects
 
-To change your personal information as well as your default language used by the interface, you can edit your profile by clicking *Edit my profile*. Currently, the avatar image comes from gravatar.com (go on their <a href="http://en.gravatar.com/" target="_blank">website</a> to register if you want your customized user photo).
+To change your personal information as well as your default language used by the interface, you can edit your profile by clicking *Edit my profile*. Currently, the avatar image comes from gravatar.com (go on their [website](http://en.gravatar.com/) to register if you want your customized user photo).
```
jnussbaum, line 5: markdown-style links are preferred over HTML-syntax

irmastnt (approved):

PR #1560, opened 2024-04-23
fix: alerts user when a resource cannnot be erased

derschnee68:
resolves DEV-3528


irmastnt (changes requested):
I think we should handle it globally, just display message sent by the backend

irmastnt (approved):

PR #1562, opened 2024-04-23
fix(system component): don't dispatch loading state for users; faster loading, prevent reloads on tab change 

domsteinbach:
resolves DEV-3530

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,4 +232,5 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r docs/requirements.txt
+      - run: make docs-lint
       - run: make docs-build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## [11.9.0](https://github.com/dasch-swiss/dsp-das/compare/v11.8.0...v11.9.0) (2024-04-19)
+
+
+### Enhancements
+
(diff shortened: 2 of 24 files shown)


irmastnt (commented):
@domsteinbach would you consider to adjust this code according to official material documentation: https://material.angular.io/components/tabs/overview#tabs-and-navigation

irmastnt (changes requested):
see previous comment

domsteinbach:
> would you consider to adjust this code according to official material documentation: https://material.angular.io/components/tabs/overview#tabs-and-navigation
> Also users should not be loaded when "System Administration" or "All projects" tab is opened (see screenshot). It should be loaded only when user clicks on "All Users" tab. 

Funny, it was my very intention to not let the tabs route, as it is a really, really bad UX. I prefer to have things ready without any waiting time when someone is changing tabs in a ui. It was not expensive to preload the users imho, but of value. Otherwise, when the tabs are routing, it is annoying: When a user navigates to system, then there is another loading time for when the tab is changed to /users. The users list is also destroyed after changing the tab back. Every tab change will reload the projects/all the users with all the users images again and again (which needs to be fixed there of course).

irmastnt (approved):

irmastnt:
Thanks @domsteinbach. Its really better now. 
Its not necessary to load users when projects tab is opened. Preloading data is a guessing game, possibly this data may never be used. Therefore this technique should be used only for exceptional cases. Also user wants to have latest data before starting working with it.
Lets have it as a common way to load data for tabs.

PR #1564, opened 2024-04-24
refactor: ontologies / properties editor

derschnee68:
resolves DEV-3302


irmastnt:
BUGS:

- 
image

- when creating new resource, save button should be disabled, until control validation is passing
- saving resource error:
- 
image

- change detection issue: all resource properties is shown only after change detection is triggered

PR #1569, opened 2024-05-06
fix(loading state): use correct loading state

domsteinbach:
resolves DEV-3530

### DIFF
--- a/libs/vre/shared/app-state/src/lib/user/user.selectors.ts
+++ b/libs/vre/shared/app-state/src/lib/user/user.selectors.ts
@@ -12,6 +12,11 @@ export class UserSelectors {
   }
 
   @Selector([UserState])
+  static usersLoading(state: UserStateModel): boolean | undefined {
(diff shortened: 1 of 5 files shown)


libs/vre/shared/app-state/src/lib/user/user.selectors.ts
```diff
@@ -11,6 +11,11 @@ export class UserSelectors {
     return state.isLoading;
   }
 
+  @Selector([UserState])
+  static usersLoading(state: UserStateModel): boolean | undefined {
```
derschnee68, line 15: only boolean

derschnee68 (approved):

libs/vre/shared/app-state/src/lib/user/user.selectors.ts
domsteinbach: Indeed not needed. Copy/pasted it from user loading state.

PR #1581, opened 2024-05-14
fix(erasing resource): reintroduce necessary comment

domsteinbach:
resolves DEV-3627


libs/vre/shared/app-resource-properties/src/lib/erase-resource-dialog.component.ts
```diff
@@ -40,7 +58,7 @@ export class EraseResourceDialogComponent {
     const payload = new DeleteResource();
     payload.id = this.data.resource.res.id;
     payload.type = this.data.resource.res.type;
-    payload.deleteComment = this.comment ?? '';
+    payload.deleteComment = this.eraseForm.get('comment')?.value;
```
derschnee68: typesafe

derschnee68 (approved):

PR #1595, opened 2024-05-16
fix: permissions in property values

derschnee68:


irmastnt (approved):
Consider using RouteConstants i.e.RouteConstants.uuidParameter

PR #1600, opened 2024-05-17
feat: assigns new project to the current user automatically, if creat…(DEV-3658)

irmastnt:
feat: assigns new project to the current user automatically, if created from my projects (DEV-3658)

### DIFF
--- a/apps/dsp-app/src/app/project/create-project-form-page/create-project-form-page.component.ts
+++ b/apps/dsp-app/src/app/project/create-project-form-page/create-project-form-page.component.ts
@@ -1,6 +1,6 @@
 import { Location } from '@angular/common';
 import { Component } from '@angular/core';
-import { Router } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { ProjectApiService } from '@dasch-swiss/vre/shared/app-api';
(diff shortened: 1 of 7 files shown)


derschnee68:
I don't think adding a new route for creating a project from user is a good idea.
Shouldn't the creator of a project be assigned as the project member by default ?

irmastnt:
> I don't think adding a new route for creating a project from user is a good idea. Shouldn't the creator of a project be assigned as the project member by default ?

Project may be created for other members as well. Or without members at all.

apps/dsp-app/src/app/project/create-project-form-page/create-project-form-page.component.ts
```diff
@@ -66,6 +66,16 @@ export class CreateProjectFormPageComponent {
         selfjoin: true,
         status: true,
       })
+      .pipe(
+        map(projectResponse => {
+          if (this._router.url.includes(RouteConstants.assignCurrentUser)) {
```
derschnee68, line 71: dont use a map, use a tap or better, add this in subscription

derschnee68 (changes requested):

derschnee68 (approved):

PR #1604, opened 2024-05-21
fix: Various add/edit resource property value bugs (DEV-3667)

irmastnt:
resolves DEV-3666, DEV-3667


libs/vre/shared/app-resource-properties/src/lib/property-values.component.ts
```diff
@@ -31,43 +28,15 @@ import { propertiesTypeMapping } from './resource-payloads-mapping';
...
-        if (_params['project']) {
-          return _params['project'];
-        }
-        throw new Error('Project uuid param not found');
-      })(params, parentParams);
-      const projectIri = this._projectService.uuidToIri(uuid);
-      return ProjectService.IsProjectAdminOrSysAdmin(user, userProjectGroups, projectIri);
-    })
-  );
+  @Input() itemTpl!: TemplateRef<any>;
```
derschnee68, line 33: you can leave the @Input on top of the component, as it is best practice to have them on top

libs/vre/shared/app-resource-properties/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "noImplicitOverride": true,
     "noPropertyAccessFromIndexSignature": true,
     "noImplicitReturns": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "strictPropertyInitialization": false,
```
derschnee68, line 11: this stays true    "strictPropertyInitialization"

derschnee68 (changes requested):

derschnee68 (approved):

PR #1613, opened 2024-05-27
fix: comparison viewer works again

derschnee68:
resolves DEV-3586


apps/dsp-app/src/app/main/guard/ontology-class-instance.guard.ts
irmastnt, line 1: This guard should prevent from creating new resource by members which does not have access to project.
I.e.: user which does not have rights to create resource for project: 3ABR_2i8QYGSIDvmP9mlEw
still can try to do that by going to http://localhost:4200/project/3ABR_2i8QYGSIDvmP9mlEw/ontology/incunabula/Book/add

libs/vre/shared/app-state/src/lib/resource/resource.actions.ts
irmastnt, line 1: Please be aware of using services instead of state, it may have some downsides:
https://stackoverflow.com/questions/63916095/why-do-we-need-ngrx-even-we-can-do-everything-with-services

Also, we have to be consistent.

apps/dsp-app/src/app/workspace/comparison/comparison.component.html
irmastnt, line 1: Remove this file

irmastnt (changes requested):

irmastnt (approved):

PR #1626, opened 2024-05-29
fix: annotations are displayed after opening an image resource

derschnee68:


libs/vre/shared/app-resource-page/src/lib/annotation-tab.component.ts
```diff
@@ -10,16 +11,30 @@ import { Subscription } from 'rxjs';
     *ngFor="let annotation of regionService.regions; trackBy: trackAnnotationByFn"
     [id]="annotation.res.id"
     [class.active]="annotation.res.id === selectedRegion">
-    <app-properties-display [resource]="annotation" [properties]="annotation.resProps" [displayLabel]="true" />
+    <app-properties-display
+      [resource]="annotation"
+      [properties]="annotation.resProps"
+      [displayLabel]="true"
+      [linkToNewTab]="
```
irmastnt, line 18: looks a bit ugly, maybe change to function with string interpolation

irmastnt (approved):

PR #1628, opened 2024-05-30
fix: shows properties in create resource form which does not have gui…

irmastnt:
…order (DEV-3673)


derschnee68:
no longer maintained, close.

PR #1632, opened 2024-06-03
fix: highlighting image region after creating new one (#DEV-3700) 

irmastnt:
fix: highlighting image region after creating new one (#DEV-3700)
prevents from loading resource list if removed resource is region


BalduinLandolt (approved):
Looks reasonable. Is the "canvas-click event doesnt work anymore" something that needs to be investigated?

PR #1638, opened 2024-06-10
fix: some dialogs are displayed as drawers

derschnee68:


apps/dsp-app/src/app/user/create-user-page/create-user-dialog.component.html
irmastnt, line 1: remove this file

irmastnt (changes requested):

irmastnt (approved):

PR #1643, opened 2024-06-11
refactor: split video component with its features

derschnee68:
Add a video-more-button component that handles
- file replacement
- share functionnality
- download


irmastnt (changes requested):
Video preview and video-more is child components of video component. Move them into  video folder.

irmastnt (approved):

PR #1646, opened 2024-06-13
refactor: compact small html and scss components into one file

derschnee68:


irmastnt (commented):
According to official Angular style guide it does make sense to have one file if there is not more than 3 lines: https://angular.dev/style-guide#style-05-04
I dont mind to have slightly bigger files, but those are definitely too big.

derschnee68:
alright, I close it then

irmastnt:
> alright, I close it then

@derschnee68 maybe we need to define what deviations from common conventions we want to have? This could help us to align better.

PR #1651, opened 2024-06-14
fix: display ontology labels in prefffered language (DEV-3739)

mpro7:
resolves DEV-3739

### DIFF
--- a/apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
+++ b/apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
@@ -43,10 +43,6 @@ export class OntologyClassItemComponent implements OnInit, AfterViewInit, OnDest
 
   @ViewChild('resClassLabel') resClassLabel: ElementRef;
 
-  //   get resourceClassType(): ResourceClassDefinitionWithPropertyDefinition {
-  //     return this.resource.res.entityInfo.classes[this.resource.res.type];
-  //   }
-
   get results$(): Observable<number> {
     return combineLatest([
       this._store.select(OntologyClassSelectors.classItems),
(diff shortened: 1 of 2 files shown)


apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
```diff
@@ -66,6 +67,8 @@ export class OntologyClassItemComponent implements OnInit, AfterViewInit, OnDest
     },
   };
 
+  ontologiesLabel: string;
```
derschnee68, line 70: ontologiesLabel?: string;

apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
```diff
@@ -111,6 +115,20 @@ export class OntologyClassItemComponent implements OnInit, AfterViewInit, OnDest
...
 
+  /**
+   * Gets the ontology label in the preferred language of the user
+   */
+  private getOntologiesLabelsInPreferredLanguage(): void {
+    // TODO: this method and changes in below PR are workourund, the label selection should be invastigated deeper to find out why DE is set as default
+    // https://github.com/dasch-swiss/dsp-das/pull/1651
+    const prefferedLanguage = this._store.selectSnapshot(UserSelectors.language);
+
+    if (this.resClass.labels) {
```
derschnee68, line 126: so if this condition is not met, ontologiesLabel stays undefined, is that what we want ? (just asking I don't know what's the right behaviour)

derschnee68 (approved):

apps/dsp-app/src/app/project/ontology-classes/ontology-class-item/ontology-class-item.component.ts
mpro7: I think this is not necessary, because of the guard I added: `this.resClass.labels[0].value`. If there is no match between label languages and preferred language, the first one from the labels array is applied, and at least one label is required AFAIK.

PR #1657, opened 2024-06-19
fix: no duplicates in compound, no isRegionOf property for annotation

derschnee68:


libs/vre/shared/app-common/src/lib/generateProperty.ts
irmastnt, line 1: avoid using string literals, use js-lib constants instead

irmastnt (changes requested):

irmastnt (approved):

PR #1662, opened 2024-06-24
fix(resource permissions): check resource permissions for replacing files

domsteinbach:
resolves DEV-3600

### DIFF
--- a/libs/vre/shared/app-common/src/lib/resource.util.ts
+++ b/libs/vre/shared/app-common/src/lib/resource.util.ts
@@ -1,8 +1,13 @@
-import { PermissionUtil, ReadResource } from '@dasch-swiss/dsp-js';
+import { PermissionUtil, ReadResource, ReadValue } from '@dasch-swiss/dsp-js';
 
 export class ResourceUtil {
-  public static isEditableByUser(resource: ReadResource) {
-    const permissions = PermissionUtil.allUserPermissions(resource.userHasPermission as 'RV' | 'V' | 'M' | 'D' | 'CR');
+  public static isEditableByUser(instance: ReadResource | ReadValue) {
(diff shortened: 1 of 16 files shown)


libs/vre/shared/app-common/src/lib/resource.util.ts
```diff
@@ -0,0 +1,13 @@
+import { PermissionUtil, ReadResource, ReadValue } from '@dasch-swiss/dsp-js';
+
+export class ResourceUtil {
+  public static userCanEdit(instance: ReadResource | ReadValue) {
+    const permissions = PermissionUtil.allUserPermissions(instance.userHasPermission as 'RV' | 'V' | 'M' | 'D' | 'CR');
+    return permissions.indexOf(PermissionUtil.Permissions.M) !== -1;
```
derschnee68, line 6: could you replace it with: return persmissions.includes(PermissionUtil.Permissions.M);

It makes more sense IMO

libs/vre/shared/app-common/src/lib/resource.util.ts
```diff
@@ -0,0 +1,13 @@
...
+  }
+
+  public static userCanDelete(instance: ReadResource | ReadValue) {
+    const permissions = PermissionUtil.allUserPermissions(instance.userHasPermission as 'RV' | 'V' | 'M' | 'D' | 'CR');
+    return permissions.indexOf(PermissionUtil.Permissions.D) !== -1;
```
derschnee68, line 11: same here

derschnee68 (approved):

libs/vre/shared/app-common/src/lib/resource.util.ts
domsteinbach: nicer indeed.

PR #1678, opened 2024-07-03
fix(cardinality change): include other properties, set the new cardinality

domsteinbach:
resolves DEV-3840


domsteinbach:
Short and crispy PR hotfixing two issues:
- include other properties when updating via jsLib's replaceCardinalityOfResourceClass
- actually send a changed cardinality

derschnee68 (approved):

PR #1681, opened 2024-07-05
feat: disables form submit button, if form does not have any changes (DEV-3829)

irmastnt:
resolves DEV-3829

### DIFF
--- a/apps/dsp-app/src/app/project/reusable-project-form/reusable-project-form.component.ts
+++ b/apps/dsp-app/src/app/project/reusable-project-form/reusable-project-form.component.ts
@@ -56,7 +56,10 @@ export class ReusableProjectFormComponent implements OnInit {
   };
   @Output() afterFormInit = new EventEmitter<ProjectForm>();
 
+  private initialData: string;
+
   form: ProjectForm;
+
   readonly shortcodePatternError = {
     errorKey: 'pattern',
     message: 'This field must contain letters from A to F and 0 to 9',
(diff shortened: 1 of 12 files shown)


apps/dsp-app/src/app/project/reusable-project-form/reusable-project-form.component.ts
```diff
@@ -65,13 +68,18 @@ export class ReusableProjectFormComponent implements OnInit {
   readonly keywordsValidators = [Validators.minLength(3), Validators.maxLength(64)];
   readonly descriptionValidators = [Validators.minLength(3), Validators.maxLength(40960)];
 
+  get hasChanges(): boolean {
```
derschnee68, line 71: please don't use this: Angular form knows when the form has changed value.
Something like form.dirty

derschnee68 (changes requested):
in everything you have done, I think you are not leveraging the angular form.

<button [disabled]="form.invalid" /> is enough, no ?

apps/dsp-app/src/app/project/reusable-project-form/reusable-project-form.component.ts
irmastnt: @derschnee68 checking if the form is dirty is not enough, it only check direct fields of the form but not the composite components

irmastnt:
Requirements were changed. Closing this PR.

PR #1688, opened 2024-07-11
fix: reimplement cardinality change

domsteinbach:
resolves DEV-3862, DEV-3872, DEV-3873, DEV-3839, DEV-3837, DEV-3900


derschnee68:
Could you describe the goal and context of this PR Dominique please ?

domsteinbach:
> Could you describe the goal and context of this PR Dominique please ?

There are four or five linear issues linked which are describing what this PR is doing. Here is a short recap:

The whole management of cardinalities has been removed at some point and was refactored to be a simple form control. This leads to errors and bugs only. There is currently no correct way to define cardinalities properly. This PR reintroduces a bug free and correct way to managing the cardinalities

The cardinalities are currently part of the property form - which is just wrong logically and conceptually. There is simply no such a thing like a property having its own cardinality - not in RDF nor in the DSP's data model. A property only has a cardinality on a class. (I mean what should happen if a property is required? Is it required for any class to be implemented? A property can be required to one class while it is not for another class). As the DSP-Api can not store a cardinality of a property without a class the current implementation will always lead to errors. This PR fixes this.

The app does not check whether a cardinality change is possible or not which often leads to errors. This PR reintroduces the use of the canDo checks

As the whole concept of cardinalities is more complex and not a simple property of a property it should definitely not be part of the property form. It is not a simple form control.

There are several other bugs fixed by this PR. With the complete removal from any property form it is also much clearer to a user what the cardinalities are ontologically.

irmastnt (approved):

apps/dsp-app/src/app/project/ontology/resource-class-info/add-property-menu.component.ts
```diff
@@ -135,36 +136,36 @@ export class AddPropertyMenuComponent {
...
+    const addCard = new UpdateResourceClassCardinality();
+    addCard.id = this.resourceClass.id;
+
+    addCard.cardinalities = [
+      {
+        propertyIndex: prop.propDef.id,
+        cardinality: 1,
+        guiOrder: maxGuiOrderProperty + 1,
+      },
+    ];
```
derschnee68, line 160: is cardinality: 1 good ?

derschnee68 (approved):

apps/dsp-app/src/app/project/ontology/resource-class-info/add-property-menu.component.ts
domsteinbach: Yup, simple default which has no side effects whatsoever.
/**
     * Cardinality 0-1 (optional).
     */
    "_0_1" = 1,

PR #1703, opened 2024-07-22
fix: in seach, reload the selected resource after editing

derschnee68:
resolves DEV-3772


libs/vre/shared/app-resource-page/src/lib/resource-fetcher.component.ts
```diff
@@ -24,6 +24,7 @@ export class ResourceFetcherComponent implements OnChanges, OnDestroy {
     this._resourceFetcherService.settings.imageFormatIsPng.next(false);
 
     this._resourceFetcherService.resource$.subscribe(resource => {
+      console.log('fetched', resource);
```
irmastnt, line 27: remove

irmastnt (approved):

PR #1704, opened 2024-07-23
fix: removed appLabels prefix from localization labels

irmastnt:


derschnee68:
@irmastnt could you please close this if it's no longer maintained please ? thanks

PR #1709, opened 2024-07-25
fix: prevents from saving empty property values (DEV-3919)

irmastnt:
resolves DEV-3919

### DIFF
--- a/libs/vre/shared/app-resource-properties/src/lib/create-resource-form.component.ts
+++ b/libs/vre/shared/app-resource-properties/src/lib/create-resource-form.component.ts
@@ -232,7 +232,11 @@ export class CreateResourceFormComponent implements OnInit {
     const propertiesObj: { [index: string]: CreateValue[] } = {};
 
     Object.keys(this.form.controls.properties.controls)
-      .filter(iri => this.form.controls.properties.controls[iri].controls.some(control => control.value.item !== null))
+      .filter(iri =>
+        this.form.controls.properties.controls[iri].controls.some(
+          control => control.value.item !== null && control.value.item !== ''
+        )
+      )
       .forEach(iri => {
         propertiesObj[iri] = this._getValue(iri);
       });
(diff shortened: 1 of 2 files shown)


libs/vre/shared/app-resource-properties/src/lib/create-resource-form.component.ts
```diff
@@ -232,7 +232,11 @@ export class CreateResourceFormComponent implements OnInit {
     const propertiesObj: { [index: string]: CreateValue[] } = {};
 
     Object.keys(this.form.controls.properties.controls)
-      .filter(iri => this.form.controls.properties.controls[iri].controls.some(control => control.value.item !== null))
+      .filter(iri =>
+        this.form.controls.properties.controls[iri].controls.some(
+          control => control.value.item !== null && control.value.item !== ''
```
derschnee68, line 237: this should stays null, because '' could be a valid value in some cases. I will propose a fix

irmastnt: Also, I noticed that an exception is being raised when trying to save 0 as a decimal value:
@type: "http://www.w3.org/2001/XMLSchema#decimal"
@value: "0"

dsp.errors.InconsistentRepositoryDataException: Unexpected object of http://rdfh.ch/0811/LX1Y4tLSRiuxyKgw5ICpgQ/values/zRqfaZPGS0-qgtdQRNOYFA http://www.knora.org/ontology/knora-base#valueHasDecimal: 0",

domsteinbach (approved):
Concerning the review of @derschnee68 : 
I think the behaviour was once that empty strings can not be saved. In that regard all good. That empty strings can not be saved seems wrong to me btw.

libs/vre/shared/app-resource-properties/src/lib/create-resource-form.component.ts
irmastnt: @derschnee68 We are merging this now, but I will leave the ticket open and assign it to you for later improvements.

PR #1713, opened 2024-07-29
fix: create user as system administrator (DEV-3934)

irmastnt:
resolves DEV-3934

### DIFF
--- a/apps/dsp-app/src/app/user/create-user-page/create-user-dialog.component.ts
+++ b/apps/dsp-app/src/app/user/create-user-page/create-user-dialog.component.ts
@@ -56,7 +56,11 @@ export class CreateUserDialogComponent {
     user.systemAdmin = this.form.controls.systemAdmin.value;
     user.status = true;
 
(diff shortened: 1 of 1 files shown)


domsteinbach (approved):
Could be shorter:

`this.projectIri ?  this._projectService.uuidToIri(this.projectUuId): undefined;`

For me strange that uuidToIri() returns a string which might not be the iri.

irmastnt:
> Could be shorter:
> 
> `this.projectIri ? this._projectService.uuidToIri(this.projectUuId): undefined;`
> 
> For me strange that uuidToIri() returns a string which might not be the iri.

this method expects a string not an object

domsteinbach:
> > Could be shorter:
> > `this.projectIri ? this._projectService.uuidToIri(this.projectUuId): undefined;`
> > For me strange that uuidToIri() returns a string which might not be the iri.
> 
> this method expects a string not an object

Yes obviously. But uuidToIri() returns either the iri or strangewise the uuid. Therefore was my remark. It should return the iri, but not the uuid as a fallback.

PR #1721, opened 2024-08-02
fix: property slide toggle instead of an icon (DEV-3752)

irmastnt:
resolves DEV-3752


irmastnt:
replaced with switch toggle component

PR #1722, opened 2024-08-05
feat: implement third party iiif image

domsteinbach:
resolves DEV-3335


libs/vre/shared/app-representations/src/lib/still-image/still-image.component.ts
```diff
@@ -482,31 +477,70 @@ export class StillImageComponent implements OnInit, OnChanges, OnDestroy {
...
+    ) {
+      this._loadExternalIIIF();
+    }
+  }
 
-    if (this.imagesSub) {
-      this.imagesSub.unsubscribe();
+  private _loadInternalImages() {
+    if (this.imageFileValue instanceof ReadStillImageFileValue) {
+      console.log('loading internal image');
```
derschnee68, line 493: remove console log

PR #1723, opened 2024-08-06
refactor: separate display logic from business logic

domsteinbach:
resolves DEV-3699


derschnee68 (approved):

libs/vre/shared/app-resource-properties/src/lib/properties-display.component.ts
```diff
@@ -42,10 +42,15 @@ import { sortByKeys } from './sortByKeys';
...
 
     <!-- list of properties -->
-    <ng-container *ngIf="myProperties$ | async as myProperties">
-      <ng-container *ngIf="myProperties.length > 0; else noProperties">
+    <ng-container>
+      <ng-container *ngIf="properties.length > 0; else noProperties">
         <app-property-row
-          *ngFor="let prop of myProperties; let last = last; trackBy: trackByPropertyInfoFn"
+          [ngClass]="
+            (showAllProperties$ | async) || (prop.values && prop.values.length > 0)
```
irmastnt, line 49: try to reduce code duplication by introducing an observable method which returns the class name

irmastnt (changes requested):

libs/vre/shared/app-resource-properties/src/lib/properties-display.component.ts
domsteinbach: Jup, of course.

irmastnt (approved):

PR #1732, opened 2024-08-14
feat: add mathjax to html value properties

derschnee68:


irmastnt (approved):

libs/vre/shared/app-mathjax/src/lib/math-jax.directive.ts
```diff
@@ -0,0 +1,20 @@
+import { Directive, ElementRef, OnInit } from '@angular/core';
+import { MathJaxService } from './math-jax.service';
+
+@Directive({
+  selector: '[appMathjax]',
+})
+export class MathJaxDirective implements OnInit {
+  constructor(
+    private _ele: ElementRef,
+    private mathJaxService: MathJaxService
```
mpro7, line 10: `mathJaxService` should be probably also underscored.

mpro7 (approved):

PR #1757, opened 2024-09-02
fix: pass projectUuid via input (DEV-4055)

domsteinbach:
resolves DEV-4055


domsteinbach:
Simply solved by passing the projectUuid per inputs, as the components involved will be refactored in that regard anyway with my next PR.

irmastnt (approved):

PR #1774, opened 2024-09-11
feat: prohibit use of internal images (DEV-4092)

domsteinbach:
resolves DEV-4092


libs/vre/shared/app-representations/src/lib/third-party-iiif/third-party-iiif.component.ts
```diff
@@ -56,6 +61,7 @@ export class ThirdPartyIiifComponent implements ControlValueAccessor, OnInit, On
     { errorKey: 'invalidIiifUrl', message: 'The provided URL is not a valid IIIF image URL' },
     { errorKey: 'previewImageError', message: 'The image cannot be loaded from the third party server' },
     { errorKey: 'infoJsonError', message: 'The iiif info json cannot be loaded from the third party server' },
```
mpro7, line 63:
```suggestion
    { errorKey: 'invalidIiifUrl', message: 'The provided URL is not a valid IIIF image URL.' },
    { errorKey: 'previewImageError', message: 'The image cannot be loaded from the third party server.' },
    { errorKey: 'infoJsonError', message: 'The IIIF info JSON cannot be loaded from the third party server.' },
```
I know it's out of the scope, but could you please improve above error messages by applying my suggestion? Thanks :)

mpro7 (approved):

libs/vre/shared/app-representations/src/lib/third-party-iiif/third-party-iiif.component.ts
domsteinbach: Of course, thanks.

PR #1804, opened 2024-09-18
fix: can replace a file in representations (DEV-4118)

irmastnt:
resolves DEV-4118

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [11.20.1](https://github.com/dasch-swiss/dsp-das/compare/v11.20.0...v11.20.1) (2024-09-23)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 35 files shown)


domsteinbach:
Please load the project in the store for cases when there is no project in the store. This is the case when we route to a resource directly or via the search. Then there is no project in the store. We could use resource.attachedToProject and then load it into the store.

irmastnt:
> Please load the project in the store for cases when there is no project in the store. This is the case when we route to a resource directly or via the search. Then there is no project in the store. We could use resource.attachedToProject and then load it into the store.

@domsteinbach  should work now

irmastnt:
> > > Please load the project in the store for cases when there is no project in the store. This is the case when we route to a resource directly or via the search. Then there is no project in the store. We could use resource.attachedToProject and then load it into the store.
> > 
> > 
> > @domsteinbach should work now
> 
> When you search for a resource in a new tab - without having worked within the project before, it does not work to replace a file. Err. Msg displayed in the snackbar: "Cannot read properties of undefined (reading 'shortcode')"

@domsteinbach I have added attachedToProject in the upload component parameters. It works everywhere now.

PR #1832, opened 2024-09-27
fix: shows 403 status error when incoming resource representation cannot be loaded due to lack of permissions (DEV-4146)

irmastnt:
resolves DEV-4146


libs/vre/shared/app-resource-page/src/lib/compound/compound.service.ts
```diff
@@ -57,22 +61,32 @@ export class CompoundService {
     this.compoundPosition.page = page;
   }
 
+  // incoming representation is not provided due to lack of permissions
+  isNextPageAvailable(page: number) {
+    if (!page || !this.compoundPosition) {
+      return false;
+    }
+
```
domsteinbach, line 69: Why can't a user not go to the next page if the viewer is on the last page of the offset of 25?

irmastnt: @domsteinbach I am trying to cover the edge case when BE gives the count of incoming resources but user cant really see it if he does not have permission

domsteinbach (approved):

PR #1835, opened 2024-09-30
feat: accept encoded segments for region, size and rotation (DEV-4140)

domsteinbach:
resolves DEV-4140

### DIFF
--- a/libs/vre/shared/app-representations/src/lib/third-party-iiif/third-party-iiif.ts
+++ b/libs/vre/shared/app-representations/src/lib/third-party-iiif/third-party-iiif.ts
@@ -57,15 +57,15 @@ export class IIIFUrl {
       return false;
     }
 
-    if (!this.rotation_ex.test(this._rotationSegment)) {
+    if (!this.rotation_ex.test(this._rotationSegment) && this.rotation_ex.test(decodeURI(this._rotationSegment))) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


irmastnt (approved):

libs/vre/shared/app-representations/src/lib/third-party-iiif/third-party-iiif.ts
```diff
@@ -57,15 +57,15 @@ export class IIIFUrl {
       return false;
     }
 
-    if (!this.rotation_ex.test(this._rotationSegment)) {
+    if (!this.rotation_ex.test(this._rotationSegment) && this.rotation_ex.test(decodeURI(this._rotationSegment))) {
```
mpro7, line 60:
```suggestion
    if (!this.rotation_ex.test(this._rotationSegment) && !this.rotation_ex.test(decodeURIComponent(this._rotationSegment))) {
```
Although it works you should use `decodeURIComponent` instead and negate the test result, because in fact it decodes the segment correctly. Here is the documentation for encoding function which explain the scope: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
Please update below statements too.

mpro7 (changes requested):

mpro7 (approved):

PR #1855, opened 2024-10-10
fix: the component is reloaded on resource change

derschnee68:


libs/vre/shared/app-resource-page/src/lib/resource-fetcher.component.ts
```diff
@@ -39,11 +38,8 @@ export class ResourceFetcherComponent implements OnChanges, OnDestroy {
         return;
       }
 
-      if (resource.isRegion) {
-        this._renderAsRegion(resource);
-        return;
-      }
-
+      this.resource = undefined;
+      this._cdr.detectChanges();
```
irmastnt, line 42: why double detectChanges ?

irmastnt: @derschnee68 let's have it like this, if it works, at least for now.

irmastnt (approved):

PR #1869, opened 2024-10-22
refactor: property info component (DEV-4240)

domsteinbach:
resolves DEV-4240


irmastnt (approved):

libs/vre/shared/app-ontology/src/lib/property-info/property-info.component.ts
```diff
@@ -53,148 +54,120 @@ export interface ShortInfo {
     ]),
   ],
 })
-export class PropertyInfoComponent implements OnChanges, AfterContentInit {
-  @Input() propDef: ResourcePropertyDefinitionWithAllLanguages;
+export class PropertyInfoComponent implements OnInit {
+  @Input() propDef!: ResourcePropertyDefinitionWithAllLanguages;
```
derschnee68, line 58: add @Input({required: true}) when a resource is a mandatory input

derschnee68 (approved):

libs/vre/shared/app-ontology/src/lib/property-info/property-info.component.ts
domsteinbach: Yup, agree. Will be there.

PR #1883, opened 2024-11-11
fix: remove two way binding (DEV-4339

domsteinbach:
resolves DEV-4339

### DIFF
--- a/libs/vre/shared/app-resource-page/src/lib/compound/compound-viewer.component.ts
+++ b/libs/vre/shared/app-resource-page/src/lib/compound/compound-viewer.component.ts
@@ -1,16 +1,24 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
+import { DspResource } from '@dasch-swiss/vre/shared/app-common';
 import { CompoundService } from './compound.service';
 
 @Component({
   selector: 'app-compound-viewer',
   template: `
     <ng-container *ngIf="compoundService.compoundPosition">
-      <ng-container *ngIf="compoundService.incomingResource as incomingResource">
+      <ng-container *ngIf="incomingResource">
         <app-still-image class="dsp-representation" [resource]="incomingResource.res" [compoundMode]="true" />
       </ng-container>
     </ng-container>
   `,
 })
-export class CompoundViewerComponent {
+export class CompoundViewerComponent implements OnInit {
+  incomingResource: DspResource | undefined;
   constructor(public compoundService: CompoundService) {}
+
+  ngOnInit() {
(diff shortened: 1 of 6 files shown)


irmastnt (approved):

derschnee68 (approved):

libs/vre/shared/app-resource-page/src/lib/resource.component.ts
```diff
@@ -39,10 +37,11 @@ export class ResourceComponent implements OnInit {
   @Input({ required: true }) resource!: DspResource;
   @Input({ required: true }) isDifferentResource!: boolean;
   representationsToDisplay!: ReadFileValue;
-  isCompoundNavigation: boolean | null = null;
+  isCompoundNavigation: boolean = false;
```
derschnee68, line 40: isCompoundNavigation should be undefined (or null) by default

libs/vre/shared/app-resource-page/src/lib/compound/compound-viewer.component.ts
```diff
@@ -1,16 +1,24 @@
...
+import { Component, OnInit } from '@angular/core';
+import { DspResource } from '@dasch-swiss/vre/shared/app-common';
 import { CompoundService } from './compound.service';
 
 @Component({
   selector: 'app-compound-viewer',
   template: `
     <ng-container *ngIf="compoundService.compoundPosition">
-      <ng-container *ngIf="compoundService.incomingResource as incomingResource">
+      <ng-container *ngIf="incomingResource">
```
derschnee68, line 9: <ng-container *ngIf="compoundService.incomingResource$ | async as incomingResource">

libs/vre/shared/app-resource-page/src/lib/compound/compound.service.ts
```diff
@@ -5,14 +5,17 @@ import { IncomingService } from '@dasch-swiss/vre/shared/app-common-to-move';
...
 
 /**
  * Service to handle compound resources, which are resources that are composed of multiple resources.
  */
 @Injectable()
 export class CompoundService {
   compoundPosition?: DspCompoundPosition;
-  incomingResource?: DspResource;
+
+  private _incomingResource = new BehaviorSubject<DspResource | undefined>(undefined);
```
derschnee68, line 17: Could be a Subject, but it's ok otherwise

derschnee68 (approved):

libs/vre/shared/app-resource-page/src/lib/resource.component.ts
domsteinbach: It does not need to. The components display logic should be changed instead imho.

PR #1885, opened 2024-11-12
fix: image download (DEV-4321) (DEV-4318)

irmastnt:
resolves DEV-4321, DEV-4318

### DIFF
--- a/libs/vre/shared/app-representations/src/lib/still-image/still-image.component.ts
+++ b/libs/vre/shared/app-representations/src/lib/still-image/still-image.component.ts
@@ -1,6 +1,7 @@
 import {
   AfterViewInit,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ElementRef,
   Input,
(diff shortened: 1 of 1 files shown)


libs/vre/shared/app-representations/src/lib/still-image/still-image-toolbar.component.ts
```diff
@@ -96,15 +98,21 @@ export class StillImageToolbarComponent {
   }
 
   download() {
-    const projectShort = this._store.selectSnapshot(ProjectsSelectors.currentProject)!.shortcode;
+    const projectShort = this.attachedProject?.shortcode;
```
domsteinbach, line 101: in case of the compound view it is always undefined.

libs/vre/shared/app-representations/src/lib/still-image/still-image.component.ts
```diff
@@ -33,6 +33,7 @@ import { StillImageHelper } from './still-image-helper';
       <app-still-image-toolbar
         *ngIf="isViewInitialized"
         [resource]="resource"
+        [attachedProject]="attachedProject"
```
domsteinbach, line 36: in case of the compound viewer always undefined

domsteinbach (changes requested):
The compound viewer does not pass any attachedProject to the image components. I also would prefer to get the attached project by a state.

irmastnt:
> The compound viewer does not pass any attachedProject to the image components. I also would prefer to get the attached project by a state.

Should be ok now

domsteinbach (approved):

PR #1886, opened 2024-11-13
fix: osd navigator 401 issue (DEV-4317)

irmastnt:
resolves DEV-4317


derschnee68 (approved):

libs/vre/shared/app-representations/src/lib/still-image/still-image.component.ts
```diff
@@ -50,13 +50,13 @@ export class StillImageComponent implements AfterViewInit, OnDestroy {
   isPng = false;
 
   constructor(
-    protected osd: OpenSeaDragonService,
+    protected osdService: OpenSeaDragonService,
```
mpro7, line 53: Why is this protected? Shouldn't also be private?

PR #1953, opened 2024-11-28
fix: allows to download original asset (DEV-4402)

irmastnt:
resolves DEV-4402

### DIFF
--- a/libs/vre/resource-editor/representations/src/lib/archive/archive.component.ts
+++ b/libs/vre/resource-editor/representations/src/lib/archive/archive.component.ts
@@ -58,8 +58,8 @@ export class ArchiveComponent implements OnChanges {
     }
   }
 
-  download(url: string) {
-    this._rs.downloadFile(url);
+  download() {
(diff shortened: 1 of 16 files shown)


libs/vre/resource-editor/representations/src/lib/archive/archive.component.ts
```diff
@@ -58,8 +58,8 @@ export class ArchiveComponent implements OnChanges {
     }
   }
 
-  download(url: string) {
-    this._rs.downloadFile(url);
+  download() {
+    this._rs.downloadProjectFile(this.src.fileValue, this.attachedProject?.shortcode);
```
derschnee68, line 62: for this and all other instances, why can't we add inside: _rs.downloadProjectFile a viariable:
const attachedProject = this._store ...
rather than passing it from the resource-representation.component, and then to every child component until we reach that service ?

irmastnt: I thought resource were not available everywhere. Looks like it might be taken from the store.

derschnee68 (approved):

PR #2027, opened 2025-01-07
test: resource project admin e2e tests with populated data (DEV-4438)

irmastnt:
resolves DEV-4438


mpro7 (commented):
Would `test` fit better as a PR prefix? Also what about all the commented out lines? Do we need them? I would remove all.

irmastnt:
> Would `test` fit better as a PR prefix? Also what about all the commented out lines? Do we need them? I would remove all.

Yes, please add a prefix. 
We can enable skipped tests as soon as proper test data is added from dsp-tools. I prefer to leave it.

mpro7 (approved):

PR #2032, opened 2025-01-10
fix: property form list value item is focused by keypress when mat-me…

irmastnt:
…nu is open in advanced search (DEV-4498)

resolves DEV-


derschnee68:
Maybe try to use a select instead of a menu so that we don't need to code this behaviour

PR #2037, opened 2025-01-14
chore: fix angular-eslint versions and bump node.js on the CLI to v20.9.0

mpro7:
- downgrade angular-eslint deps to v17.3.0
- add Dependabot eslint rules
- set more precise node.js version on the CLI and bump it to 20.9.0


irmastnt (approved):

.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           fetch-depth: 0 # gets additionally all tags which we need
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20.9.0
```
derschnee68, line 21: can't you use a global variable for the node version ?
like node-version: {{NODE_VERSION}}

derschnee68 (changes requested):

derschnee68 (approved):

PR #2038, opened 2025-01-14
fix: each haslinktoclass property is displayed as linked resource

derschnee68:
resolves DEV-4489


mpro7 (approved):

irmastnt (changes requested):
It breaks functionality from: https://linear.app/dasch/issue/DEV-4379/opening-a-segment-of-an-audiovideo-in-new-tab-the-link-says-undefined

irmastnt (approved):

PR #2050, opened 2025-01-21
fix: use getters (DEV-4507)

domsteinbach:
resolves DEV-4507


libs/vre/resource-editor/representations/src/lib/document/document.component.ts
```diff
@@ -72,9 +81,12 @@ export class DocumentComponent implements OnChanges {
...
       );
       this.fileType = this._getFileType(this.src.fileValue.filename);
-      if (this.fileType === 'pdf') {
-        this.elem = document.getElementsByClassName('pdf-viewer')[0];
-      }
+    }
+  }
+
+  ngAfterViewInit() {
+    if (this.fileType === 'pdf') {
```
irmastnt, line 88: create small method to get rid of all duplicates on this file with this piece of code

irmastnt (changes requested):

libs/vre/resource-editor/representations/src/lib/document/document.component.ts
domsteinbach: As this is a bug fix PR, I did not care about any existing duplicates. I just moved two lines into a working correct hook. I actually prefer to simply fix and then refactor in a separate PR. Refactoring this component could go way further.

irmastnt (approved):

PR #2067, opened 2025-01-28
test: remove user from a project (DEV-2328)

irmastnt:
resolves DEV-2328


libs/vre/pages/system/system/src/lib/users/users-list/users-list.component.html
```diff
@@ -20,8 +20,12 @@ <h3 class="mat-body subtitle">
...
       </span>
       <h2 class="mat-headline-6">
-        <span *ngIf="projectUuid">{{ list.length | i18nPlural : itemPluralMapping['member'] }}</span>
-        <span *ngIf="!projectUuid">{{ list.length | i18nPlural : itemPluralMapping['user'] }}</span>
+        <span *ngIf="projectUuid" data-cy="member-count"
+          >{{ list.length | i18nPlural : itemPluralMapping['member'] }}</span
+        >
+        <span *ngIf="!projectUuid" data-cy="user-count"
+          >{{ list.length | i18nPlural : itemPluralMapping['user'] }}</span
+        >
```
mpro7, line 28: Question about formatting. Is this guys what you have agreed on or just a Prettier bug? I see it looks different and weird for me than in other files touched in this PR.
```html
<tag attributes
  >value</tag>
>
```
vs what is in below changes and what looks more readable:
```html
<tag
  attributes>
  <tag2>value</tag2>
</tag>
```

irmastnt: @mpro7 It just exceeds 120 characters in a line which is the rule of prettier

PR #2069, opened 2025-01-29
fix: update paginator style

derschnee68:
new style, overviewed with Daniela :
Screenshot 2025-01-29 at 12 35 36

### DIFF
--- a/libs/vre/ui/ui/src/lib/pager/pager.component.scss
+++ b/libs/vre/ui/ui/src/lib/pager/pager.component.scss
@@ -40,7 +40,10 @@
 }
 
 .paginator-item-range {
-  min-width: 8em;
+  display: flex;
+  flex-direction: row-reverse;
+  padding-right: 16px;
+  font-size: 0.7em;
 }
 
 /* Remove number up/down buttons For Chrome, Edge, and Safari */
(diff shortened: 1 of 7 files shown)


mpro7 (approved):

libs/vre/ui/ui/src/lib/pager/pager.component.scss
```diff
@@ -40,7 +40,10 @@
 }
 
 .paginator-item-range {
-  min-width: 8em;
+  display: flex;
+  flex-direction: row-reverse;
+  padding-right: 16px;
+  font-size: 0.7em;
```
mpro7, line 46: Should we maybe try to make it as a rule, to sort alphabetically styles, which I believe will make them easier to read, especially in longer blocks?

domsteinbach (approved):

libs/vre/ui/ui/src/lib/pager/pager.component.scss
derschnee68: we should discuss this in FE meeting, the best IMO would be to have a linter rule so that it's done automatically. Also this would change a lot with css global classes

PR #2070, opened 2025-01-29
fix: jumps to selected annotation region (DEV-4542)

irmastnt:
resolves DEV-4542


irmastnt:
fixed with https://github.com/dasch-swiss/dsp-das/pull/2074

PR #2086, opened 2025-02-06
refactor: remove unused variables

derschnee68:
resolves DEV-


apps/dsp-app/src/app/app.component.ts
```diff
@@ -38,10 +37,6 @@ export class AppComponent implements OnInit, OnDestroy {
     }
   }
 
-  ngOnDestroy(): void {
-    this._localizationService.destroy();
```
irmastnt, line 42: In general, destroying services is a good practice to avoid memory leaks and reset the state. If this service doesn't have any active subscriptions then it is ok to be removed.

libs/vre/shared/app-helper-services/src/lib/base.service.ts
irmastnt, line 1: Base service should contain a common code for all services. Currently, it only has isDestroyed subject which should be used at the component level to check if the service was destroyed after a tab was closed or the page was refreshed, mostly for subscriptions. Sentry indicates that we have a lot of those kind of leaks and we should keep reducing them. So instead of deleting it, I suggest thinking about in which services we should use it.

irmastnt (approved):

apps/dsp-app/src/app/app.component.ts
derschnee68: it does not

libs/vre/shared/app-helper-services/src/lib/base.service.ts
derschnee68: base service has been deleted, cause not used

PR #2088, opened 2025-02-07
fix: footnotes escape html only once

derschnee68:
Fix the bug of a footnote being html escaped twice (now once)
in ```<footnote content="ESCAPING" />```


derschnee68:
EVERYTHING WORKS EXCEPT WHEN YOU ADD A LINK eg google.fr IN THE FOOTNOTE (bad display)

irmastnt (approved):

libs/vre/ui/ui/src/lib/ck-editor/unescape-html.ts
```diff
@@ -0,0 +1,10 @@
+export function unescapeHtml(str: string) {
```
mpro7, line 1: Does it make sense to extract this single function like that? Would it make more sense to keep it where it was or move it to `CkEditorComponent` or ideally to service which handles footnote (not sure if this exists, but if not creating it for single function doesn't make sense).

mpro7 (approved):
Please find one remark below.

libs/vre/ui/ui/src/lib/ck-editor/unescape-html.ts
derschnee68: it is used in two different nx libs so that seems like the cleanest way

PR #2103, opened 2025-02-11
fix(ResourceLinkDialogComponent): guard against undefined state while subscribing to observable  (DEV-4557)

mpro7:
resolves DEV-4557

### DIFF
--- a/libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.ts
+++ b/libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.ts
@@ -72,10 +72,15 @@ export class ResourceLinkDialogComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit() {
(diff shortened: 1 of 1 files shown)


libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.ts
irmastnt, line 1: It does not solve the issue. In the best-case scenario, an exception should be reproduced locally. Need to investigate why projects are not being provided.

irmastnt (changes requested):

libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.ts
mpro7: Correct. I cannot reproduce this issue, but did further investigation which IMO leads to problem with state management. Will reassign the issue and add more info to the ticket.

PR #2110, opened 2025-02-14
fix: geoname component style (DEV-4588)

irmastnt:
resolves DEV-4588


derschnee68:
What does this PR do apart from linting style things and change the name of one file ?
Is it just the trackFn that fixes the issue ?

irmastnt:
> What does this PR do apart from linting style things and change the name of one file ? Is it just the trackFn that fixes the issue ?

@derschnee68 please check the ticket to see the aim of this fix first. 
For the lining, in ASCII/Unicode, uppercase letters come before lowercase letters; this rule is also used in other app files exactly like it is in this pr. If your linter ignores this rule try to find the cause of it.

derschnee68:
close due to the same issue being fixed by PR #2113

PR #2116, opened 2025-02-17
fix: allows resource linking for project admin (DEV-4551)

irmastnt:
resolves DEV-4551


libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.html
```diff
@@ -3,7 +3,7 @@
 <mat-dialog-content>
   <form
     [formGroup]="form"
-    *ngIf="((isSysAdmin$ | async) === true) || ((usersProjects$ | async)?.length && (isCurrentProjectAdminOrSysAdmin$ | async) === true); else notProjectMember">
+    *ngIf="((isSysAdmin$ | async) === true) || ((usersProjects$ | async)?.length && ((isCurrentProjectAdminOrSysAdmin$ | async) === true || (isCurrentProjectMember$ | async) === true)); else notProjectMember">
```
domsteinbach, line 6: *ngIf="(usersProjects$ | async)?.length; else notProjectMember"

Actually checking the length of this.usersProjects$ would be enough to evaluate the project membership of the user.

But can we use the isCurrentProjectAdminSysAdminOrMember Selector? And can we remove this.usersProjects$ completely and use this._store.select(ProjectsSelectors.currentProject) for setting this.selectedProject?

domsteinbach (approved):

libs/vre/pages/ontology/ontology-classes/src/lib/resource-link-dialog/resource-link-dialog.component.html
irmastnt: Possibly, we can improve it in the next release

PR #2127, opened 2025-02-18
chore: e2e tests can have code coverage option

derschnee68:


mpro7 (commented):
Looks good, but it would be also nice add new Dependabot group to keep deps updates organized. Also, did you think about adding the coverage to GH action?

domsteinbach (approved):

PR #2132, opened 2025-02-24
refactor: fix some tsc issues in resource editor

derschnee68:
resolves DEV-


libs/vre/core/error-handler/src/lib/js-lib-parsed-error.ts
```diff
@@ -0,0 +1,5 @@
+export class JsLibParsedError extends Error {
```
mpro7, line 1: Typo? Present tense form sounds more natural.
```suggestion
export class JsLibParseError extends Error {
```

domsteinbach (approved):

PR #2134, opened 2025-02-24
fix: highlights value from route and redesign the properties in the resource editor

derschnee68:
resolves DEV-4625


libs/vre/resource-editor/resource-properties/src/lib/property-value.component.scss
```diff
@@ -1,13 +1,25 @@
 @use '../../../../../../apps/dsp-app/src/styles/config' as *;
 
-.item {
+:host ::ng-deep .app-progress-indicator {
```
domsteinbach, line 3: :: ng-deep?!

domsteinbach (approved):
Can you get the desired result without the use of ng-deep?
It would also be great if we get rid of the inline styles - or at least definitely not mix them with the styles in one component.

PR #2138, opened 2025-02-26
fix: resource editor manages its own resource deletion

derschnee68:


libs/vre/resource-editor/resource-editor/src/lib/resource-fetcher.component.ts
```diff
@@ -1,52 +1,65 @@
...
 
   ngOnChanges() {
+    this.loading = true;
     this._resourceFetcherService.onDestroy();
     this._resourceFetcherService.onInit(this.resourceIri);
 
-    if (this._subscription) {
-      this._subscription.unsubscribe();
-    }
+    this._unsubscribe();
```
domsteinbach: This _unsubscribe() seems wrong to me. If it works, then only because the resourcefetcher needs more time for the request.

Imho there is also no need to destroy anything or unsubscribe. Subscribing to this._resourceFetcherService.resource$ could happen once. It must happen onInit or in the constructor though.

domsteinbach (approved):

PR #2148, opened 2025-03-04
fix: remove unused translations (DEV-4598)

mpro7:
resolves DEV-4598

All locale files hass been cleaned up and aligned, so they are ready now to get translated.


derschnee68 (approved):

BalduinLandolt (approved):
LGTM, you can merge this as is and address https://linear.app/dasch/issue/DEV-4676/investigate-unused-translations later, or include that here as well, as you prefer

PR #2161, opened 2025-03-06
fix: bring back, fix and improve some translations (DEV-4676)

mpro7:
resolves DEV-4676


libs/vre/pages/project/project/src/lib/edit-project-form-page.component.ts
```diff
@@ -54,17 +55,18 @@ export class EditProjectFormPageComponent {
   );
 
   constructor(
+    private _actions: Actions,
+    private _ar: ActivatedRoute,
```
derschnee68, line 59: we use route in the whole app, keeping it would stay consistent

derschnee68 (approved):

PR #2172, opened 2025-03-12
add RM translations

mpro7:
resolves DEV-


editio (dismissed):
ok, 👍, finished now with the Italian version

mpro7:
RM translations will be provided most probably by the other party...

PR #2192, opened 2025-03-20
docs: add legal settings

danielasubotic:


BalduinLandolt (approved):
In the long term, it will be good to make a clear distinction what is defined for the entire project, and what is defined for each asset. But I think this will be easier to convey, once FE is implemented and we have screenshots to illustrate.

We should eventually also have the cardinalities (1 for copyright and license; 1-n for authorship) in the documentation. But I suppose this can wait too.

PR #2194, opened 2025-03-24
fix: permissions in the resource editor only depend on the resource and not the store

derschnee68:


derschnee68:
already managed in main

derschnee68:
already implemented in main

PR #2201, opened 2025-03-25
refactor: improve row styling of resource editor

derschnee68:
resolves DEV-

### DIFF
--- a/libs/vre/resource-editor/resource-properties/src/lib/property-value.component.scss
+++ b/libs/vre/resource-editor/resource-properties/src/lib/property-value.component.scss
@@ -8,18 +8,6 @@
   position: relative;
 }
 
-.resource-editor-value {
-  flex: 1;
-
-  &.highlighted {
-    background: $yellow_100;
-  }
-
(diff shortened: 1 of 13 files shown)


libs/vre/resource-editor/resource-properties/src/lib/property-value.component.scss
```diff
@@ -8,15 +8,16 @@
   position: relative;
 }
 
-.resource-editor-value {
+.value {
   flex: 1;
+  padding: 16px;
```
domsteinbach, line 13: border-radius: 12px; would be nice. can you do that?

domsteinbach (approved):
two minor wishes for change in the comments.

libs/vre/resource-editor/resource-properties/src/lib/property-value.component.scss
derschnee68: that's styling issue that PM needs to decide (I personnaly don't like it haha)

PR #2202, opened 2025-03-27
refactor: improve resource fetcher logic

derschnee68:
- resource.isDeleted logic is only handled in resource-fetcher.component (delete other places)
- translation is removed from ResourceFetcherService
- resourceFetcherService logic is simplified
- resourceFetcherComponent handles notFound and Deleted cases


libs/vre/pages/ontology/ontology-classes/src/lib/list-view/list-view.component.ts
```diff
@@ -93,12 +103,13 @@ export class ListViewComponent implements OnChanges, OnInit, OnDestroy {
...
     private _cd: ChangeDetectorRef,
-    private _store: Store
+    private _store: Store,
+    private _translateService: TranslateService
   ) {}
 
   ngOnInit(): void {
     this.componentCommsSubscriptions.push(
-      this._componentCommsService.on([Events.resourceLanguageChanged], () => this.doSearch()),
+      this.translation$.subscribe(() => this.doSearch()),
```
domsteinbach, line 112: unsubscribe onDestroy?
ngOnDestroy(): void {
  this.componentCommsSubscriptions.forEach(sub => sub.unsubscribe());
}


And we could remove this.ngUnsubscribe by replacing all takeUntil(this.ngUnsubscribe)) with take(1) I think.

libs/vre/resource-editor/representations/src/lib/resource-fetcher.service.ts
```diff
@@ -1,77 +1,43 @@
...
-import { BehaviorSubject, Subject, Subscription } from 'rxjs';
-import { map, switchMap, take } from 'rxjs/operators';
+import { combineLatest, Observable, ReplaySubject, Subject } from 'rxjs';
+import { map, startWith, switchMap } from 'rxjs/operators';
 
 @Injectable()
 export class ResourceFetcherService {
-  private _resourceIri!: string;
-  private _loadResourceSubject = new BehaviorSubject(null);
+  private _loadResourceSubject = new ReplaySubject<Observable<DspResource>>(1);
```
domsteinbach, line 12: Does replay an erased resource. When the fetcher onInits after erasing, it requests the resource and gets a 404 as the erased resource can not be requested anymore. Due to the errors the whole list and the component get stuck somehow and no navigation is possible anymore.

domsteinbach (changes requested):
Could you also add 

`[disabled]="resource.res.isDeleted"`

to the edit label button in the resource header component? Any attempt to edit a deleted resources label will end in an error.

domsteinbach (approved):

PR #2228, opened 2025-04-08
feat: the user can search for lists in the list view of resource editor

derschnee68:
resolves DEV-4777


libs/vre/resource-editor/resource-properties/src/lib/viewer-components/list-viewer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!this.control.value) {
+      return;
+    }
+
+    this.labels$ = (this._dspApiConnection.v2.list.getNode(this.control.value) as Observable<ListNodeV2>).pipe(
+      switchMap(v => this._dspApiConnection.v2.list.getList(v.hasRootNode!)),
+      map(v => {
+        const tree = ListViewerComponent.lookFor([v as ListNodeV2], this.control.value);
+        const finalTree = tree as ListNodeV2[];
+        const nodeId = finalTree[finalTree.length - 1].id;
```
domsteinbach, line 52:
```
const tree = ListViewerComponent.lookFor([v as ListNodeV2], this.control.value) as ListNodeV2[];
        const nodeId = tree[tree.length - 1].id;
```

PR #2253, opened 2025-04-25
feat: add copyright holders, licenses and authorship to file values

derschnee68:
resolves DEV-


libs/vre/resource-editor/resource-properties/src/lib/value-components/link-value.component.ts
```diff
@@ -137,13 +131,14 @@ export class LinkValueComponent implements OnInit, AfterViewInit, OnDestroy {
   openCreateResourceDialog(event: any, resourceClassIri: string, resourceType: string) {
     let myResourceId: string;
     event.stopPropagation();
-    const projectIri = (this._store.selectSnapshot(ProjectsSelectors.currentProject) as ReadProject)?.id;
+    const project = this._store.selectSnapshot(ProjectsSelectors.currentProject)!;
```
domsteinbach, line 134: undefined if routed to the resource without prior project context, e.g. https://app.dev.dasch.swiss/resource/0838/WGp1GewbTOONchBiW6q6wg

domsteinbach (changes requested):
The project in the link-value.component.ts
`const project = this._store.selectSnapshot(ProjectsSelectors.currentProject)!; ` is undefined when routed to
https://app.dev.dasch.swiss/resource/0838/WGp1GewbTOONchBiW6q6wg. A new resource can not be created and linked.

The css/styles are set to values. Can we use the centrally defined ones? Translations would be nice as well.

libs/vre/resource-editor/resource-properties/src/lib/authorship-form-field.component.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  removeItem(item: string) {
+    this.selectedItems = this.selectedItems.filter(i => i !== item);
+    this._updateFormControl();
+  }
+
+  private _addItem(inputValue: string) {
+    this.chipInput.clear();
+
+    if (this.selectedItems.includes(inputValue)) {
```
domsteinbach, line 93: can we change that into:
    if (!inputValue || this.selectedItems.includes(inputValue)) {
    
This prevents a tab to add an empty chip which then will lead to a bad request

domsteinbach (approved):

PR #2277, opened 2025-05-06
fix: name and family name of the resource author is displayed as a priority (DEV-4837)

irmastnt:
resolves DEV-4837


libs/vre/resource-editor/resource-editor/src/lib/resource-info-bar.component.ts
```diff
@@ -19,9 +19,9 @@ import { filter, map, take } from 'rxjs/operators';
         <span *ngIf="resourceAttachedUser"
           >by
           {{
-            resourceAttachedUser.username
-              ? resourceAttachedUser.username
-              : resourceAttachedUser.givenName + ' ' + resourceAttachedUser.familyName
+            resourceAttachedUser.givenName || resourceAttachedUser?.familyName
```
mpro7, line 22: I bet both `givenName` and `familyName` should be both presented.
```suggestion
            resourceAttachedUser?.givenName ?? resourceAttachedUser?.familyName
```

irmastnt: @mpro7 Yes, I tested it manually. The missing fields will not be displayed, only the present names. If both are missing, the username will be displayed.

mpro7 (approved):

PR #2279, opened 2025-05-08
Feature/dev 4854 redesign landing page

domsteinbach:
resolves DEV-4854


irmastnt:
image

derschnee68:
it is normal behaviour that some tiles are broken. we don't have all the images yet

domsteinbach:
I also dislike the 404s but everything else seems an overhead for me. So I am fine

domsteinbach:
Concerning the 404's:  We could use the logo property of a Project aka ReadProject, which is not used by anyone imho. We should check that though. With that we would have a neat solution.

Every other solution I dislike because it is an overhead of additional code or manual kept lists. So I accept the fact that there are 404's until the backend delivers new data or we use the logo property. 
I like the generic dummy template and I dislike dummy images which we also need to be maintained and there is no difference between a dummy image and a real image. Simpler is: If there is an image, nice, if not, template.

PR #2306, opened 2025-05-20
fix: resource editor displays a message when resource access is unauthorized

derschnee68:
resolves DEV-


apps/dsp-app/src/assets/i18n/en.json
```diff
@@ -160,6 +160,7 @@
     "versionNotValid": "The version you requested is not valid. The latest version is displayed.",
     "notFound": "This resource is not found.",
     "deleted": "This resource has been deleted.",
+    "unauthorized": "You do not have access to this resource.",
```
mpro7, line 163:
```suggestion
    "unauthorized": "You do not have the access to this resource.",
```
You can also apply this little change to other locale files, but don't have too (as they going to be translated anyway).

mpro7 (approved):
Check small improvement below before merging:

PR #2312, opened 2025-05-26
feat: Add Project Resources Metadata export feature (DEV-4938)

mpro7:
resolves DEV-4938

The is the first implementation of Project Resources Metadata export feature, that contains:
- static page with download button,
- translations of supported languages,
- metadata are saved now only to CSV format, but it is prepared to do TSV and JSON too.


mpro7:
> Looks good as far as I can tell. My only question: Should there be the english version of the text added to `it.json` and `rm.json`, or will this work correctly already?

I actually forgot to paste it there, but nothing is braked if I switch the language to missing ones, but I will update these locale files. Thanks for spotting!

mpro7:
> Looks good as far as I can tell. My only question: Should there be the english version of the text added to `it.json` and `rm.json`, or will this work correctly already?

I actually forgot to paste it there, but nothing is broken. If you switch the language to missing ones, the default should apply, which is EN, but I update missing locale files. Thanks for spotting!

PR #2331, opened 2025-06-06
feat: make metadata export user friendly and improve the code (DEV-4938)

mpro7:
Adds progress indicator while downloading the metadata and SnackBars to inform about success or failure.


BalduinLandolt (approved):
I don't feel qualified to really say anything about this PR - I'm laxcking knowledge of both Angular and how things are done in the App. But it looks reasonable

PR #2350, opened 2025-06-17
fix: does not display default authorhships if they are already used 

derschnee68:
resolves DEV-


irmastnt (changes requested):
Translations missing

PR #2468, opened 2025-08-11
fix(resource editor): load resource logic(DEV-4895)

domsteinbach:
resolves DEV-4895


libs/vre/core/state/src/lib/resource/resource.state.ts
```diff
@@ -106,7 +106,7 @@ export class ResourceState {
   }
 
   @Action(SetCurrentResourceAction)
-  setCurrentOntologyAction(ctx: StateContext<ResourceStateModel>, { resource }: SetCurrentResourceAction) {
+  setCurrentResourceAction(ctx: StateContext<ResourceStateModel>, { resource }: SetCurrentResourceAction) {
```
domsteinbach, line 109: Does this do anything at all?

PR #2479, opened 2025-08-21
feat: improve the project view with better selection

derschnee68:
resolves DEV-5364
resolves DEV-5365
resolves DEV-5372
resolves DEV-5373
resolves DEV-5374

## Features
- Improve the view of data browser (list of resources within a resource class)
- A user is now able to select resources from diferent pages

## Refactor
- remove project, ontologies, ontology-class, resource stores
- add a ProjectPageService so that the project is accessible everywhere


mpro7 (approved):
Overall looks good. There are some empty SCSS files created which you could remove. 

What we didn't talk about and what should be checked while testing on DEV server is performance testing before merging. If it is not affected, it is ready to merge (some tests are keep failing BTW).

PR #3039, opened 2026-04-23
chore: use common stub factories in storybook

derschnee68:


derschnee68:
@claude review

domsteinbach (approved):
Maybe rm the unused export

projectApiServiceStub is exported but unused

PR #3102, opened 2026-05-11
feat(advanced search): enable escaping and searching of regex

domsteinbach:
As the operator "is like" is parsed and executed as regex we need to enable escaping of regex meta characters in order to search also for literal characters used by regex. 

Example:

A user wants to search for a literal Asterisk "" . Then the input should be: one escape backslash to escape the regex char "" -> "\*"

The app then needs to send the query as a quadruple escape: "\\\\" (every layer needs its own escaping …)

In order to search for a literal quote string '"' the user also can escape it once, e.g. '\"'. The app however does only need to triple the escape for this case (the regex layer is not applied) so the quote char survives


domsteinbach:
@claude review

mpro7 (approved):

PR #3106, opened 2026-05-13
refactor(advanced-search): use cascading nested menu (DEV-6523)

domsteinbach:
DEV-6523

Move NestedMenuComponent from libs/vre/resource-editor/ to a new shared UI library libs/vre/ui/nested-menu/ (pure move, no behaviour change) and consume it from both the Resource Editor and the Advanced Search list-value picker.

Advanced Search previously used mat-autocomplete with a recursive template that flattened the hierarchy into an indented list — long wrapping labels overlapped neighbouring options and depth was only conveyed by leading whitespace. The Resource Editor already solved this with a cascading mat-menu component; this change promotes that component to libs/vre/ui/ so both call sites consume it from one canonical source.


mpro7:
@claude review

mpro7 (approved):

PR #3164, opened 2026-06-11
fix: prevent destruction if the resource does not change (DEV-6611)

domsteinbach:
Fixes the issue that child components get destroyed after updating or adding property values for two types of resources:
- plain resources
- compound resources

(All other resource types were not affected). Destruction and nulling of the type forced a reload of the resource and the properties, so users lost their focus after updating a property.

The fix: We compare the id/iri of the resource and only reset and destroy when the resource changes.

If the new resource has the same id/iri it must have the same type (we can not change a resources type), so we don't need to null and reset the type, so the child components do not get destroyed and the changes run through without destroying and reloading the properties.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,10 +180,10 @@ jobs:
         include:
           # Heavy runner: resource.cy.ts (many API requests per test), data-model-class.cy.ts (DB reset per test), plus smaller specs to balance load
           - runner: heavy
-            specs: cypress/e2e/system-admin/resource.cy.ts,cypress/e2e/system-admin/resource-edit-preserves-view.cy.ts,cypress/e2e/system-admin/data-model-class.cy.ts,cypress/e2e/system-admin/still-image.cy.ts,cypress/e2e/system-admin/file-representation.cy.ts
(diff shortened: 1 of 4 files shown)


mpro7:
@claude review

.github/workflows/ci.yml
```diff
@@ -180,7 +180,7 @@ jobs:
         include:
           # Heavy runner: resource.cy.ts (many API requests per test), data-model-class.cy.ts (DB reset per test), plus smaller specs to balance load
           - runner: heavy
-            specs: cypress/e2e/system-admin/resource.cy.ts,cypress/e2e/system-admin/data-model-class.cy.ts,cypress/e2e/system-admin/still-image.cy.ts,cypress/e2e/system-admin/file-representation.cy.ts
+            specs: cypress/e2e/system-admin/resource.cy.ts,cypress/e2e/system-admin/resource-edit-preserves-view.cy.ts,cypress/e2e/system-admin/data-model-class.cy.ts,cypress/e2e/system-admin/still-image.cy.ts,cypress/e2e/system-admin/file-representation.cy.ts
```
mpro7: `cypress/e2e/system-admin/resource-edit-preserves-view.cy.ts` should be moved from "heavy" to "rest" test runner.

mpro7 (approved):

PR #3195, opened 2026-06-26
chore: migrate esbuild application builder, drop ngx-build-plus (DEV-6687)

mpro7:
[DEV-6687](https://linear.app/dasch/issue/DEV-6687)

## Summary
- Migrate the `dsp-app` production build from the **deprecated** `@angular-devkit/build-angular:browser` (webpack) builder to `@angular-devkit/build-angular:application` (esbuild — equivalent to `@angular/build:application`, re-exported, so no new dependency). The `:browser` builder is deprecated and slated for removal in a future Angular major.
- Remove `ngx-build-plus` entirely. Its only remaining use was the **local-only** e2e coverage instrumentation (`serve-app-e2e` + `webpack.coverage.js`), which is not run in CI.
- Output stays flat (`outputPath.browser: ""`), so the Dockerfile (`COPY ./dist/apps/dsp-app`) and the `sentry:sourcemaps` script are unchanged. Sentry/Faro observability is unaffected — hidden production sourcemaps are still emitted for `sentry-cli` to inject + upload.

## Changes
**Build migration — esbuild application builder** (`apps/dsp-app/project.json`)
- `build` executor `:browser` → `:application`; `main` → `browser`; `polyfills` string → array; `outputPath` → `{ base, browser: "" }` to keep `dist/apps/dsp-app` flat.
- Removed webpack-only options (`vendorChunk` / `namedChunks` / `buildOptimizer` / `aot`) from **all** build configurations (`development`, `test-server`, `dev-server`, `ls-test-server`, `stage-server`).
- Extended `allowedCommonJsDependencies` to cover esbuild's CommonJS-bailout warnings (`reflect-metadata`, `jsonld`, `@opentelemetry/api`, `@opentelemetry/semantic-conventions`, `.../trace/internal-types`).
- `serve` / `extract-i18n` stay on `@angular-devkit/build-angular:dev-server` / `:extract-i18n` — the officially supported pairing that auto-detects the application target and serves via Vite.

**Drop ngx-build-plus + webpack e2e coverage instrumentation**
- Removed the `serve-app-e2e` target and the `e2e:coverage` configuration.
- Deleted `webpack.coverage.js`, `tools/lib-paths.js`, `tools/lcov-parser.js`.
- Removed the `e2e-coverage` npm script and the `@cypress/code-coverage` wiring in `cypress.config.ts` / `cypress/support/e2e.ts`.
[...]

PR #3198, opened 2026-06-26
fix: CJS interop fixes after application builder upgrade

domsteinbach:
Follow-up to #3195 (`migrate esbuild application builder, drop ngx-build-plus`).
The new `@angular-devkit/build-angular:application` executor uses esbuild for
`build` and Vite for `ng serve`. Both follow strict ESM interop for CommonJS
modules — the old webpack-based builder silently collapsed those imports.
Three places in the codebase were relying on the old behaviour. This PR fixes
all of them.

### 1. `libs/dsp-js/src/api/v2/**` — drop CJS `require` of jsonld

Five v2 endpoints replace

```ts
declare let require: any;
const jsonld = require('jsonld/dist/jsonld.js');
```

with a single static ESM import:

```ts
import jsonld from 'jsonld/dist/jsonld.js';
```

### 2. `libs/vre/ui/ui/src/lib/ck-editor/ck-editor.component.ts` — unwrap CKEditor default export

`ckeditor5-custom-build` is a UMD bundle that assigns `ClassicEditor` to
`module.exports` (and ends with `module.exports = module.exports.default`).
`import * as Editor` collapsed to the class under webpack; under esbuild it
becomes `{ default: ClassicEditor, ... }` and `editor.create(...)` throws
`TypeError: this.editor.create is not a function` the moment a rich-text
value is edited. Replaced with a defensive `EditorNs.default ?? EditorNs`
unwrap that works under either builder.

### 3. `libs/vre/resource-editor/.../osd-drawer.service.ts` — align openseadragon import

Same CJS-default-export shape as ckeditor. Two sibling files in the same
folder already use the default-import form (`import OpenSeadragon from 'openseadragon'`);
this one was still on `import * as OpenSeadragon`, a latent regression that
would only fire when drawing regions over an image. Aligned with the
siblings.

### Consumers of `@dasch-swiss/dsp-js` on npm

No behavioural change. Bundler-based consumers (webpack, esbuild, Rollup,
Vite) get the same binding as before. Pure-ESM consumers (Node ESM, Deno) —
who previously hit `Dynamic require of "jsonld/dist/jsonld.js" is not
supported` — now work too. The published JS is now standards-compliant ESM
instead of ESM-with-embedded-`require()`-calls.

### This repo

- Local `npm run start-dev` boots again on a clean tree (no more `Dynamic
  require of jsonld/dist/jsonld.js`).
- Editing any rich-text value (and indirectly the parent forms that mount
  CKEditor) works again, locally and on `app.dev.dasch.swiss` once deployed.
- Region drawing on still images is protected against the same regression.

## Technical details

The published `@dasch-swiss/dsp-js` package is compiled with `module: ES2020`
and ships as ESM. The old `require()` pattern emitted literal `require()`
calls inside that ESM output — invalid by the spec, but tolerated because
every known consumer bundled the result, and bundlers rewrite `require()` at
build time. Vite (introduced as the local dev server by #3195) serves modules
as native browser ESM and does not rewrite `require()` calls — so the same
literal that bundlers silently fixed now reaches the browser unchanged and
throws.

The CKEditor / OpenSeadragon issue is the symmetric case: those UMD modules
expose the class via `module.exports`, and webpack's `import * as`
implementation collapsed the namespace to the class. esbuild follows the
spec — `import * as X` on a CJS module yields a Module Namespace Object with
the CJS shape as the `default` key. Whatever the old code accessed on `X`
directly is now on `X.default`. Defensive unwrap (`X.default ?? X`) covers
both behaviours.

`esModuleInterop: true` in `tsconfig.base.json` is the load-bearing setting
throughout — it lets the static `import jsonld from …` resolve to the same
`module.exports` object that `require()` returned, and lets the default
import on openseadragon resolve to the class.

### DIFF
--- a/libs/dsp-js/src/api/v2/list/lists-endpoint-v2.ts
+++ b/libs/dsp-js/src/api/v2/list/lists-endpoint-v2.ts
@@ -1,10 +1,8 @@
+import jsonld from 'jsonld/dist/jsonld.js';
 import { catchError, map, mergeMap } from 'rxjs';
 import { AjaxResponse } from 'rxjs/ajax';
 import { ListNodeV2 } from '../../../models/v2/lists/list-node-v2';
 import { Endpoint } from '../../endpoint';
-
-declare let require: any; // http://stackoverflow.com/questions/34730010/angular2-5-minute-install-bug-require-is-not-defined
-const jsonld = require('jsonld/dist/jsonld.js');
 
 /**
  * Handles requests to the lists route of the Knora API.
--- a/libs/dsp-js/src/api/v2/ontology/ontologies-endpoint-v2.ts
+++ b/libs/dsp-js/src/api/v2/ontology/ontologies-endpoint-v2.ts
@@ -1,3 +1,4 @@
+import jsonld from 'jsonld/dist/jsonld.js';
 import { catchError, map, mergeMap } from 'rxjs';
 import { AjaxResponse } from 'rxjs/ajax';
 import { ApiResponseError } from '../../../models/api-response-error';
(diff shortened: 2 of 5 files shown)


domsteinbach:
Renamed branch and broadened scope — superseded by a new PR. See follow-up.

PR #3230, opened 2026-07-07
fix(legal): correct the project Rights Statement display and holder clearing

danielasubotic:
Fixes [DEV-6718](https://linear.app/dasch/issue/DEV-6718), [DEV-6720](https://linear.app/dasch/issue/DEV-6720) and [DEV-6722](https://linear.app/dasch/issue/DEV-6722)

## Motivation

Three correctness fixes to the project **Data view / legal settings** "Resource Rights Statement":

1. **Authorship leaked onto the project view** (DEV-6718). The authorship row rendered on the project card and dialog. Authorship is per-resource; at the project level it's only a default template, not a property of the whole project.
2. **License and copyright holder weren't independent** (DEV-6720). The whole statement was gated on having a license, so a project with only a copyright holder (no license) showed nothing.
3. **Couldn't clear the copyright holder** (DEV-6722). Once set, clearing the holder in Settings → Legal and saving didn't persist.

## Summary

- **Authorship**: new `showAuthorship` input (default `true`); the two project surfaces pass `[showAuthorship]="false"`. Resource views keep showing authorship.
- **License / holder independence**: `configured` is now `licenseLabel || copyrightHolder`; the License row is guarded by its own `@if (licenseLabel)`; the two project-surface `@if`s include `rights.copyrightHolder`. The admin "uncategorized" callout appears only when neither is set.
- **Clearing the holder**: the settings form now omits an empty/whitespace holder from the PUT so the (replace-semantics) endpoint clears it via `None` — exactly as the license clears through its `null` "none" option. Previously it sent `dataCopyrightHolder: ""`, which the backend `CopyrightHolder` value object rejects (`nonEmpty`) → 400.

## Key Changes

- `resource-rights-statement.component.ts`: `showAuthorship` input + gated authorship row; `configured = licenseLabel || copyrightHolder`; License row guarded by `@if (licenseLabel)`; doc comments.
- `project-description-page.component.html` & `project-short-description.component.ts`: `[showAuthorship]="false"`; outer `@if` broadened to include `rights.copyrightHolder`.
- `resource-side-legal-form.component.ts`: omit an empty copyright holder from the update payload so it can be cleared.
- `resource-rights-statement.stories.ts`: regression stories — `HidesAuthorshipOnProjectLevelDisplay`, `ShowsCopyrightHolderOnlyWhenNoLicense`, `ShowsLicenseOnlyWhenNoCopyrightHolder`.

## Challenges and Decisions

- **Explicit `showAuthorship`** over reusing context state — keeps resource-side behaviour untouched and the project-level intent obvious; breaks no existing stories.
- **Clear via omission**: the resource-side PUT replaces, so a missing field becomes `None` and clears. The backend already supports this — the fix is purely to stop sending `""`.

## Gotchas

- The `defaultResourceAuthorship` binding on the two project surfaces was removed (dead once authorship is hidden there).
- The resource-side container and `ProjectDataRightsService` need no changes — the container delegates visibility to `configured`, and the service derives holder/license independently.

## Test Plan

- **Storybook**: existing stories stay green; new stories assert authorship hidden at project level, holder-only renders the holder row, license-only renders the license row.
- **Manual** (project Data view, card + "Read more"): license-only → license; holder-only → holder; both → both; neither → nothing (admins see the callout); no Authorship row in any case. On a resource → Authorship still renders.
- **Manual** (Settings → Legal): set a copyright holder, save; then clear it, save → it is removed and stays blank on reload.

_Note: the PR preview serves unhashed bundles with a long cache — hard-refresh (Cmd+Shift+R) after each deploy to see changes._

### DIFF
--- a/libs/vre/pages/project/project/src/lib/description/project-description-page.component.html
+++ b/libs/vre/pages/project/project/src/lib/description/project-description-page.component.html
@@ -48,7 +48,7 @@
             [licenseLabel]="rights.licenseLabel"
             [licenseUrl]="rights.licenseUrl"
             [copyrightHolder]="rights.copyrightHolder"
-            [defaultResourceAuthorship]="rights.defaultDataAuthorship"
+            [showAuthorship]="false"
             [isAdmin]="(hasProjectAdminRights$ | async) ?? false"
             (editLegalInfo)="goToLegalSettings()" />
         </section>
--- a/libs/vre/pages/project/project/src/lib/description/project-description-page.component.html
+++ b/libs/vre/pages/project/project/src/lib/description/project-description-page.component.html
@@ -41,7 +41,7 @@
 
     <!-- Data-side Resource Rights Statement (project-level). Distinct from the hand-typed caption above. -->
     @if (dataRights$ | async; as rights) {
-      @if (rights.licenseLabel || (hasProjectAdminRights$ | async)) {
+      @if (rights.licenseLabel || rights.copyrightHolder || (hasProjectAdminRights$ | async)) {
         <mat-divider style="margin: 24px 0" />
         <section class="project rights">
           <app-resource-rights-statement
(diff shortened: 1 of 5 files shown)


derschnee68:
@claude review

domsteinbach (approved):

PR #3242, opened 2026-07-17
fix: clarify authorship input requires Enter, commit pending value on blur (DEV-6761)

danielasubotic:
Fixes https://linear.app/dasch/issue/DEV-6761

## Motivation

Reported via customer support (Plain thread T-192): in the legal-metadata **Authorship** field, a typed author name was silently dropped unless the user pressed **Enter**, and on *Legal Settings → Resource side* the **Save changes** button stayed greyed out because typing never dirtied the form. It is not a functional bug — the "press Enter" affordance was invisible.

## Summary

Make committing an author obvious and safe wherever authorship is entered: commit a pending value on blur, and show a persistent hint. The same commit-on-Enter-only pattern lived in three chip inputs; this brings them in line with the existing `chip-list-input` component, which already uses `matChipInputAddOnBlur`.

## Key Changes

- **`authorship-chip-editor`** (shared `ui/ui`): add `[matChipInputAddOnBlur]="true"` + an optional `hint` input. Reused by **both** the Legal Settings resource-side form and the data-resource creator, so one change covers both surfaces.
- **`resource-rights-statement`** (per-resource inline editor): add `addOnBlur` + a `mat-hint`. Clicking the Save (💾) icon now flushes the pending name before `saveEdit()` emits.
- **`authorship-form-field`** (file/asset resource creator, has autocomplete): add a persistent `mat-hint`; remove a dead disabled-autocomplete-option that never actually rendered. **No** `addOnBlur` here — with an open suggestion list, blur can double-commit the typed value alongside the selected option.
- New i18n key `legal.dataSide.authorshipHint` in `en/de/fr/it` (domain terms per language: *Urheber:innen* / *auteur·ice* / *autore*).

## Challenges and Decisions

- Deliberately did **not** add COMMA as a chip separator — author names are often written "Last, First", which a comma separator would wrongly split into two chips.
- Residual micro-case: on the settings page the Save button is disabled until the form is dirty, and clicking a *disabled* button doesn't blur the input — so pure "type → click greyed Save" (with no other interaction) still needs Enter, which the now-visible hint instructs.

## Gotchas

- Romansh (`rm`) has no i18n file — it falls back to English at runtime (DEV-6629), so only `en/de/fr/it` were touched.

## Test Plan

- *Legal Settings → Resource side*: type an author, click into another field **without** pressing Enter → the name becomes a chip and **Save changes** enables; save + reload confirms persistence.
- *Resource Rights Statement*: edit authorship (✏️), type a name, click the save (💾) icon without pressing Enter → the pending name is included.
- Hint text visible under each authorship field (checked EN + FR).
- App compiles and serves cleanly against the dev API.

### DIFF
--- a/apps/dsp-app/src/assets/i18n/de.json
+++ b/apps/dsp-app/src/assets/i18n/de.json
@@ -5,6 +5,7 @@
       "license": "Lizenz",
       "copyrightHolder": "Rechteinhaber:in",
       "authorship": "Urheber:innen",
+      "authorshipHint": "Namen eingeben und mit Enter bestätigen, um Urheber:innen hinzuzufügen.",
       "edit": "Bearbeiten",
       "removeAuthor": "{{name}} entfernen",
       "uncategorized": "Die rechtlichen Angaben dieser Ressource wurden noch nicht festgelegt — bitte ergänzen.",
--- a/apps/dsp-app/src/assets/i18n/en.json
+++ b/apps/dsp-app/src/assets/i18n/en.json
@@ -5,6 +5,7 @@
       "license": "License",
       "copyrightHolder": "Copyright holder",
       "authorship": "Authorship",
+      "authorshipHint": "Type a name and press Enter to add each author.",
       "edit": "Edit",
       "removeAuthor": "Remove {{name}}",
       "uncategorized": "This resource's legal information hasn't been set yet — please add it.",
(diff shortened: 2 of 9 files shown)


derschnee68:
@claude review

derschnee68 (approved):

PR #3245, opened 2026-07-18
fix(legal): show the already-set license immediately in Legal Settings

danielasubotic:
Fixes https://linear.app/dasch/issue/DEV-6763

## Motivation

In **Settings → Legal Settings → Resource side**, when a project already has a license set, the **License** field renders blank for a moment when the tab opens and the label only appears after a short delay — it looks as though no license is set. Reported as DEV-6763 (same "renders before async data" family as DEV-6746).

## Summary

The selected license value is available immediately from the loaded project, but the `<mat-select>` options are loaded from a separate, uncached request. `mat-select` can only render a selected value's label once a matching `<mat-option>` exists, so the field stays blank for one round-trip. Seed the dropdown with the already-selected license so its option exists on first render; the fetched catalog then supersedes the seed.

## Key Changes

- `resource-side-legal-form.component.ts`: build a synchronous seed option for `project.dataLicense` and `startWith` it on `ccLicenses$`; extract a shared `_toCcLicenseOption` helper used by both the seed and the fetched list.
- Add `resource-side-legal-form.component.spec.ts`: regression coverage — with `getLicenses` mocked to `NEVER`, `ccLicenses$` must still emit the selected license synchronously; plus the fetched list superseding the seed and the no-license case.

## Challenges and Decisions

- The CC-BY label comes from the static `CC_BY_SUMMARY_KEYS` map, so the seed needs **no** network — no caching or spinner required.
- Chose seeding over a loading spinner or a root-level catalog cache: it is the smallest change that fixes the resting-state blank, adds no new DI (so no Storybook/spec fallout), and `track lic.iri` keeps the option stable when the seed is swapped for the full list (no flicker).

## Gotchas

- Not a `compareWith` issue — the option values and the form control are both string IRIs, so matching is correct; the label was merely late.

## Test Plan

- `nx test vre-pages-project-project` — 146/146 pass (incl. 3 new regression cases; the `NEVER` case fails if the `startWith` seed is removed)
- Built and served locally from this branch against the stage backend (compiles and runs). Reviewer visual check: open a project that has a resource-side license set → Settings → Legal Settings → Resource side; the License label now shows immediately instead of appearing after the `.../legal-info/licenses` request resolves.

### DIFF
--- a/libs/vre/pages/project/project/src/lib/project-settings/legal/resource-side-legal-form.component.spec.ts
+++ b/libs/vre/pages/project/project/src/lib/project-settings/legal/resource-side-legal-form.component.spec.ts
@@ -1,0 +1,94 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { LegalInfoApiService } from '@dasch-swiss/vre/3rd-party-services/api';
+import { ProjectDataRightsService } from '@dasch-swiss/vre/shared/app-helper-services';
+import { NotificationService } from '@dasch-swiss/vre/ui/notification';
+import { TranslateService } from '@ngx-translate/core';
+import { NEVER, of } from 'rxjs';
+import { ResourceSideLegalFormComponent } from './resource-side-legal-form.component';
+
+const CC_BY_40 = 'http://rdfh.ch/licenses/cc-by-4.0';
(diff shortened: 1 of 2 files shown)


derschnee68:
@claude review

derschnee68 (approved):

PR #3247, opened 2026-07-18
fix(project): render the Description settings submit button with the form

danielasubotic:
Fixes https://linear.app/dasch/issue/DEV-6746

## Motivation

On a project's **Settings → Description** (the tab it lands on), the **Submit** button rendered before the Description form appeared — for a moment the user saw a lone Submit button over empty content, which looked broken.

## Summary

`EditProjectFormPageComponent` gated the form on the async project fetch but left the Submit button ungated, so the button painted on the first change-detection pass while the form only appeared after the project GET resolved (and the child form finished building). Gate the button on the built `form` so it renders together with the fields.

## Key Changes

- `edit-project-form-page.component.ts`: wrap the Submit button in `@if (form)` so it appears with the form fields (both keyed on the built `FormGroup`).
- Add `edit-project-form-page.component.spec.ts`: asserts the submit button is absent before the project loads and before the form is built, and present once built.

## Challenges and Decisions

- Gated on `form` (the built `FormGroup`, set via the child's `afterFormInit`) rather than only `formData$`, because `form` is the exact readiness signal for the fields — this also closes the residual `allProjects$`/`_buildForm` sub-gap and removes a latent NPE (`onSubmit()` reads `this.form.value`, and the always-present button could be clicked before `form` existed).
- No behaviour change to the enabled/disabled state; matches the `@if (project$ | async)` gating already used in the sibling `LegalSettingsComponent`.

## Gotchas

- The misleadingly-named `description/` folder is the read-only view; the Settings edit tab is `EditProjectFormPageComponent` in `reusable-project-form/`.

## Test Plan

- `nx test vre-pages-project-project` — 146/146 pass (incl. 3 new cases; verified the spec fails if the `@if (form)` gate is removed)
- Served locally against stage and confirmed the Submit button no longer appears ahead of the form.

### DIFF
--- a/libs/vre/pages/project/project/src/lib/reusable-project-form/edit-project-form-page.component.spec.ts
+++ b/libs/vre/pages/project/project/src/lib/reusable-project-form/edit-project-form-page.component.spec.ts
@@ -1,0 +1,79 @@
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { FormGroup } from '@angular/forms';
+import { By } from '@angular/platform-browser';
+import { ProjectApiService } from '@dasch-swiss/vre/3rd-party-services/api';
+import { LocalizationService } from '@dasch-swiss/vre/shared/app-helper-services';
+import { NotificationService } from '@dasch-swiss/vre/ui/notification';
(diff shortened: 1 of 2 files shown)


derschnee68:
@claude review

derschnee68 (approved):

PR #3250, opened 2026-07-21
fix: show Incoming Links label and spinner while loading (DEV-6775)

danielasubotic:
Fixes DEV-6775

## Motivation

On the resource page, the loading spinner for **Incoming Links** was gone. Instead of hiding the whole section while the links load, the "Incoming Links" property should already be shown, with a spinner while the links are still loading.

## Summary

The loading (`@else`) branch of `IncomingLinksPropertyComponent` set `[isEmptyRow]="true"`. `PropertyRowComponent` hides a row when `showAllProperties` is off **and** `isEmptyRow` is true — and `showAllProperties` defaults to off (only filled properties are shown). So during loading the entire row (label **and** spinner) was hidden, and nothing appeared until the links resolved. This was a regression from #2660 (DEV-5448), which hardcoded `isEmptyRow="true"` on the loading branch; previously it was computed as `!loading && …`, i.e. `false` while loading.

## Key Changes

- `incoming-links-property.component.ts` (loading branch only):
  - `[isEmptyRow]="false"` so the row — and therefore the "Incoming Links" label — is shown immediately while loading.
  - Spinner placed on the left with top spacing via a shrink-to-fit wrapper (`display: inline-block; padding-top`), instead of the shared indicator's default horizontal centering.

## Challenges and Decisions

- The shared `AppProgressIndicatorComponent` centers itself (`margin: … auto`). `::ng-deep` is banned and its centered default is correct for its other (page-loader) usages, so rather than change it, the spinner is left-aligned locally by wrapping it in a shrink-to-fit container, which collapses the `auto` margins to the left.
- **Value alignment is intentionally out of scope.** The incoming-link values (and standoff-link values) sit ~12px further right than other property values because those rows omit `[singleRow]="false"`. That affects all value types and will be fixed uniformly in a separate issue, so this PR leaves the loaded-value rendering byte-for-byte identical to `main`.

## Gotchas

- `[singleRow]="false"` is kept **only** on the loading branch, where it governs the spinner container padding and shows no values — it does not touch value alignment.

## Test Plan

- Open a resource with incoming links (e.g. `https://app.stage.dasch.swiss/project/v0B3nQt2Q-6aZiFGRe88KQ/data/dasch/Author`) and reload; confirm the "Incoming Links" label appears immediately with a spinner on the left while links load (throttle the network to make it linger).
- Confirm the loaded incoming-link and standoff-link values are unchanged vs. `main`.

### DIFF
--- a/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/incoming-links-property.component.ts
+++ b/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/incoming-links-property.component.ts
@@ -38,8 +38,11 @@ import { sortByKeys } from './property-value/sortByKeys';
         [tooltip]="'resourceEditor.propertiesDisplay.incomingLinkTooltip' | translate"
         [label]="'resourceEditor.propertiesDisplay.incomingLinkLabel' | translate"
         [borderBottom]="true"
-        [isEmptyRow]="true">
-        <app-progress-indicator [compact]="true" />
+        [singleRow]="false"
+        [isEmptyRow]="false">
+        <div style="display: inline-block; padding-top: 12px">
+          <app-progress-indicator [compact]="true" />
+        </div>
       </app-property-row>
     }
   `,
--- a/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/incoming-links-property.component.ts
+++ b/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/incoming-links-property.component.ts
@@ -39,7 +39,8 @@ import { sortByKeys } from './property-value/sortByKeys';
         [label]="'resourceEditor.propertiesDisplay.incomingLinkLabel' | translate"
         [borderBottom]="true"
         [singleRow]="false"
-        [isEmptyRow]="false">
+        [isEmptyRow]="true"
+        [loading]="true">
         <div style="display: inline-block; padding-top: 12px">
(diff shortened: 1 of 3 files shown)


domsteinbach (approved):
It works. However I dislike that the dedicated input flag "isEmptyRow" is used in order to display a loading state. Different concepts are coupled. Display a loading state should depend on an actual loading state or on the existence or non existence of the  "@Input() prop: PropertyInfoValues | undefined;" in the child component. Those are preexisting design flaws of the child component.

danielasubotic:
Good point — I've decoupled the two concepts in a384bce.

`PropertyRowComponent` now has a dedicated `loading` input instead of leaning on `isEmptyRow`. The hide logic is `showAllProperties === false && isEmptyRow && !loading`, so:

- `isEmptyRow` keeps its real meaning ("the row has no values"),
- `loading` keeps the row visible while data is still coming in.

The incoming-links loading branch is now honest: `isEmptyRow=true` (no values yet) plus `loading=true`. Behaviour is identical for the loaded branch and the other two consumers, since `loading` defaults to `false`. Added a `LoadingRow` story to lock the visible-while-loading case in.

The other flaw you mentioned — `prop` being passed to the row but never read there, with `isEmptyRow` just restating `prop.values.length === 0` — I've left out of scope here so this PR stays a focused fix. Happy to file a separate ticket to clean up the child's API if you think it's worth it.

PR #3290, opened 2026-07-28
fix(advanced-search): translate list-value and resource-class chips (DEV-6857)

domsteinbach:
## Summary
- **List-value and resource-class chips now retranslate on language switch** — previously they were stuck showing either a language-frozen label ("Berufs-/Meisterprüfung" no matter which UI language was active) or the raw IRI ("Education Type equals http://rdfh.ch/lists…"). Both surfaces are fixed with one aligned change.
- The URL no longer persists a `valueLabel` for list values or resource-class `Matches` — those labels live in data the search page already fetches (list tree / ontology). Link-value chips still persist `valueLabel` because their labels have no multi-language source we already have on hand.
- On URL rehydration, node-shaped operands (list nodes, resource classes, linked resources) now decode as `IriLabelPair` — even without a `valueLabel` — so the model routes them to `NodeValue`, and the chip pipe resolves the current-language label from live data at render time.

## Why
Fixes [DEV-6857](https://linear.app/dasch/issue/DEV-6857/advanced-search-translate-selected-list-items-acc-to-chosen-language). The redesign in #3236 introduced a URL-as-source-of-truth model for the advanced-search state. On the wire the URL only carries one language of each label, and the rehydration path assigned bare-string operands to `selectedObjectValue`, which the setter routed to `StringValue` — the chip pipe then rendered the string verbatim (the IRI). This restores the translation behaviour promised by DEV-6645 for the new chip design.

## What changed
- `build-statements.ts` — new `_hasNodeShapedObject(fp, predicate)` dispatcher mirrors the model's `objectType` branching and produces an `IriLabelPair` operand for list values / resource-class Matches / link values. Plain literals still stay as strings.
- `advanced-search-bar.component.ts` — the URL writer skips `valueLabel` for list-value and resource-class-Matches chips. Kept for link values (no alternative label source).
- `service/list-node-label.resolver.ts` — new service, cached per list root, exposes a synchronous `getLabels(rootIri, nodeIri)` for the chip pipe and emits `changes$` after each load. Shares the fetch with the popover field via the same `DynamicFormsDataService.getListWithAllLanguages$` (one HTTP call per list root).
- `chip-label.pipe.ts` — impure pipe resolves list-value labels via `ListNodeLabelResolver`, resource-class labels via `OntologyDataService.resourceClasses$`. Falls back to stored labels → IRI when the source is still loading. Calls `ChangeDetectorRef.markForCheck()` on async updates so `filter-chip`'s OnPush view re-runs `chipLabel` when data arrives.
- Storybook: new list-value chip and resource-class Matches chip stories.
- Tests: 5 new specs in `build-statements.spec.ts`, 4 in `chip-label.pipe.spec.ts`, 3 in `advanced-search-bar.component.spec.ts`, plus a new `list-node-label.resolver.spec.ts` (7 tests). Total 227 tests green in the advanced-search lib.

## Backwards compatibility (not needed as of requirements)
- Old shared URLs containing a `valueLabel` for list / resource-class chips still decode — the resolver's multi-language labels take precedence, so the chip translates as soon as the tree loads.
- No URL contract changes required.

Follow up issue on potential double request after fresh reload by URL is created. 
https://linear.app/dasch/issue/DEV-6888/advanced-search-consolidate-duplicate-list-fetch-between-chip-resolver

🤖 Generated with [Claude Code](https://claude.com/claude-code); Edited by a human!!!


derschnee68:
@claude review

derschnee68 (approved):

PR #3291, opened 2026-07-28
fix(search): stop the spinner and show a persistent failure state when a search fails

mpro7:
[DEV-6866](https://linear.app/dasch/issue/DEV-6866)

## Summary

- A failed fulltext search left the spinner spinning **forever**: `loading` was only ever cleared on the success path inside `map()`, and the error snackbar auto-dismisses after 5s — so a user who looked away saw only a spinner and reasonably concluded results were still coming.
- Both search components now render a persistent, retryable failure state that is deliberately distinguishable from "no results found".
- A failing **count** query no longer discards results that arrived fine, and no longer fabricates a total. The count is the more expensive of the two queries (it re-runs the identical WHERE clause — see DEV-6809), so "results OK, count timed out" is the likely real-world case.

## Changes

**New shared UI** — `SearchFailedComponent` (`libs/vre/ui/ui`)
- `error_outline` icon + retry button, reusing `CenteredMessageComponent`; new `pages.search.searchFailed.{title,message}` in en/de/fr/it, retry label reuses the existing `ui.common.actions.retry`.

**Fulltext search** — `search-result.component.ts`
- `catchError` clears `loading`, raises a `failed` signal, and returns `of(null)` so the stream stays alive for retry. Retry re-runs the search on the current page.
- The error is still handed to the `ErrorHandler`, so the snackbar — which names the specific cause (timeout / no connection / no permission) that the generic panel cannot — keeps appearing. It is now a supplement, not the only signal.

**Advanced search** — `advanced-search-results.component.ts`
- Replaces `of([])` on error, which rendered the "no results found" empty state and told the user their search legitimately matched nothing when it had never completed. It also swallowed the error outright, so a failed advanced search previously produced *no* production feedback at all.

**Unknown result count** — `ResourceResultService`, `resources-list.component.ts`
- `numberOfResults` is now `number | null`, so the compiler forces every consumer to handle an unknown count. `resources-list` states **"Result count unavailable"** instead of asserting a total.
- This replaces an earlier revision of this PR that substituted the current page's length. That was wrong: on a 10,000-hit search returning a full page of 25, the UI asserted "25 results" and — because `25 > 25` is false — silently dropped the paginator, i.e. it failed precisely at the page-size boundary. Paging is still unavailable without a count, but the absence is now visible rather than disguised as a small result set.

**Docs** — `apps/dsp-app/src/assets/i18n/CLAUDE.md`
- It instructed running three `scripts/*.sh` translation scripts and reading `docs/TRANSLATION_REFACTORING.md`; none exist, so every "run the parity verification script" step was a dead end that silently skipped the check. Replaced with the inline check actually used here, plus the Storybook static-translations caveat.

## Test Plan

- **New** `search-result.component.spec.ts` (the component had no spec): spinner stops and `failed` is raised on error, the error still reaches the `ErrorHandler`, results survive a failed count with the count reported as `null`, and retry re-runs the search.
- **Confirmed the guard is load-bearing** — removing `loading.set(false)` from the new `catchError` fails it with `Expected: false / Received: true`.
- A results-query failure now also clears `numberOfResults`, so a 1,000-hit search followed by a failing page change no longer leaves the service reporting a stale total while in the failure state. Covered by a test reproducing exactly that sequence, verified to fail without the fix.
- Advanced-search spec extended with the same scenarios, including that a failure no longer emits `[]` — the regression that produced the misleading empty state.
- **Retry click path covered end to end.** `SearchFailedComponent` has its own story driving the click through the real output binding, and both container failure stories are now click-throughs (stub fails once, then succeeds; `beforeEach` resets the counter so the sequence no longer depends on render count). **Verified by deleting `(retry)="onRetry()"`: the app still builds and every unit test still passes, while the `SearchFails` story fails.**
- 5 new Storybook stories across 4 files, all passing in chromium via `test-storybook`.
- `nx affected` across **16 projects**: lint and test green. `dsp-app:build` clean. i18n key parity verified for de/fr/it.

## Count-failure observability — and why there is nothing to report to

A review round asked for the failed count query to be reported to telemetry rather than swallowed. Investigating that turned up a bigger finding, now filed as **[DEV-6872](https://linear.app/dasch/issue/DEV-6872)**:

| Channel | State | Evidence |
|---|---|---|
| Faro | `enabled: false` in **every** env, prod included | `config.prod.json` -> `instrumentation.faro.enabled` |
| Rollbar | `enabled: false` | `config.prod.json` |
| Sentry | **live** on prod/stage | `main.ts` excludes only `demo`/`dev`/`local-dev`/`ls-test`/`perf-01`/`test-rdu` |
| `SearchFlowLogger` | dev-only `console.warn` | every method early-returns on `!isDevMode()` |

`AppErrorHandler.handleError` reaches Sentry/Faro **only in its final `else` branch**, so an `ApiResponseError` — every dsp-api failure — has never reached telemetry at all. This PR therefore removed a snackbar, not a telemetry signal; there was none to remove.

What this PR does about it: advanced search logs the failed count through its dev-only `SearchFlowLogger` (free, helps local debugging), and both count handlers carry a comment pointing at DEV-6872. Deliberately **not** a snackbar — the results rendered fine, and an error toast over a working result list is noise. The systemic fix (a `report`-without-`notify` seam so every branch reaches Sentry) belongs in DEV-6872, not here.

## Out of scope

- Deferred: the issue's "consider a *still searching…* hint for 20s+ requests" — flagged as *consider*, not in the acceptance criteria, and needs its own threshold design.
- Deferred: an unknown-total paging mode in `PagerComponent` (it reads `lastPageIndex` in four places and would need a "was this page full" input) so forward paging could survive a failed count.
- **[DEV-6871](https://linear.app/dasch/issue/DEV-6871)** / #3292 — the same defect shape in `ResourcesListFetcherComponent` (project sidenav resource list), which had no `catchError` at all.


derschnee68:
@claude review

derschnee68:
@mpro7 I'm a bit concerned about the two major issues, what do you think?

domsteinbach:
I am also against fabricating a count. If there is an error, error. Not showing something which is often correct and also has side effects on the paginator itself. I'd go for error is error. No features while erroring. And one source of truth for the count.

mpro7:
@claude review

domsteinbach (approved):

PR #3298, opened 2026-07-29
fix(error-handler): report API errors to telemetry instead of only the non-HTTP branch

mpro7:
[DEV-6872](https://linear.app/dasch/issue/DEV-6872)

## Summary

- `AppErrorHandler.handleError` reported from its final `else` only — and to **neither** channel from the other three, since `_sendErrorToSentry` and `_sendErrorToFaro` were both confined to that branch. Every dsp-api failure (a triplestore timeout, a 500, a 403) was invisible in production once the snackbar auto-dismissed after five seconds.
- Reporting is split out into `ErrorReportingService`, which reports to telemetry **without** notifying the user, so a degraded path can record a failure it deliberately does not surface.
- The three count queries that absorb their own failure to keep a working result list on screen now report. That is the production signal DEV-6809 (count-query cost) and DEV-6864 (HTTP 500s) had none of.

Both transports are fed on every branch, so this is the prerequisite for **either** channel carrying API errors — turning Faro on without it would still deliver nothing, because in three of four branches there was no reporting call to deliver from.

## Why a plain `captureException` would not have worked

Neither `ApiResponseError` (it extends `ApiResponse`, a plain object) nor Angular's `HttpErrorResponse` extends `Error`. Captured raw, Sentry files all of them as one *"Non-Error exception captured with keys: error, method, status, url"* issue, and Faro's `pushError` rejects them outright. The service re-expresses them as a real `Error`.

Since there is no stack trace, Sentry groups by message — and dsp-api puts search terms, resource IRIs and the project IRI in the URL, so left alone that is one issue per request. Grouping is therefore pinned to `['dsp-api', method, status, route]`, where `route` keeps the first two path segments plus a `/count` marker.

## What lands on an event

| Field | Sentry | Faro |
|---|---|---|
| `status`, `method`, `route` | tags | context |
| full request URL | `extra` | context |
| server's reason (`knora-api:error` / `message` / `error`, truncated at 2000 chars) | `extra` | context |
| severity | `level` | `dsp.severity` field |
| grouping key | `fingerprint` | n/a — Faro groups by message and stack |

The tag/`extra` split is Sentry's constraint alone: it caps tag values at 200 characters and indexes them, so a unique Gravsearch URL belongs in `extra`. Faro's context has neither limit, so it receives everything — withholding `extra` there would leave Faro with strictly less than Sentry. The severity is a plain field for Faro because `pushError` has no level of its own.

The server's reason matters because the status alone is not actionable: it says a request was rejected, not *which constraint* rejected it. `AppErrorHandler` already reads those fields to build the snackbar and then drops them. `HttpErrorResponse.message` is deliberately **not** used as a fallback — it is composed from the status and URL, both already on the event — so a present reason means the server really did say something.

## Volume and severity

Three independent controls, because grouping, deduplication and alerting want different granularity:

- **Grouping** is coarse: `['dsp-api', method, status, route]`. The reason is *not* in the fingerprint, so one endpoint stays one Sentry issue.
- **Deduplication** is finer: the same fingerprint *and* the same reason is suppressed for 60s per tab. A user hammering retry through a triplestore timeout sends one event, not one per click — but two *different* rejections of one endpoint are two different failures and both get through, because keying on the fingerprint alone would silently discard the second reason. The trade is deliberate: a burst of distinct 4xx on one route is no longer collapsed to a single event.
- **Severity** carries the rest: a 5xx is captured at `error`, everything else at `warning`. Status 0 counts as a warning — the app already maps it to the "no internet" message, reading it as the user's own connectivity rather than a fault of ours.

JS errors are unchanged throughout: passed to Sentry as-is, no fingerprint, no level (so Sentry's default `error` applies), no severity field, and never deduplicated.

**Decision worth noting:** 400s are reported, including the ones that are really rejected user input. The alternative was filtering to 5xx/network only, which would also have dropped the frontend/API contract bugs that show up as 400s. The severity split is what keeps that affordable — an invalid date does not fire the same alerts as a triplestore timeout.

**Widening worth noting:** response bodies can echo submitted values, so slightly more user data reaches telemetry. It is incremental rather than a new class — full URLs already go, carrying search terms and resource IRIs.

## Follow-up: enabling Faro

The issue asked whether Faro should be switched on or removed as dead configuration. **Decided: it stays, as the intended source of truth.** Enabling it is out of scope here because it is not a config flip — `instrumentation.faro.collectorUrl` is `http://localhost:12345/collect` in **all five** environments including `prod`, and `tracingCorsUrls` is empty everywhere except `dev-server`. It needs real collector endpoints provisioned per environment plus a ~300KB lazy-loaded bundle turned on in production. This PR makes sure that, once enabled, Faro actually receives API errors.

The 60s dedup window is also tuned to Sentry's per-event cost. If Faro becomes primary, suppressing less (or suppressing per transport) is worth revisiting — a knob, not a defect.

## Changes

**`libs/vre/core/error-handler`**
- New `ErrorReportingService` with `report(error, context?)` — normalization, fingerprinting, reason extraction, per-reason dedup, severity grading, both transports at parity.
- `AppErrorHandler` reports before branching and delegates its two private `_sendErrorTo*` methods. Snackbar messages and the non-prod `console.error` are unchanged.

**Call sites** — `SearchResultComponent`, `AdvancedSearchResultsComponent`, `ResourcesListFetcherComponent` report their swallowed count-query failure with component/operation context. The comments claiming no production channel exists are corrected.

**`types/window.d.ts`** — `captureException` accepts the `captureContext` argument.

## Test Plan

- `nx run vre-core-error-handler:test` — 34 pass. Covers a report call on each of the four `handleError` branches (the fix-confirmation tests: all three snackbar branches reported nothing before), plus normalization against the real dsp-api URL shapes, reason extraction from both error shapes and its absence when the body carries none, truncation, the reason being excluded from the fingerprint but included in the dedup key, the dedup window and its expiry, severity grading per status class, Faro receiving every field Sentry gets, context tags, the absent-SDK no-op, and one failing transport not stopping the other.
- `nx run-many --target=test --projects=vre-pages-search-search,vre-pages-search-advanced-search,vre-pages-project-project` — pass, each with a new test asserting the count failure is reported *and* still shows no snackbar.
- `nx run-many --target=lint` on all four projects — clean.
- `nx run dsp-app:build` — succeeds, covering the global `window.d.ts` change.

Sentry contract verified against the installed `@sentry/core` v10 rather than from memory: `{ tags, extra, fingerprint, level }` is discriminated as a `CaptureContext` (all four keys are in `captureContextKeys`), and `applyFingerprintToEvent` concatenates onto an otherwise-empty `event.fingerprint`, so the array is the whole grouping key. Nothing else in dsp-app calls `setFingerprint`.

## Screenshots

No UI change — all three call sites are deliberately invisible to the user.

### DIFF
--- a/libs/vre/core/error-handler/src/index.ts
+++ b/libs/vre/core/error-handler/src/index.ts
@@ -1,4 +1,5 @@
 export * from './lib/app-error-handler';
 export * from './lib/app-error';
+export * from './lib/error-reporting.service';
 export * from './lib/user-feedback-error';
 export * from './lib/js-lib-parsed-error';
--- a/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
+++ b/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
@@ -1,3 +1,4 @@
+import { HttpErrorResponse } from '@angular/common/http';
 import { TestBed } from '@angular/core/testing';
 import { ApiResponseError } from '@dasch-swiss/dsp-js';
 import { AppConfigService } from '@dasch-swiss/vre/core/config';
@@ -5,6 +6,8 @@ import { NotificationService } from '@dasch-swiss/vre/ui/notification';
 import { TranslateService } from '@ngx-translate/core';
 import { AjaxError } from 'rxjs/ajax';
 import { AppErrorHandler } from './app-error-handler';
+import { ErrorReportingService } from './error-reporting.service';
+import { UserFeedbackError } from './user-feedback-error';
 
 /**
  * Builds an `ApiResponseError` wrapping an `AjaxError` with the given status and JSON response body —
@@ -32,9 +35,11 @@ describe('AppErrorHandler', () => {
   let handler: AppErrorHandler;
   let openSnackBar: jest.Mock;
   let instant: jest.Mock;
+  let report: jest.Mock;
 
   beforeEach(() => {
     openSnackBar = jest.fn();
+    report = jest.fn();
     // Echo the key back so assertions can distinguish "showed a translated fallback" from "showed the API reason".
     instant = jest.fn((key: string) => key);
 
@@ -44,6 +49,7 @@ describe('AppErrorHandler', () => {
         { provide: NotificationService, useValue: { openSnackBar } },
         { provide: TranslateService, useValue: { instant } },
         { provide: AppConfigService, useValue: { dspInstrumentationConfig: { environment: 'dev' } } },
+        { provide: ErrorReportingService, useValue: { report } },
         // NgZone is the real one from the zone test env; its run() executes synchronously, so the
         // snackbar call inside displayNotification is observable in the assertions.
       ],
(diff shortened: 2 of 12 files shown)


mpro7:
@claude review

domsteinbach (approved):

PR #3299, opened 2026-07-29
fix(i18n): pick data-side labels via a fixed fallback chain (DEV-6875)

domsteinbach:
## Summary

- `pickPreferredLanguageString` now walks a deterministic `en > de > fr > it > rm` fallback chain when the preferred UI language is missing (or its value is empty/whitespace-only), instead of returning "first non-empty in whatever order the API returned".
- The chain is independent of the current UI language, so the same data resolves to the same label across API endpoints and across users. Displayed labels and sort keys (`SortingHelper.sortByLocalizedString`, `StringifyStringLiteralPipe`) stay in sync automatically.
- Extended specs to cover the fallback order, order-independence across input arrays, and whitespace-only skipping mid-chain.

## Why

Reported in DEV-6875: with `rm` selected as UI language a single project surfaced labels in three different languages depending on the component:
- `ResourceClassSidenavComponent` → Italian (v3 endpoint returned `it, fr, en, de`)
- `DataClassPanelComponent` → German (v2 endpoint returned `de, en, fr, it`)
- `ResourceHeaderComponent` → German

Root cause was the picker's `value.find(l => l.value && ...)` — literally "first in array". Fixing it in one central function repairs every consumer of the pipe *and* the sorter.

## Behavior change to be aware of

- Any list sorted via `sortByLocalizedString` may reorder for users whose current language isn't present in the data — that's the intended win (stable ordering across users), but reviewers/QA should expect visible diffs in the ontology sidenav, list-page, resource-class chips, property pickers, etc.
- Anywhere the code still reads `labels[0]` / `comments[0]` directly bypasses this picker and will keep showing whatever the API returned first — worth a follow-up sweep but out of scope for this PR.

## Test plan

- [ ] `nx run app-helper-services:test` — 37/37 pass locally
- [ ] `nx run vre-library-app-string-literal:test` — 14/14 pass locally
- [ ] Manual: switch UI to `rm` on stage, open a project — confirm the resource class labels in sidenav, panel, and header all show the same string (English) for the same resource class.
- [ ] Manual: confirm ontology list sorting under `rm` UI matches what a `de` user sees for data with no `rm` label.

### DIFF
--- a/libs/vre/shared/app-helper-services/src/lib/pick-preferred-language-string.spec.ts
+++ b/libs/vre/shared/app-helper-services/src/lib/pick-preferred-language-string.spec.ts
@@ -9,15 +9,15 @@ describe('pickPreferredLanguageString', () => {
     expect(pickPreferredLanguageString(labels, 'en')).toBe('Documentation');
   });
 
-  it('returns the first non-empty value when preferred language is missing', () => {
+  it('falls back through the fixed chain (en first) when preferred language is missing', () => {
     const labels = [
       { language: 'de', value: 'Dokumentation' },
       { language: 'en', value: 'Documentation' },
     ];
-    expect(pickPreferredLanguageString(labels, 'fr')).toBe('Dokumentation');
+    expect(pickPreferredLanguageString(labels, 'fr')).toBe('Documentation');
   });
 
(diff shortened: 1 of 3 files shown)


derschnee68:
@claude review

derschnee68 (approved):

PR #3300, opened 2026-07-29
chore(ci): sharpen the Claude review prompt against speculative findings

mpro7:
Prompt and tool-permission changes to the `claude-code-review` job. No action version bump — `@v1` already dereferences to `v1.0.183` (commit `be7b93b`, 2026-07-25), the newest release; the floating major tracks it, and no v2 exists.

## Why

Two properties of the old prompt worked against useful review.

**It forbade the reads that refute a false flag.** *"Focus your review only on files and lines changed in this PR. Do not review unrelated parts of the repository."* Claude reasonably reads that as "do not open other files" — and those are precisely the reads needed to turn a suspicion into a verdict. Checking whether a symbol has any remaining call sites, or whether a root-provided service resolves without explicit test providers, means leaving the diff. Scope of *findings* and scope of *reading* are different things, and the prompt conflated them.

**It asked for a graded list rather than a filtered one.** *"Grade each issue as Critical, Major, or Minor"* is a taxonomy with no floor: anything noticed can be filed as Minor rather than dropped, which is how review output fills with "consider extracting this" and "this could be more robust".

## What changed

| | Before | After |
|---|---|---|
| Reading | "do not review unrelated parts of the repository" | read callers, tests, neighbouring implementations and git history freely; report only on changed lines |
| Diff context | implicit, `gh pr diff` default is 3 lines | read changed files in full |
| Finding bar | grade everything Critical/Major/Minor | name a triggering input and a wrong outcome |
| Verification | none | a **separate agent per Critical/Major finding**, given the finding and the code but not the reasoning behind it, told to argue it is wrong and default to "refuted" when ambiguous. If subagents are unavailable the job errors out rather than posting an unverified review |
| Weak findings | downgraded to Minor | **deleted**; "no findings" stated to be a valid outcome |
| Dependency claims | asserted from memory | verify against the installed version or the library's docs, or delete the finding |
| Test coverage | "test coverage (only for modified code)" | name a test that would fail without the fix, or say none exists |
| Guidance file | `CLAUDE.md` | `REVIEW.md`, if present |
| Tools | read-only `gh` commands | plus `Grep`, `Glob`, `WebSearch`, `WebFetch`, `Task`, `gh api` |

The `CLAUDE.md` swap matters because this repo's `CLAUDE.md` is interactive operating rules — permission gates, planning ceremony, "ask before editing." In CI there is nobody to ask, so most of it is noise to a reviewer and some of it is actively misleading.

Reading files was **already** in the action's base tool set ([configuration.md](https://github.com/anthropics/claude-code-action/blob/main/docs/configuration.md): *"The base GitHub tools are always included. Use `--allowedTools` to add additional tools"*), so the capability was there and the old prompt was talking Claude out of using it. Search, web lookup and `Task` genuinely were not available — those are the real additions.

## Review feedback applied

Both of @BalduinLandolt's comments, in `a635cd964`:

- **Verification moved out of the reviewing agent.** Self-refutation is compromised by the reasoning that produced the finding, so it now goes to a separate agent per finding. Caveat: a subagent still receives a prompt written by the parent, so the framing leak is reduced rather than eliminated — hence the explicit instruction to withhold the reasoning.
- **Tools granted.** Concrete case from the sibling PR #3298: its two load-bearing claims were Sentry API contracts — whether `{tags, extra, fingerprint}` parses as a `CaptureContext`, and the 200-character tag-value cap. Neither is answerable from a diff. A reviewer without search asserts that class of claim from memory, which is exactly the false-flag category this PR exists to kill.

- **Failure is loud, not silent** (second round of feedback). The original fallback had Claude note the problem and check findings itself, which degrades to the exact anchored self-check the subagent pass removes, while still producing a review that reads as verified. Losing subagents entirely is now a hard stop: a short error, no review. An individual spawn failing is separate — retry once, then keep that finding labelled `unverified`, since one flaky spawn should not sink a whole review.

`Task` is **undocumented** for `claude-code-action` — `Read`, `Edit`, `Write`, `Bash` and `WebSearch` appear in the docs' `--allowedTools` examples, subagents nowhere. The CLI has the tool and `claude_args` passes through, so it likely works. **If it does not, the first `@claude review` will return an error instead of a review** — that is deliberate, and the fastest way to turn the unknown into an answer.

## Timing

`timeout-minutes` stays at **15**. Measured across the 128 successful runs since the last prompt change (2025-11-26): 2 min median, 6 min at p90, 14 min max. Ample in the normal case. Because the new prompt does more work per run and that 14-minute run existed, the prompt now tells Claude to verify in descending order of severity and report the remainder marked unverified rather than letting the job time out — a partial review is useful, a timed-out one is not.

## What this does not fix

The reviewer still cannot run tests, lint, or the build — no `Bash` beyond the allow-listed `gh` commands — so any claim about runtime behaviour stays unverified. Granting `Bash(npx nx *)` would let a comment-triggered job execute repo code, which is not a trade worth making on a public repo.

Adding `WebSearch`/`WebFetch` also widens the prompt-injection surface, since a diff or comment could induce a fetch to an attacker-chosen URL. `Bash` stays confined to read-only `gh` patterns and the job cannot push, so the realistic worst case is a wasted run rather than a leak.

## Follow-ups

- **DEV-6884** — the job has no author gating: any GitHub account that can comment on a PR in this public repo can trigger it and spend the org's API key. Out of scope here; tracked separately as security hardening.
- **DEV-6879** — a `REVIEW.md` for this repo, the larger remaining win and independent of this change: the prompt governs review *discipline*, `REVIEW.md` encodes what *this* repo cares about. Also read by the local `/eng:reviewing` workflow, so one file sharpens both paths.

## Test Plan

- `.github/workflows/claude.yml` parses (`js-yaml`); resolved config verified — 13 allowed tools, 72-line prompt, both jobs at `timeout-minutes: 15`.
- No behavioural change to the `claude-general` job, the triggers, the concurrency group, or the failure notification.
- Verification is the next `@claude review` on a real PR: fewer findings, each naming a concrete failure, and either evidence that subagents ran or the explicit statement that they were unavailable.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -42,28 +42,47 @@ jobs:
             REPO: ${{ github.repository }}
             PR NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
 
-            Please review this pull request and provide feedback on:
-            - Code quality and best practices
-            - Potential bugs or issues
-            - Performance considerations
-            - Security concerns
-            - Test coverage (only for modified code)
+            Review this pull request for correctness, security, performance, and test coverage.
 
-            Focus your review **only on files and lines changed in this PR**.
-            Do **not** review unrelated parts of the repository.
-            Do **not** make any code changes, run commands, or push code; only provide analysis and suggestions.
+            ## Scope
 
-            Structure your review as follows:
-            1. Summary of PR changes
-            2. Identified issues (prioritize Critical issues, including bugs, security vulnerabilities, or regressions):
-              - Grade each issue as Critical, Major, or Minor.
-              - Only Critical issues should prevent the PR from being ready to merge.
-              - Minor or style issues should be noted but do not affect merge readiness.
-            3. Recommendations for improvement (aligned with the issue severity)
-            4. Clear statement whether the PR is ready to merge (based **only** on Critical issues)
+            Report issues **only in lines this PR changed**. But READ WHATEVER YOU NEED to decide
+            whether an issue is real: open the changed files in full (the diff hunks alone give too
+            little surrounding context), plus their callers, their tests, neighbouring
+            implementations of the same pattern, and read-only git history. Reading widely is
+            required; reporting widely is not.
 
-            Be constructive, precise, and concise. Avoid verbose explanations unless necessary for clarity.
-            Use the repository's CLAUDE.md for guidance on style, conventions, and repository-specific best practices.
+            ## Every finding must survive this test
+
+            Name the concrete input or state that triggers it, and the wrong output, crash, or
+            regression that results. Then try to refute your own finding by checking the source.
+
+            DELETE — do not downgrade — any finding where:
+            - you cannot name a specific input and a specific wrong outcome;
+            - the justification is "more robust", "best practice", or "could break if X" where X
+              does not occur in this codebase;
+            - the code is already correct under TypeScript, RxJS, or Angular semantics;
+            - it is a deliberate decision explained in a comment (read the comment before flagging).
+
+            Three defensible findings beat twelve speculative ones. Reporting no findings is a
+            valid and useful outcome.
+
+            ## Test coverage
+
+            For a bug fix, check that a test exists which would FAIL without the change — name it,
+            or state that none exists. Do not ask for tests that only restate an implementation
+            line or assert a config value.
+
+            ## Output
+
+            1. One-paragraph summary of what the PR does.
+            2. Findings, most severe first. For each: `file:line`, the failure scenario, and the
+               fix. Grade Critical / Major / Minor, where Critical means a bug, security hole, or
+               regression that reaches users.
+            3. Ready to merge: yes or no, based **only** on Critical findings.
+
+            Be precise and concise. Read REVIEW.md at the repository root if it exists; its rules
+            override these defaults. Do **not** change code, push, or run commands.
             Use `gh pr comment` to leave your review as a comment on the PR.
 
           claude_args: '--allowed-tools "Bash(gh issue view:*),Bash(gh search:*),Bash(gh issue list:*),Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr list:*)"'
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -75,9 +75,14 @@ jobs:
             the claim against the source, and to default to "refuted" where the evidence is
             ambiguous. Drop every finding its verifier refutes.
 
-            If spawning agents is unavailable to you, say so explicitly in your output and run the
-            same pass yourself, re-deriving each finding from the code without reference to your
-            earlier reasoning.
+            If you cannot spawn agents at all, do NOT check the findings yourself and do NOT post a
+            review. Post only a short error stating that subagents were unavailable and what
+            happened when you tried. A review that reads as verified but was not is worse than no
+            review, and quietly falling back to self-checking restores the exact anchoring this
+            section exists to remove.
+
+            If an individual spawn fails, retry it once; if it fails again keep that finding but
+            label it `unverified` and carry on. One flaky spawn should not sink a whole review.
 
             Apply the deletion test to Minor findings without a verifier.
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -55,7 +55,7 @@ jobs:
             ## Every finding must survive this test
 
             Name the concrete input or state that triggers it, and the wrong output, crash, or
-            regression that results. Then try to refute your own finding by checking the source.
+            regression that results.
 
             DELETE — do not downgrade — any finding where:
             - you cannot name a specific input and a specific wrong outcome;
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -42,28 +42,47 @@ jobs:
...
+
+            DELETE — do not downgrade — any finding where:
+            - you cannot name a specific input and a specific wrong outcome;
+            - the justification is "more robust", "best practice", or "could break if X" where X
+              does not occur in this codebase;
+            - the code is already correct under TypeScript, RxJS, or Angular semantics;
+            - it is a deliberate decision explained in a comment (read the comment before flagging).
+
+            Three defensible findings beat twelve speculative ones. Reporting no findings is a
+            valid and useful outcome.
```
BalduinLandolt, line 68: I would run this in a subagent rather than just in the agent itself, so that the decision does not get influenced by the reasoning that led to the warning in the first place.

Maybe something like  
"For every finding, start an agent doing an adversarial cross-check, to try to refute it as a false-negative."  
I'm sure Claude can come up with a better phrasing, but you get my idea.

.github/workflows/claude.yml
```diff
@@ -42,28 +42,47 @@ jobs:
...
+            2. Findings, most severe first. For each: `file:line`, the failure scenario, and the
+               fix. Grade Critical / Major / Minor, where Critical means a bug, security hole, or
+               regression that reaches users.
+            3. Ready to merge: yes or no, based **only** on Critical findings.
+
+            Be precise and concise. Read REVIEW.md at the repository root if it exists; its rules
+            override these defaults. Do **not** change code, push, or run commands.
             Use `gh pr comment` to leave your review as a comment on the PR.
 
           claude_args: '--allowed-tools "Bash(gh issue view:*),Bash(gh search:*),Bash(gh issue list:*),Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr list:*)"'
```
BalduinLandolt: I could imagine it to yield better results, if it has more tools at its disposal (e.g. grep for repo searches, or web search).

BalduinLandolt (approved):
Certainly an improvement, but I'm sure we can do even better. See my comments for two thoughts

.github/workflows/claude.yml
mpro7: Applied in dc403ced8 — you're right, and the self-check was the weak part of this.

One caveat worth naming: a subagent spawned by the reviewer still receives a prompt *written by* the reviewer, so the framing leak is reduced rather than eliminated. To get as much of the benefit as possible the prompt now says to hand the verifier the finding and the surrounding code but explicitly **not** the reasoning that produced it, and to default to "refuted" where the evidence is ambiguous.

Two constraints I hit while implementing it:

**`Task` is undocumented for `claude-code-action`.** `Read`, `Edit`, `Write`, `Bash` and `WebSearch` all appear in the docs' `--allowedTools` examples; subagents appear nowhere. The underlying CLI has the tool and `claude_args` passes straight through, so it very likely works — but instructing "spawn an agent" when the tool is absent would get us narrated fake verification, which is worse than an honest inline check. So the prompt carries a fallback clause: if spawning agents is unavailable, say so in the output and re-derive each finding from the code without reference to the earlier reasoning. The first real run answers the question for us.

**The 15-minute timeout.** Per-finding verifiers multiply wall-clock, and a timeout posts "⚠️ Claude Code Review failed" instead of a review — a worse result than a mildly anchored self-check. Raised the review job to 25 minutes and scoped verification to Critical and Major findings; Minor ones just get the deletion test. Left `claude-general` at 15.

.github/workflows/claude.yml
mpro7: Applied in dc403ced8. Tool list is now:

```
Grep, Glob, WebSearch, WebFetch, Task,
Bash(gh issue view|search|issue list|pr comment|pr diff|pr view|pr list|api)
```

Reading files turned out to already be in the action's base set — per [configuration.md](https://github.com/anthropics/claude-code-action/blob/main/docs/configuration.md), *"the base GitHub tools are always included. Use `--allowedTools` to **add** additional tools"*, and the base set covers file operations and read-only git. So the capability was there and the old prompt was talking Claude out of using it (*"do not review unrelated parts of the repository"*, which reads as "don't open other files"). Search and web lookup genuinely weren't available, so those are the real additions.

Concrete case for web search, from the PR this one grew out of: on #3298 the two load-bearing claims were Sentry API contracts — whether `{tags, extra, fingerprint}` is parsed as a `CaptureContext` rather than an `EventHint`, and whether tag values really cap at 200 characters. Neither is answerable from the diff. I settled them by reading the installed `@sentry/core` and fetching Sentry's docs; a reviewer without those tools asserts that class of claim from memory, which is exactly the false-flag category this PR is trying to kill. There's now a prompt section saying not to assert a dependency's API from memory — verify against the installed version or the library docs, or delete the finding.

Also added `Bash(gh api:*)` so it can confirm a referenced issue or PR actually exists.

One downside worth stating rather than sliding past: web access on a job triggered by any PR comment widens the prompt-injection surface, since a diff or comment could induce a fetch to an attacker-chosen URL. `Bash` stays confined to read-only `gh` patterns and the job can't push, so the realistic worst case is a wasted run rather than a leak — low, but not zero. Happy to drop `WebFetch` and keep only `WebSearch` if you'd rather narrow it.

.github/workflows/claude.yml
BalduinLandolt: makes sense. Maybe we could add a clause "if you fail to spawn a subagent, do not conclude the review, but instead return with a simple error message indicating what happened"

.github/workflows/claude.yml
BalduinLandolt: I think that should be ok, as long as the reviewer doesn't have permission to write to the repo, the attack surface is really small. (nothing secret in this repo, which could leak; a prompt injection that makes a bad PR pass review is a small risk)

PR #3306, opened 2026-08-03
fix(advanced-search): anchor ?mainRes for class-less non-fulltext searches (DEV-6894)

domsteinbach:
## Problem

In Advanced Search, a search with **no resource class** ("All resources") filtering only on **Resource label** — *is like* / *equals* / *does not equal* — fails with **HTTP 400**:

```
Types could not be determined for one or more entities: ?mainRes
```

The generated Gravsearch has nothing typing `?mainRes`:

```sparql
WHERE {
?mainRes rdfs:label ?label .
?mainRes rdfs:label ?res0 .
FILTER regex(?res0, "Einstieg", "i") .
}
```

dsp-api's type inspection (`SearchResponderV2` → `GravsearchTypeInspectionRunner`, over the WHERE clause) requires a type for every entity, including `?mainRes`. A class restriction or a `matchFulltext` term types it — **`rdfs:label` does not**.

Introduced by the advanced-search refactor **#3236**, which removed the generic `?mainRes a knora-api:Resource .` anchor. That removal is a genuine perf optimization for the fulltext path (the anchor defeats matchFulltext index anchoring), but it broke the class-less label-only case.

## Fix

Re-introduce `?mainRes a knora-api:Resource .` **only when no class is selected AND no fulltext term is present** — the exact gap where nothing else types `?mainRes`. The per-class `UNION` is **not** reintroduced.

`_restrictToResourceClassStatement`:
- class selected → `?mainRes a <class> .`
- no class + fulltext → `''` (matchFulltext types `?mainRes`)
- no class + no fulltext → `?mainRes a knora-api:Resource .`

## Verification (live, api.dev — BIZ project 0828)

| Scenario | Before | After |
|---|---|---|
| class-less label _is like_ "Einstieg" | **400** | **200, 2 results** |
| class-less label _equals_ | **400** | **200, 1 result** |
| class-less **fulltext** | 200 (anchor-free) | 200 (still anchor-free — perf preserved) |
| class selected | 200 | 200 (class is the anchor) |

## Tests

- Corrected 2 pre-existing tests that encoded the buggy anchor-less output.
- Added 5 tests covering the three-way anchor branch.
- Full advanced-search lib suite green (218 tests), lint clean.

## Scope

Fixes the HTTP-400 class of bugs from the DEV-6889 analysis. The multi-valued "does not equal" leak (DEV-6889 proper) and the duplicate `rdfs:label` binding are tracked separately.

Closes DEV-6894.

### DIFF
--- a/libs/vre/pages/search/advanced-search/src/lib/service/gravsearch.service.ts
+++ b/libs/vre/pages/search/advanced-search/src/lib/service/gravsearch.service.ts
@@ -56,13 +56,17 @@ export class GravsearchService {
       '?mainRes knora-api:isMainResource true .\n' +
       `${constructStatements}\n` +
       '} WHERE {\n' +
-      // NB: no generic `?mainRes a knora-api:Resource .` anchor. Measured against the dev DB it is the
-      // dominant cost — it defeats matchFulltext's index anchoring and forces a full project-wide
-      // resource scan (60-80s for some terms). `?mainRes` is always typed by something else: the class
-      // restriction, matchFulltext (its first arg is resource-typed), or a property statement (its
-      // subject's domain). The one shape with none of those (no class, no fulltext, no filter) is not
-      // generated (gravsearchQuery$ returns null). Verified: parity with /v2/search and 14-20x faster.
-      `${this._restrictToResourceClassStatement(resourceClassIri)}\n` +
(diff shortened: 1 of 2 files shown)


BalduinLandolt (approved):
looks reasonable to me

PR #3309, opened 2026-08-03
fix(advanced-search): negate over whole property for value/list "does not equal" (DEV-6889)

domsteinbach:
Two related fixes to the Advanced Search Gravsearch generator, in separate commits.

## Commit 1 — negate over the whole property for value/list "does not equal"

**Problem:** A filter like **Genre does not equal "Mural"** still returned resources that have "Mural" as *one of several* values. On a multi-valued value/list property, the generated filter was satisfied by the resource's *other* value:
- value types emitted `FILTER (?resNval != X)` — another value satisfies `!=`;
- list values emitted `FILTER NOT EXISTS { ?resN <listNode> X }`, but `?resN` was bound by the shared *outer* object projection, so another list value satisfied the negation.

**Fix:** for these two NotEquals cases, move the object projection **inside** the `FILTER NOT EXISTS` block (value types additionally use the positive `=` there), so the negation covers the **whole property**: "there is no value of this property equal to X".
- `whereStatement`: suppress the shared outer `objectProjection` **only** for value/list NotEquals.
- `_whereStatementForValueComparison` / `_getWhereStatementForListObjectComparison`: emit the projection inside their NOT EXISTS block.
- `constructStatement`: don't project the now-inner variable for these cases.
- **Untouched:** label NotEquals (keeps `!=` — labels are single-valued), link NotEquals (already correct via its own self-contained NOT EXISTS), and every non-NotEquals operator.

## Commit 2 — drop the duplicate `rdfs:label` binding in label filters

**Problem:** A "Resource label" filter emitted a second `?mainRes rdfs:label ?resN` projection on top of the `?mainRes rdfs:label ?label` the query assembly already binds for ordering, producing a redundant duplicate triple in every label search:
```sparql
?mainRes rdfs:label ?label .
?mainRes rdfs:label ?res0 .      # duplicate
FILTER regex(?res0, "Einstieg", "i") .
```

**Fix:** filter on the assembly's `?label` variable directly (extracted as the shared `LABEL_VARIABLE` constant) and skip the writer's own projection for the ResourceLabel type. Label behaviour otherwise unchanged (equals `=`, notEquals `!=`, matches `matchLabel`, isLike `regex`). Result:
```sparql
?mainRes rdfs:label ?label .
FILTER regex(?label, "Einstieg", "i") .
```

## Verification (live, api.dev — BIZ project 0828)

Captured the app's real emitted queries; both the **result** and **count** requests are accepted (HTTP 200) with the corrected structure:

- **List NotEquals** (`Education Area does not equal Vorschule`): 187 → **164** results; every resource whose Education Area is Pre-school is excluded (even multi-valued ones).
- **Text NotEquals** (`Description does not equal X`): emits `FILTER NOT EXISTS { ?mainRes <pred> ?resN . ?resN valueAsString ?resNval . FILTER(?resNval = X) }` — accepted.
- **Label is-like** (`Einstieg`): single `rdfs:label` triple, filters on `?label`, accepted (200), 2 results.
- **Count query**: byte-identical body to the search query → same fixes apply; pagination works.
- **Link / Label NotEquals**: unchanged, no regression.

Unit tests updated for the changed shapes (value/list NotEquals now assert the NOT EXISTS form; label tests assert `?label`). Full advanced-search lib suite green (218 tests), lint clean.

## Related

Builds on #3306 (DEV-6894, the class-less `?mainRes` type anchor). Broader analysis of the advanced-search refactor regressions is on the parent issue.

Closes DEV-6889.


BalduinLandolt (approved):
from what I can tell, looks good 👍

PR #3315, opened 2026-08-05
fix(error-handler): stop AppErrorHandler from throwing on the paths that still could (DEV-6872)

mpro7:
Follow-up to #3314, from the multi-perspective review of that PR. #3314 fixed the wiring defect that made `handleError` throw; this closes the paths that can still make it throw, and corrects a comment that claimed more than the change delivered.

A throwing `ErrorHandler` is expensive out of proportion to its cause: it costs the user the snackbar for every error, and — because the search and sidenav components call the handler inside `catchError` — it errors their stream, so the retry button goes dead.

## 1. A failing telemetry path no longer swallows the snackbar

`handleError` reports before it branches, unguarded, so an exception from `ErrorReportingService.report()` took every error message in the app with it. That is exactly how one undefined dependency became an app-wide snackbar outage on stage.

Telemetry is the auxiliary half of the method; the snackbar is the half the user depends on. A broken reporting path now costs telemetry and nothing else.

## 2. A 409 no longer throws — and now says what went wrong

```ts
} else if (error.status === 409) {
  message = (error as AjaxError).response['knora-api:error'];   // no `response` on HttpErrorResponse
```

Only the JS-LIB `AjaxError` carries `response`. A 409 from the generated OpenAPI client is an `HttpErrorResponse`, which keeps its body on `error`, so this read threw inside the handler. The vendored spec declares a 409 for 171 operations — reachable from ordinary use, not a corner case.

Every one of those 409s answers with a `{ message }` `ConflictException`, so reading only `knora-api:error` would have stopped the throw while dropping the server's explanation — a duplicate project shortcode telling the user to contact support. The branch now reads the body through a reader shared with `ErrorReportingService`, which already knew all four shapes dsp-api answers with (`knora-api:error` → `message` → `error` → bare string), with the contact-support message as the fallback.

## 3. A 400 with an empty body no longer throws

The `ApiServices` 400 branch fell through to `error.error.message` past two guards that let `null` through, and Angular leaves `error` null for a 400 with no body — how a gateway or proxy answers. Nothing to show the user in that case, but not by way of an exception.

## 4. Comment corrected

The ordering comment added in #3314 claimed a throw in the global handler could not take the retry panel down. Half true: the failure state is committed first, so the eternal spinner cannot return, but a throw from `catchError`'s selector still propagates as an error notification, `switchMap` forwards it outward, and retry goes dead. The comment now states what the ordering buys and points at the real invariant — which items 1–3 are what establish.

## Verification

Every fix has a spec that fails against the unfixed code: reverting them turns 5 of the 41 error-handler tests red, including the throwing-report case with the exact `TypeError` seen on stage, the `{ message }` 409, and the empty-bodied 400.

Tests, lint and `nx run dsp-app:build` green for `vre-core-error-handler`, `vre-pages-search-search`, `vre-pages-search-advanced-search`, `vre-pages-project-project`, `dsp-app`.

## Not addressed

- A **400** whose body is `{ 'knora-api:error': … }` — the hand-written JSON-LD v2 services — falls off the end of that branch and shows no snackbar at all. Same DEV-6872 theme, but it changes a working path rather than stopping a throw; worth its own issue.
- `ResourcesListFetcherComponent` has no `ChangeDetectionStrategy.OnPush`, unlike its two siblings.
- The advanced-search failure/retry path is covered by a Storybook `play()` story but has no Jest spec.


mpro7:
@claude review

libs/vre/core/error-handler/src/lib/error-reporting.service.ts
```diff
@@ -143,34 +144,16 @@ export class ErrorReportingService {
    */
   private static _reasonOf(error: unknown): string | undefined {
     if (error instanceof ApiResponseError) {
-      return typeof error.error === 'string'
-        ? error.error || undefined
-        : ErrorReportingService._reasonFromBody(error.error.response);
+      return typeof error.error === 'string' ? error.error || undefined : reasonFromErrorBody(error.error.response);
```
domsteinbach: Why error.error || undefined ?
ApiResponseError's error property is typed string or AjaxError and defaults to an empty string. There shouldnt be the need of "|| undefined". The whole _reasonOf() returns always a string imho. So we could also check the callers. They might also be simpler due to the strict return type of string only.

domsteinbach (approved):
Please check my comment

PR #3320, opened 2026-08-07
fix(search): show the server's reason in the search failure panel

mpro7:
[DEV-6866](https://linear.app/dasch/issue/DEV-6866)

Follows the reopen: *"Searching for `de*` gives a 400 immediately, but the spinner goes forever."*

## What actually caused the reported spinner — and why it is already gone

Not a missing error path. **`AppErrorHandler` was throwing**, so the `catchError` that stops the spinner never finished.

My PR #3298 added a 4th constructor dependency to `AppErrorHandler` (`ErrorReportingService`), while `app.config.ts` still declared three:

```ts
{ provide: ErrorHandler, useClass: AppErrorHandler, deps: [NotificationService, AppConfigService, NgZone] }
```

`useClass` **plus an explicit `deps` list** constructs positionally, so `_errorReporting` was `undefined` and `handleError` threw a `TypeError`. At that time the search `catchError` called `handleError` first, so `loading.set(false)` and `failed.set(true)` never ran — for **every** error, which is why the report says "other errors weren't handled" rather than naming a single class.

That was fixed by **#3314** (`provideAppErrorHandler()`, no hand-written deps) — shipped in **v13.12.2**, merged four hours after the report — and hardened by **#3315**, which reorders the failure handlers so the state is committed before the global handler runs. Verified by driving the real `de*` 400 through the real `AppErrorHandler`: the spinner stops, the panel appears, the snackbar carries the reason, and the stream stays alive so Retry works.

**Please retest on an environment running ≥ v13.12.2.**

## What this PR changes

The remaining defect the reopen exposes, which no version fixes: for `de*` the panel says *"Something went wrong… Please try again"*. Retrying is useless — the server will reject that query every time — and the sentence that would let the user fix it (*"a wildcard search term must contain at least 3 characters besides the wildcard"*) only existed in the 5-second snackbar.

- **`userFacingReason(error)`** (`vre/core/error-handler`) returns dsp-api's explanation only for the statuses whose body is meant for the user — **400** (names the broken constraint) and **409** (names the conflict). 403/404/504 already have curated messages; 500 is excluded because its body carries server internals. It also strips the `dsp.errors.BadRequestException:` prefix the older v2 shape adds.
- **`reasonFromApiError(error)`** reads the reason off a whole `ApiResponseError`/`HttpErrorResponse`; `ErrorReportingService` was duplicating that shape and now delegates.
- **`SearchFailedComponent`** takes an optional `reason` and shows it instead of the generic message. Wired into all three failure paths (fulltext search, advanced search, sidenav resource list).

The panel now carries the same sentence as the snackbar, and falls back to the generic wording when the failure has nothing actionable to say.

## Test Plan

- New `api-error-reason.spec.ts`: body-shape precedence, whole-error reading, the real `de*` body, the `dsp.errors.BadRequestException:` strip, the 409 case, and that 500/502/504/403/404 bodies are withheld.
- `search-result.component.spec.ts`: a 400 carrying the real wildcard message surfaces it; a failure with no usable reason falls back to `undefined`.
- New `WithServerReason` story asserting the explanation replaces the generic advice.
- `nx affected` across **16 projects**: lint and test green. `dsp-app:build` clean. Storybook interaction tests pass in chromium.

## How this class of mistake gets avoided

Three separate things had to line up, and each is worth a different guard:

1. **`useClass` + `deps` is a silent trap.** The list is positional and TypeScript cannot check it against the constructor, so adding a dependency compiles fine and breaks DI at runtime. Dropping `deps` and letting Angular read the class's own metadata — what `provideAppErrorHandler()` now does — makes the desync structurally impossible. A lint rule banning `deps:` next to `useClass:` would enforce it repo-wide; happy to add it if wanted.
2. **Nothing tested the app's real DI wiring.** The break lived in `app.config.ts`, which no spec exercises — my whole DEV-6866 suite stayed green while the app was broken. A smoke test that resolves the real provider set would have caught it in CI.
3. **A failure handler must not depend on a side-effecting call succeeding.** Putting `handleError` first meant one throw defeated the entire fix. Commit local state first, notify outward last — now done in all three components.

### DIFF
--- a/libs/vre/core/error-handler/src/index.ts
+++ b/libs/vre/core/error-handler/src/index.ts
@@ -1,3 +1,4 @@
+export * from './lib/api-error-reason';
 export * from './lib/app-error-handler';
 export * from './lib/app-error-handler.providers';
 export * from './lib/app-error';
--- a/libs/vre/core/error-handler/src/lib/api-error-reason.spec.ts
+++ b/libs/vre/core/error-handler/src/lib/api-error-reason.spec.ts
@@ -1,0 +1,83 @@
+import { HttpErrorResponse } from '@angular/common/http';
+import { ApiResponseError } from '@dasch-swiss/dsp-js';
+import { AjaxError } from 'rxjs/ajax';
+import { reasonFromApiError, reasonFromErrorBody, userFacingReason } from './api-error-reason';
+
+const makeApiResponseError = (status: number, response: unknown): ApiResponseError => {
+  const ajax = Object.create(AjaxError.prototype) as AjaxError;
+  Object.assign(ajax, { status, response, message: `ajax error ${status}`, name: 'AjaxError' });
+  const error = Object.create(ApiResponseError.prototype) as ApiResponseError;
+  Object.assign(error, { status, error: ajax, url: '/v2/search/x', method: 'GET' });
+  return error;
+};
+
+describe('reasonFromErrorBody', () => {
+  it.each([
+    ['a bare string body', 'plain failure', 'plain failure'],
+    ['an empty string body', '', undefined],
+    ['a JSON-LD knora-api:error', { 'knora-api:error': 'jsonld reason' }, 'jsonld reason'],
+    ['an OpenAPI { message }', { message: 'openapi reason' }, 'openapi reason'],
+    ['an { error } body', { error: 'error-field reason' }, 'error-field reason'],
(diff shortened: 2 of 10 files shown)


mpro7:
@claude review

domsteinbach (approved):

PR #3326, opened 2026-08-10
fix(error-handler): show the reason of a 400 whose body is knora-api:error

mpro7:
[DEV-6922](https://linear.app/dasch/issue/DEV-6922)

## Summary

- A 400 arriving as an `HttpErrorResponse` with a `{ "knora-api:error": … }` body showed **no snackbar at all**. The branch tested three shapes in turn and returned, so the shape the hand-written JSON-LD v2 services answer with (`ResourceLegalV2ApiService` and friends) matched nothing — the failure reached Sentry and the user, who had caused the 400, was told nothing.
- All four shapes now read through `reasonFromErrorBody`, the precedence already shared with telemetry and the search failure panel. This path now agrees with the JS-LIB one on which field wins when a body carries several.
- One snackbar per 400 instead of one replacing another, and the half of dsp-api's exception text that names the user's mistake is the half that survives.

## Changes

**`app-error-handler.ts`** — the 400 branch becomes `handleBadRequest`, which reads the reason once and shows it. The two extractions dsp-api's raw exception text needs move into `readableExceptionText`: the exception class prefix is stripped first, and the trailing parenthesised detail is what remains for the `Invalid request (…)` shape, which carries no class. `testInvalidRequest` is gone, folded into that method. A body that opens as markup is dropped rather than shown.

**`api-error-reason.ts`** — adds `declaredMessageOf`, the guard on the tidying. A declared `{ message }` is a sentence dsp-api wrote for the user, so extracting a trailing parenthesis out of it would swap the whole explanation for its last clause. Shape knowledge stays in the module that owns it.

### Behaviour deltas

| 400 body | before | after |
|---|---|---|
| `{ 'knora-api:error': … }` | *nothing* | the reason |
| `{ error: 'dsp…: msg' }` | ` msg` | unchanged |
| `{ error: 'dsp…: msg (detail)' }` | `msg (detail)` then `detail` | ` msg (detail)`, one bar |
| `'bare string (detail)'` | `detail` | unchanged |
| `{ message: 'x' }` | `x` | unchanged |
| `{ message: 'x (y)' }` | `x (y)` | unchanged — the guard |
| `{ error: 'no class here' }` | *nothing* | `no class here` |
| `'bare string, no parens'` | *nothing* | the string |
| `'<html>…400 from a proxy…'` | *nothing* | unchanged — dropped as markup |
| empty / no reason | *nothing* | unchanged |

Row 3 is the one place where "keep today's visible text" and "show the useful half" disagree, and it is decided by a real message: `SearchResponderV2` answers a malformed search with `BadRequestException(s"Invalid search string: '$searchValue' ($err)")`, where `$err` is Lucene's parse failure. Taking the parenthesis first showed `org.apache.lucene…ParseException: Cannot parse 'de*'` and dropped the half naming the mistake — the same text `userFacingReason` withholds from the failure panel outright (DEV-6866), and the opposite of what `handleGenericError` shows for that body.

The reason is otherwise shown verbatim, matching what the JS-LIB 400 path has always done. There is still deliberately no generic fallback here: a 400 with an empty body has nothing to say that the failed action has not already said.

## Test Plan

`nx run vre-core-error-handler:test` — 76 passed, lint and prettier clean.

New `describe` for the `HttpErrorResponse` 400 path covering all four shapes, the precedence when a body carries both fields, the `{ message }` guard, the extraction order against the real Lucene message, the single-bar property, the markup body and the silent cases. Verified against the unfixed handler: **4 of them fail** (the three `knora-api:error` cases and the bar count), the rest pass on both — so the regression guards really are guards.

### DIFF
--- a/libs/vre/core/error-handler/src/lib/api-error-reason.ts
+++ b/libs/vre/core/error-handler/src/lib/api-error-reason.ts
@@ -27,6 +27,22 @@ export function reasonFromErrorBody(body: unknown): string | undefined {
   );
 }
 
+/**
+ * The `{ message }` of an exception declared in the OpenAPI spec, when the body carries one.
+ *
+ * Lives next to `reasonFromErrorBody` so that knowledge of the response shapes stays in one module: a
+ * caller that tidies up raw exception text before showing it needs to know whether the reason it was
+ * handed is that text or a sentence dsp-api already wrote for the user.
+ */
+export function declaredMessageOf(body: unknown): string | undefined {
+  if (!body || typeof body !== 'object') {
+    return undefined;
+  }
+
+  const message = (body as { message?: unknown }).message;
+  return typeof message === 'string' && message.length > 0 ? message : undefined;
+}
+
 /** The same reason, read off a whole error rather than a body already dug out of it. */
 export function reasonFromApiError(error: unknown): string | undefined {
   if (error instanceof ApiResponseError) {
--- a/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
+++ b/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
@@ -182,6 +182,16 @@ describe('AppErrorHandler', () => {
       );
     });
 
+    it('says nothing when the body is a page rather than a message', () => {
+      // Angular assigns the raw text when an error body does not parse as JSON, so a 400 from a proxy
+      // ahead of the API arrives as a whole HTML page. MatSnackBar would show all of it, escaped.
+      handler.handleError(
+        badRequest('<html>\r\n<head><title>400 Request Header Or Cookie Too Large</title></head>\r\n</html>\r\n')
+      );
+
+      expect(openSnackBar).not.toHaveBeenCalled();
+    });
+
     it('does not throw on a 400 with an empty body', () => {
       // Angular leaves `error` null when a 400 arrives with no body, as a gateway or proxy answers.
       // There is nothing to tell the user, but throwing here costs them the snackbar for every
--- a/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
+++ b/libs/vre/core/error-handler/src/lib/app-error-handler.spec.ts
@@ -98,12 +98,96 @@ describe('AppErrorHandler', () => {
       expect(instant).toHaveBeenCalledWith('core.errorHandler.contactSupport');
       expect(openSnackBar).toHaveBeenCalledWith('core.errorHandler.contactSupport', 'error');
     });
+  });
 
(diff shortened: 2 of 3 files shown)


mpro7:
@claude review

domsteinbach (approved):

mpro7:
Triaged all three against first-party sources. **1 and 2 are real and are fixed** (098558650, 704d17002); **3 is a false flag** — the schema refutes it.

**1 — real, and the example is better than the hypothetical one.** dsp-api has exactly two messages ending in a parenthesis, both `SearchResponderV2:1166/1226`:

```scala
.mapError(err => BadRequestException(s"Invalid search string: '$searchValue' ($err)"))
```

`$err` is Lucene's parse failure, so the old ordering showed `org.apache.lucene.queryparser.classic.ParseException: Cannot parse 'de*'` and dropped `Invalid search string: 'de*'` — the exact text `userFacingReason` withholds from the failure panel (DEV-6866). Fixed as suggested: class prefix first, parenthesis only as the fallback for the `Invalid request (…)` shape that carries no class. That also changes the pre-existing `{ error: 'dsp…: msg (detail)' }` outcome from `detail` to ` msg (detail)`, which the same evidence says is the right half; the delta table records it. Both call sites are reached through dsp-js, so nothing user-visible was broken on the changed path today — the fix is a design correction, not a live bug.

**2 — real, mechanism confirmed in both backends.** `_module-chunk.mjs:1130` (XHR: `body = originalBody`, because the `{ error, text }` wrapping is guarded by `if (ok)`) and `:919-920` (fetch: `if (status < 200 || status >= 300) return text`). So the raw page does reach `HttpErrorResponse.error` and would have gone to `MatSnackBar.open` untruncated. Guarded on markup rather than on length, so dsp-api's plain-text 400s still surface; a spec covers the nginx *Request Header Or Cookie Too Large* body.

**3 — false flag.** The declared error schema is `{ message, errors: ErrorDetail[] }` (`dsp-api_spec.yaml:16849-16860`) — the companion field is `errors`, a plural array that `reasonFromErrorBody` never reads. There is no dsp-api shape carrying both `message` and `error`, so the first consequence has no input that produces it. The second needs one response to duplicate identical text across `knora-api:error` and `message`, which are emitted by two different serializers; nothing in `dsp/errors/Errors.scala` or the spec does that. Keeping `declaredMessageOf` — a `{ field, reason }` return on the shared `reasonFromErrorBody` would widen the signature used by telemetry and the failure panel to settle a shape that does not occur.

Your summary note is right and needs no change: the old `error.error?.error` branch called `.match()` on whatever that field held, so an object-valued `error` threw — `reasonFromErrorBody`'s `typeof` check removes that.

PR #3327, opened 2026-08-10
fix(search): enforce the term rules in the search inputs (DEV-6930)

mpro7:
[DEV-6930](https://linear.app/dasch/issue/DEV-6930)

## Summary

- dsp-api rejects several search terms outright, and all three search inputs sent them anyway — one round trip per keystroke-submit, ending in a 400 and an error panel. Each input now says inline what to change and issues no request.
- `GlobalSearchComponent` already declared the rules and never consulted them; `ProjectFulltextSearchPageComponent` had no validators at all behind a `formGroup.valid` guard that was therefore always true; the advanced search bar had none either.
- **The rule is not the same on both search paths** — see below. This deviates from the issue's suggested scope, deliberately.

## The two endpoints do not enforce the same rule

The issue's table was measured against `/v2/search/:term`. The advanced search bar sends its fulltext through Gravsearch `knora-api:matchFulltext`, which I measured separately against the dev API:

| Input | `/v2/search/:term` | Gravsearch `matchFulltext` |
| -- | -- | -- |
| `de` | 400 | 400 |
| `de*` | 400 | **200** |
| `hello de*` | 400 | **200** |
| `*de` | 400 | **200** |
| `ide*`, `buch de` | 200 | 200 |

Only the min-length half is shared. Applying the issue's full rule to the advanced bar would have blocked `de*` — a search that endpoint runs happily. So there are two validators, and the advanced bar gets only the half its endpoint enforces.

## Changes

**New shared validators** (`libs/vre/shared/app-common/.../search-term.validator.ts`)
- `searchTermMinLengthValidator` — whole trimmed term ≥ 3 characters, measured on the whole string rather than per token, so `buch de` passes. Holds on every search path.
- `fulltextSearchTermValidator` — the above plus the per-token rule `/v2/search/:term` adds: every whitespace-separated token carrying a wildcard needs 3 characters besides its wildcards. Rejects `de*`, `id*`, `*de`, `hello de*`; accepts `ide*`, `de*x`, `abcd*`, `*abc`.
- Empty is left to `Validators.required` in both, so the advanced bar can still clear its term.

**Enforcement**
- `GlobalSearchComponent` — `search()` returns early when invalid; message floats under the field (the toolbar has a fixed height) and suppresses the search-tips overlay while shown, since both occupy that spot.
- `ProjectFulltextSearchPageComponent` — validator added, so its existing guard now guards something.
- `AdvancedSearchBarComponent` — an invalid term is not written to the URL, so the query stays on the last term that searched.

**Message timing.** Keying the message off `touched`/`dirty` turned every input red at the second keystroke of any term (`de` on the way to `deutsch`). Each input speaks at the moment it would have searched instead: on submit for the two form inputs, at the debounce for the advanced bar, which has no submit. A deep link seeding the advanced bar with a refused term still shows the message.

**i18n** — `pages.search.termValidation.{tooShort,wildcardTooShort}` in de/en/fr/it.

## What this does not remove

DEV-6866's failure panel stays necessary. `/search/:q` reads the term from the route, so `…/search/de*` never touches a form, and the other 400s (`"unbalanced`, `AND`, `de?`, `de~`) cannot be prevented without reimplementing a Lucene parser in the client.

## Test Plan

- `nx run vre-shared-app-common:test` — 24 new cases over the measured table, both validators, including the per-token wildcard case and the multi-wildcard token.
- `nx run-many --target=lint --projects=<4 changed>` — clean.
- `nx run dsp-app:build` — clean under strict template checking.
- Full Storybook interaction suite: **518/518**, including three new `Shared / Header / Global Search` stories asserting the inline message for `de` and `de*` and its absence for `ide*`.
- Endpoint behaviour in the table above measured directly against `https://api.dev.dasch.swiss`.

## Screenshots


mpro7:
@claude review

mpro7:
Reviewed all four findings against the pinned sources; three were real and are fixed in ff2972921 and 14ce67aaa.

**🔴 #1 — advanced bar's message hidden until blur — confirmed, fixed.** Reproduced with a test before fixing: set `de`, advance the 300 ms debounce, and `mat-error` is absent from the DOM. The mechanism is as described — `_form-field-chunk.mjs:900` gates the subscript on `_control.errorState`, the default matcher (`_error-options-chunk.mjs:33`) is `invalid && (touched || form.submitted)`, there is no parent form on this control and no global `ShowOnDirtyErrorStateMatcher` in the app.

Worth noting the review understated it slightly: the deep-link seed had the *same* hole, not merely a falsified comment — `setValue(q, { emitEvent: false })` leaves the control untouched, so `?q=de` showed nothing either. Both paths now go through one `_refreshFulltextError()` helper that sets the signal and marks touched. The three requested tests are in `spec/advanced-search-bar.component.spec.ts`; the middle one fails against the unfixed code and passes against the fix.

**🟡 #2 — one-way `searchAttempted` — confirmed, fixed.** Both flags now clear on a search that goes through, so the next term starts quiet. This mattered more on the project page, where `FormGroupDirective.submitted` also stays true forever and the flag was the only thing holding the message back.

**🟡 #4 — missing `STORY_TRANSLATIONS` key — confirmed, fixed.** Added `pages.search.termValidation.tooShort` to `stories.helpers.ts`. This one was a documented trap I walked into (`apps/dsp-app/src/assets/i18n/CLAUDE.md:57`) — I checked the four locale files and not the `CLAUDE.md` sitting beside them.

**🟡 #3 — empty submit is a silent no-op — acknowledged, not changing.** The behaviour change is real, but the behaviour it replaced was navigating to `/search/` on an empty term, which did nothing useful either. Adding "Enter at least 3 characters" to an empty field reads as a complaint about a field the user has not filled in yet. Happy to revisit if a reviewer disagrees.

Verification after the fixes: `vre-shared-app-common` 38, `vre-shared-app-common-to-move` 69, `vre-pages-search-advanced-search` 242, `vre-pages-search-search` 8 — all green; lint clean on all four projects; `dsp-app:build` clean; full Storybook interaction suite 518/518.

domsteinbach (approved):

PR #3333, opened 2026-08-11
fix(error-handler): stop claiming the user is offline on any status 0 (DEV-6935)

mpro7:
[DEV-6935](https://linear.app/dasch/issue/DEV-6935)

## Summary

- `AppErrorHandler` treated an HTTP `status: 0` as proof the user is offline — *"It seems that you are not connected to internet."* But `status: 0` is what the browser reports for **every** failure it cannot attribute to a response: a CORS rejection, DNS, TLS, an aborted request, a blocking extension, and a genuine loss of connectivity.
- Found by searching `\*bc` on dev with the connection perfectly fine: the shared ingress rejects any path containing `%5C` before routing, with no CORS headers at all, so the browser exposes nothing of the response and the status collapses to `0`. The app then sent the user to check their wifi for a request our own infrastructure had refused.
- The message now names only what is known — the server was not reached — while keeping the same actionable advice.

## Changes

**Message** — `libs/vre/core/error-handler/src/lib/app-error-handler.ts:123` resolves `core.errorHandler.serverUnreachable` instead of `noInternet`, with a comment recording why a `0` cannot be read as "offline".

**i18n** — the key is renamed rather than reworded in place, so it stops asserting a cause too. All four languages; no `rm.json` exists, Romansh is bound to English by the runtime fallback loader (DEV-6629). Key sets verified identical across `en`/`de`/`fr`/`it`.

**Telemetry comments** — `error-reporting.service.ts` justified grading a `0` as a warning by asserting it is "almost always the user's own connectivity — the app says as much, mapping it to the 'no internet' message". Neither half survives this fix, and left standing the comment hands the next reader back the exact premise the fix removes. The grading itself is unchanged: whether a `0` should now be an `error` rather than a `warning` is an alerting decision, not a comment fix.

**Not included** — no `navigator.onLine === false` branch. The single message carries the same advice the offline wording gave, so the branch would buy slightly nicer phrasing in a case already covered, at the cost of a conditional, a retained key in four languages, and a test.

## Test Plan

- `nx run vre-core-error-handler:test` — 78 passed, 5 suites.
- The two new cases in `app-error-handler.spec.ts` were **verified to fail against the unfixed code** (2 failed / 76 passed with the message key reverted) rather than assumed to. They cover the `HttpErrorResponse` ingress-rejection shape and the JS-LIB path.
- `nx run vre-core-error-handler:lint`, prettier, and `tsc -p apps/dsp-app/tsconfig.app.json --noEmit` all clean.
- Repo-wide grep for both `noInternet` and the prose "no internet" returns nothing.

## Found while reviewing, filed separately

[DEV-6946](https://linear.app/dasch/issue/DEV-6946) — the IIIF branch immediately below this one is **unreachable**, so `core.errorHandler.iiifServerError` is dead in all four languages. Angular composes `message` as `` `…: ${status} ${statusText}` `` ([`_module-chunk.mjs:682`](https://github.com/angular/angular/blob/main/packages/common/http/src/response.ts)), so the `0` in the matched literal `knora.json: 0 Unknown Error` *is* the status — every input that could satisfy it has already been taken by `status === 0`. The ordering predates this PR and is present on `main`; it is a separate user-visible behaviour change needing its own test, so this PR was kept to DEV-6935's stated scope.

## Screenshots

Snackbar text only; no layout change.

### DIFF
--- a/apps/dsp-app/src/assets/i18n/de.json
+++ b/apps/dsp-app/src/assets/i18n/de.json
@@ -971,7 +971,7 @@
   },
   "core": {
     "errorHandler": {
-      "noInternet": "Es scheint, dass Sie nicht mit dem Internet verbunden sind.",
+      "serverUnreachable": "Der Server konnte nicht erreicht werden. Bitte überprüfen Sie Ihre Verbindung und versuchen Sie es erneut.",
       "iiifServerError": "IIIF-Serverfehler: Das Bild konnte nicht geladen werden. Bitte versuchen Sie es später erneut.",
       "noPermission": "Sie haben keine Berechtigung, eine oder mehrere Ressourcen zu sehen.",
       "notFound": "Die angeforderte Ressource wurde nicht gefunden.",
--- a/apps/dsp-app/src/assets/i18n/en.json
+++ b/apps/dsp-app/src/assets/i18n/en.json
@@ -971,7 +971,7 @@
   },
   "core": {
     "errorHandler": {
-      "noInternet": "It seems that you are not connected to internet.",
+      "serverUnreachable": "Could not reach the server. Please check your connection and try again.",
       "iiifServerError": "IIIF server error: The image could not be loaded. Please try again later.",
       "noPermission": "You do not have permission to see one or more resources.",
       "notFound": "The requested resource was not found.",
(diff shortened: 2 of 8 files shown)


derschnee68:
@claude review

domsteinbach (approved):

PR #3351, opened 2026-08-18
refactor: extract the blob-download anchor block into a shared helper (superseded)

mpro7:
The branch has been deleted.


mpro7:
Folded into #3349 — the helper extraction ships with the encoding fix rather than as a separate PR.

PR #3352, opened 2026-08-18
fix: show "Placeholder" in the legal overlay and stop its license column colliding into the other rows

domsteinbach:
Fixes [DEV-6982](https://linear.app/dasch/issue/DEV-6982/legal-overlay-shows-raw-urndaschplaceholder-instead-of-placeholder) and [DEV-6983](https://linear.app/dasch/issue/DEV-6983/legal-overlay-unlabelled-license-column-collides-into). Both land in the same component, so they are fixed together rather than in two conflicting PRs.

## DEV-6983 — the license column collided into the other rows

`resource-legal.component.ts` laid the panel out as a two-child `justify-content: space-between` flex row. Four defects compounded: no `gap`, no width bound on the license column, `align-items: center` vertically centring wrapped license lines against the two left-hand rows, and no label on the license row while the other two had one. Any license label over roughly 40 characters therefore interleaved into the values beside it and read as one run-on sentence.

This reproduces on ordinary data, independent of placeholders — the built-in `AI-Generated Content - Not Protected by Copyright` is 49 characters. Of the live built-in licenses, every label is ≤53 characters except the placeholder at 96.

Replaced with the `meta-grid` pattern already used by `region-preview-viewer.component.ts` (`grid-template-columns: auto 1fr`), and added the `resourceEditor.legal.license` label — which already existed in all four locales and was simply unused here. The `open_in_new` icon is now inline rather than a flex sibling, so it trails the last wrapped word instead of floating to the far right.

Before / after, at a narrow width so the label wraps:

```
Urheberrechtsinhaber   urn:dasch:placeholderPlaceholder License - Not a Real License only to be used when
Autorenschaft          urn:dasch:placeholderthe real license is not known yet.                        ↗
```

```
Copyright holder   University of Basel
Authorship         Lotte Reiniger, Hilma af Klint
License            AI-Generated Content - Not
                   Protected by Copyright ↗
```

`ResourceLegalComponent` feeds all ten representation viewers (archive, audio, audio-segment, compound, document, pdf, still-image, text, video, video-segment), so one fix covers them all.

## DEV-6982 — the raw sentinel was rendered

dsp-tools writes `urn:dasch:placeholder` into `copyrightHolder`, `authorship` and `licenseIri` when the real legal info is not known yet. dsp-app rendered the internal URN verbatim, plus the placeholder license's 96-character prose label behind a dead `<a href="urn:dasch:placeholder">` — the sentinel is both `id` and `uri`, and is not dereferenceable.

The existing `isPlaceholderFileValue()` could not be reused: it checks only `filename`, so it does not detect placeholder *legal* fields. Added `isPlaceholderLegalValue()` beside it, reusing the existing `PLACEHOLDER_FILE_SENTINEL`, and a `resourceEditor.legal.placeholder` key in en/de/fr/it — reusing this repo's established wording for "Placeholder" (Platzhalter / Espace réservé / Segnaposto) rather than inventing new terms. No `rm.json`: Romansh is bound to English by the fallback loader (DEV-6629).

Fixed in both renderers that read legal info from a file value: `resource-legal*` and `region-preview-viewer`.

## Scope: one listed file deliberately not changed

DEV-6982 also lists `libs/vre/ui/ui/src/lib/resource-rights-statement.component.ts`. I left it alone, and this is the one point reviewers pushed back on, so it is worth stating plainly.

Its license comes from the **project-level** `project.dataLicense` via `ProjectDataRightsService`, not from a file value, and the admin picker filters candidates to the `http://rdfh.ch/licenses/cc-by` prefix — so the sentinel cannot be selected through the UI. On the reproduction project `dataLicense` is `cc-by-4.0`. Covering it would also force `PLACEHOLDER_FILE_SENTINEL` into a shared lib, since `resource-editor` already imports `@dasch-swiss/vre/ui/ui` and the reverse import would create a dependency cycle.

**Caveat, tracked separately in [DEV-6994](https://linear.app/dasch/issue/DEV-6994/resource-rights-statement-renders-the-raw-placeholder-sentinel-project).** "Cannot reach it" was too strong. The vendored OpenAPI spec lists `urn:dasch:placeholder` as an *enabled* license, so `ProjectDataRightsService._resolve()` would resolve it if any non-UI path (API-direct write, import, migration) ever set `dataLicense` to the sentinel. Separately, `res.resourceAuthorship` — the data-side field, distinct from the file-value `authorship` fixed here — also renders unfiltered.

That follow-up is filed as **low priority**: placeholders are not possible on a productive environment (they occur only on test servers, where the affected display is internally visible to admins), so it is a correctness/consistency cleanup rather than a user-facing defect. The suggested fix there is a central filter in `_resolve()` — which also requires moving `PLACEHOLDER_FILE_SENTINEL` into a shared lib, since `resource-editor` already imports `@dasch-swiss/vre/ui/ui` and the reverse import would cycle.

## Verification

- 29 projects lint clean; 395 unit tests pass (uncached); production build clean under strict template typechecking
- 7 Storybook interaction tests pass, including three new stories
- Verified in the browser at a narrow width: the license wraps inside its own column, stays aligned with the values above, and the `↗` trails the last word

New coverage: a long-license-label story (the fixture gap that let DEV-6983 ship — the longest existing fixture was `University of Basel`), placeholder stories for both legal components, `region-preview-viewer` placeholder cases, and direct tests for the new helpers. The region-preview cases were checked by mutation: breaking the predicate fails two of the three.

## From the review pass

- **Accessibility.** Added `rel="noopener noreferrer"` and an `opens in a new tab` aria-label to both license links, matching the existing pattern in `resource-rights-statement.component.ts` (the `legal.dataSide.opensInNewTab` key already existed in all four locales). Gave the logo-variant link a meaningful accessible name (`[alt]="license.labelEn"` rather than `alt="license"`). Raised the `.license-pill` border to `#6b7280` on `#f9fafb`: 2.4:1 → 4.6:1, clearing the 3:1 non-text contrast bar. Distinguished the non-interactive placeholder pill from the clickable one — they were pixel-identical.
- **Translate in the template.** `joinPlaceholderLegalValues` now takes an already-translated label resolved by `| translate`, so both components drop their `TranslateService` injection and a duplicated key constant. Reviewers flagged the previous component-side `.instant()` as a latent stale-language bug under OnPush; **I could not reproduce that** — in a test harness neither form updates without a change-detection pass, and the language switcher is a click handler, so a pass always follows. The refactor is kept for the smaller API surface and convention fit, not as a bug fix, and the comments say so.
- **Caught while verifying.** An intermediate CSS `::before` separator rendered the comma visually but omitted it from `textContent`, so copy/paste and screen readers saw `Lotte ReinigerHilma af Klint`. Joining in the component keeps the separator in a real text node.

## Note for reviewers

The repo requires Node `^22.13.0` (`.nvmrc` pins 22.23.1). On Node 20 the OpenAPI generator fails with an ESM error and Storybook dies on a native-binding error. `nvm use 22.18` or later is enough.

### DIFF
--- a/apps/dsp-app/src/assets/i18n/de.json
+++ b/apps/dsp-app/src/assets/i18n/de.json
@@ -766,7 +766,8 @@
     "legal": {
       "copyrightHolder": "Urheberrechtsinhaber",
       "authorship": "Autorenschaft",
-      "license": "Lizenz"
+      "license": "Lizenz",
+      "placeholder": "Platzhalter"
     },
     "navigation": {
       "firstPage": "Erste Seite",
--- a/apps/dsp-app/src/assets/i18n/en.json
+++ b/apps/dsp-app/src/assets/i18n/en.json
@@ -568,7 +568,8 @@
     "legal": {
       "copyrightHolder": "Copyright holder",
       "authorship": "Authorship",
-      "license": "License"
+      "license": "License",
+      "placeholder": "Placeholder"
     },
     "navigation": {
       "firstPage": "First page",
--- a/apps/dsp-app/src/assets/i18n/fr.json
+++ b/apps/dsp-app/src/assets/i18n/fr.json
@@ -568,7 +568,8 @@
     "legal": {
       "copyrightHolder": "Titulaire des droits d'auteur",
       "authorship": "Auteur·ice",
-      "license": "Licence"
+      "license": "Licence",
+      "placeholder": "Espace réservé"
     },
     "navigation": {
       "firstPage": "Première page",
--- a/apps/dsp-app/src/assets/i18n/it.json
+++ b/apps/dsp-app/src/assets/i18n/it.json
@@ -568,7 +568,8 @@
     "legal": {
       "copyrightHolder": "Titolare del copyright",
       "authorship": "Autore",
-      "license": "Licenza"
+      "license": "Licenza",
+      "placeholder": "Segnaposto"
     },
     "navigation": {
       "firstPage": "Prima pagina",
--- a/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/template-switcher/viewer-components/region-preview-viewer.component.spec.ts
+++ b/libs/vre/resource-editor/resource-editor/src/lib/properties/properties-display/template-switcher/viewer-components/region-preview-viewer.component.spec.ts
@@ -4,6 +4,7 @@ import { AdminAPIApiService } from '@dasch-swiss/vre/3rd-party-services/open-api
 import { ResourceService } from '@dasch-swiss/vre/shared/app-common';
 import { provideTranslateService } from '@ngx-translate/core';
 import { of, throwError } from 'rxjs';
+import { PLACEHOLDER_FILE_SENTINEL } from '../../../../representation/is-placeholder-file-value';
 import { RepresentationService } from '../../../../representation/representation.service';
 import { ResourceFetcherService } from '../../../../representation/resource-fetcher.service';
 import { RegionPreviewViewerComponent } from './region-preview-viewer.component';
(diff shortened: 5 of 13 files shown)


mpro7:
@claude review

mpro7 (approved):

PR #3355, opened 2026-08-18
fix: stop the rights statement rendering the raw placeholder sentinel

domsteinbach:
Fixes [DEV-6994](https://linear.app/dasch/issue/DEV-6994/resource-rights-statement-renders-the-raw-placeholder-sentinel-project). Follow-up to [#3352](https://github.com/dasch-swiss/dsp-app/pull/3352), now merged.

## Why this exists

#3352 fixed the placeholder sentinel in the two **file-value** legal renderers and deliberately left the data-side Resource Rights Statement alone, on the reasoning that the sentinel "cannot reach" `project.dataLicense` because the admin picker filters candidates to the `cc-by` IRI prefix.

That reasoning only covered the UI form. `urn:dasch:placeholder` **is** a resolvable entry in the project license catalog (dsp-api's `allow-placeholder` defaults to `true`):

```
curl -s ".../admin/projects/shortcode/0870/legal-info/licenses?pageSize=100"
# → { "id": "urn:dasch:placeholder", "uri": "urn:dasch:placeholder",
#     "labelEn": "Placeholder License - Not a Real License only to be used when the real license is not known yet." }
```

So `ProjectDataRightsService._resolve()`'s `licenses.find(l => l.id === project.dataLicense)` matches it and surfaces the 96-character prose label behind a dead `urn:dasch:placeholder` link — exactly the defect #3352 fixed elsewhere.

## Scope: display only

Placeholders are an **import-time** artifact. dsp-tools writes them when a *required* thing has no data yet — a resource that needs a bitstream, or a link pointing at an entity that does not exist yet — because the backend would otherwise reject the import. They exist only on test servers carrying unfinished data for preview and modelling questions, never in production.

The frontend therefore only ever needs to **render them legibly as placeholders**. It does not create them, and it does not edit them. This PR contains no write-path changes: no filtering, no rewriting, no silently dropping a sentinel on save. Anything that would hide a placeholder rather than show it as one would work against the point of the concept.

## What changed

**Fixed centrally in `_resolve()`, not per renderer.** `ProjectDataRights` gains `isPlaceholderLicense` and `isPlaceholderCopyrightHolder`; a placeholder license short-circuits **before** the catalog lookup and emits no label or url. All four consumers (`resource-rights-statement-container`, `project-short-description`, `project-description-page`, `create-resource-form`) then render the readable marker without repeating the check.

The two visibility gates in the project description views were widened to include the flag — otherwise a project whose only legal info is a placeholder license would hide the section entirely.

The marker is resolved in the templates via `| translate` rather than snapshotted with `instant()`, so it stays correct across a runtime language switch. This covers the `aria-label` text as well, so screen readers get the marker instead of the raw URN.

**Shared-lib move.** `PLACEHOLDER_SENTINEL`, `isPlaceholderLegalValue` and `joinPlaceholderLegalValues` move to `libs/vre/shared/app-common`. They were stranded in `resource-editor`, which `ui/ui` and `app-helper-services` cannot import — `resource-editor` already imports `ui/ui`, so the reverse direction would be a dependency cycle. The helper stays dependency-free (pure functions over `string`/`string[]`, translated label passed in) so `app-common`'s dependency set is unchanged and no `@ngx-translate` dependency is added. The old path re-exports, so this lib's ten existing relative imports are untouched.

**i18n.** New `legal.dataSide.placeholder` in en/de/fr/it, reusing the wording already established for the file-value renderers. This component lives in the `legal.dataSide.*` namespace, so it gets its own key rather than reaching across into `resourceEditor.legal`.

## Severity

**Low, and deliberately so.** Placeholders are not possible on a productive environment — they occur only on test servers, where this display is internally visible to admins. This is a legibility fix, not a user-facing production defect.

## Verification

- Unit tests pass for every affected project: `vre-ui-ui` (46), `app-helper-services` (43), `vre-resource-editor-resource-editor` (399, 1 skipped)
- Lint clean across `vre-ui-ui`, `app-helper-services`, `vre-resource-editor-resource-editor`, `vre-shared-app-common`, `vre-pages-project-project`
- Storybook interaction tests cover the three placeholder display cases, asserting **absence** of anchors and the literal `'Ada Lovelace, Placeholder'` in the text content (not a CSS-only separator)
- The four new `_resolve()` cases were checked by **mutation**: removing the short-circuit fails two of them
- Verified in the browser — license row renders `License / Placeholder` with **zero** anchors

## Follow-up (not in this PR)

`resource-side-legal-form.component.ts` (project settings → legal) seeds its editable copyright-holder input and authorship chips straight from the raw project fields, and preselects the license by IRI. On a placeholder project the text input shows `urn:dasch:placeholder` and the license `mat-select` renders blank while the control still holds the sentinel — and `save()` PUTs it back. That is pre-existing code, untouched by this PR, and it is a *write* path rather than a display one, so it wants its own ticket.


mpro7:
@claude  review

mpro7 (approved):

PR #3073, opened 2026-04-29
chore: bump openseadragon to v6 and fix type errors

mpro7:
## Summary

- Cherry-picks the openseadragon 4→6 / @types/openseadragon 3→5 version bump from #3072
- Fixes two TypeScript compile errors introduced by the stricter bundled types in openseadragon v6

## What was failing

The Storybook build was failing with two `tsc` errors:

**`osd-drawer.service.ts:52`** — openseadragon v6 types `CanvasClickEvent.originalTarget` as `HTMLElement`. TypeScript no longer allows a direct `as SVGElement` cast since neither type sufficiently overlaps. Fixed with `as unknown as SVGElement`.

**`still-image.component.ts:151`** — `Viewer.open()` now expects `TileSourceSpecifier | TileSourceSpecifier[]` (a new named type with a required `tileSource` property) instead of a generic `object[]`. Fixed with `as any` (consistent with the two other `as any` casts on the same viewer reference in that method).

## Test plan

- [ ] CI lint / unit tests pass (they already passed on #3072 with no code changes)
- [ ] Storybook build passes (was the failing job in #3072 CI run)
- [ ] E2E tests pass (were timing out on #3072 due to the Storybook build failure blocking the pipeline)

Closes #3072

PR #3357, opened 2026-08-19
fix: let the mouse wheel scroll the resource page over media viewers

mpro7:
[DEV-6998](https://linear.app/dasch/issue/DEV-6998)

## Summary

Wheel zooming in the media viewers is gated behind an interaction: a pointer press inside the viewer enables it, a press anywhere else disables it again. The wheel gesture itself is unchanged for anyone working with the image — but the resource page can now be scrolled with the mouse while the pointer is merely passing over the media.

**No change to the zoom gesture, the zoom range, or any modifier key.**

## Root cause

`<body>` is the resource page's scroll element (no `height`/`overflow` in `styles.scss`) and the viewer fills most of the viewport (900px). OpenSeadragon registers a **non-passive** `wheel` listener on its canvas and calls `preventDefault()` on every wheel event to zoom — so with the pointer over the image the page could not be scrolled by mouse at all. Keyboard scrolling acts on the focused document, which is why it kept working. The SVG viewer reproduced the same trap with its own `preventDefault()`.

Measured on unfixed `main` (`app.dev.dasch.swiss`), **Firefox and Chrome alike**: `defaultPrevented: true`, `scrollY 0 → 0`.

### Corrections to the issue's analysis

- **`gestureSettingsMouse.scrollToZoom = false` would not have fixed it.** `onCanvasScroll` sets `preventDefault = true` independently of that flag, so the page still would not scroll — and wheel-zoom would be gone.
- **Not a regression from the OpenSeadragon v6 bump (#3073).** `onCanvasScroll` is byte-identical in 4.1.1, 5.0.1 and 6.0.2 — the behaviour predates it by years.
- **Not Firefox-specific.** Modern Firefox uses `wheel`, never the legacy `DOMMouseScroll` path; Chrome behaves identically. Stefan reported using Firefox *and* seeing no symptom, so Firefox was never the variable.

### Approaches rejected along the way

- **`cooperativeGestures: true`** (OSD 6.1.0) — works, but replaces wheel-zoom with Ctrl/Cmd+wheel for every user and project. An unrequested global UX change; rejected in review.
- **Zoom-limit passthrough** — pass the wheel to the page once zooming out further is impossible. Rejected after measuring the real app: it rests at zoom `0.5463` while `getHomeZoom()` is `0.5171` and `getMinZoom()` is `0.4654`, and zoom-out from rest genuinely works for three more ticks (`0.5463 → 0.4703 → 0.4657 → 0.4654`). Any correct threshold would have shrunk the image three times before the page moved. The premise "zooming out does nothing anyway" is false here, most likely because of `visibilityRatio: 1.0`.

## Changes

**Image viewer** — `open-sea-dragon.service.ts`: capture-phase `wheel` listener on `viewer.element`, gated on a `_wheelZoomEnabled` flag driven by a document-level `pointerdown`. Bypassed while `isFullPage()`. The document listener is removed on the viewer's `destroy` event, so it cannot leak across resource navigations.

Capture phase is deliberate: `stopPropagation()` keeps the event away from OpenSeadragon's listener on the inner canvas *and* from its `minScrollDeltaTime: 50` throttle, whose skipped-event branch calls `preventDefault()` unconditionally. Intercepting any later would suppress most of a wheel spin and make the page scroll stutter. The default action is untouched — the default *is* the page scroll.

**SVG viewer** — `vector-image.component.ts`: the same gate via a `document:pointerdown` host listener; fullscreen zooms unconditionally.

**PDF viewer** left alone — it never zoomed on wheel, only via its toolbar.

## Test plan

- `nx run vre-resource-editor-resource-editor:test` — 47 suites, 398 passed. `:lint` clean. `npm run build-prod` succeeds.
- **Fix-confirmation test**: a new case asserts that an unmodified wheel neither zooms nor calls `preventDefault()` before the viewer has been pressed — verified it **fails** against the unfixed handler. Plus cases for re-disabling on an outside press and for the fullscreen bypass.

### Browser verification — real app, dev API, Firefox and Chrome

| Check | Firefox | Chrome |
|---|---|---|
| 5 rapid wheel-down ticks scroll the page | PASS `0 → 600` | PASS `0 → 600` |
| gate intercepts before OpenSeadragon sees the events | PASS | PASS |
| wheel-**up** over the image scrolls the page back up | PASS `600 → 360` | PASS `600 → 360` |
| after pressing the image, wheel zooms and the page stays put | PASS | PASS |
| after pressing outside, the wheel scrolls the page again | PASS | PASS |

The full 600px for 5×120 is the smoothness evidence: nothing was throttled or suppressed.

## Known rough edges

- **No visual cue.** Someone used to wheel-zoom will wheel over the image and get a page scroll instead of a zoom. That is a sensible outcome rather than a dead gesture, and the toolbar zoom buttons are unaffected, but a cursor change or a brief hint may be wanted — happy to add it.
- **The toolbar sits outside `viewer.element`**, so clicking a toolbar zoom button disables wheel zooming until the image is pressed again. The buttons themselves keep working. Widening the activation region to the whole representation block would fix it at the cost of coupling the service to that element.
- **Touch** is unverified — no device available. The gate keys on `pointerdown`, so a tap or drag on the image activates it.

## Still open on the report

This fixes a real, longstanding, browser-independent trap, but it is **not confirmed** to be what Barbara reported: the behaviour is years old, while she describes it as recent. One question decides it, asked on the issue — *does the page scroll when the cursor is beside or below the image?* DEV-6998 should stay open until she answers, rather than being closed by this merge.

### DIFF
--- a/libs/vre/resource-editor/resource-editor/src/lib/resource/still-image/open-sea-dragon.service.ts
+++ b/libs/vre/resource-editor/resource-editor/src/lib/resource/still-image/open-sea-dragon.service.ts
@@ -26,6 +26,8 @@ export class OpenSeaDragonService {
 
   private _drawing = false;
 
+  private _wheelZoomEnabled = false;
+
   get drawing() {
     return this._drawing;
   }
@@ -57,6 +59,7 @@ export class OpenSeaDragonService {
 
     this._viewer = new OpenSeadragon.Viewer(viewerConfig);
     this._addCustomHandlers(this._viewer);
+    this._gateWheelZoomBehindInteraction(this._viewer);
   }
 
   toggleDrawing(): void {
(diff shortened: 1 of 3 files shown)


mpro7:
It turned out that it is not an issue anymore.

PR #3360, opened 2026-08-19
feat: show the resource class on search results (DEV-5452)

derschnee68:
## Problem

Fulltext search returns resources of different classes, but the result list showed only the resource label — so there was no way to tell a `Letter` from a `Book` without clicking each hit. Reported by a researcher on `symblight`.

## Change

Each search result now shows its resource class on the metadata line below the label:

```
Brief an Anna Müller
Letter | Found in: Text | Project: anything
```

Notable details:

- **The metadata line used to be gated on `foundIn.length > 0`**, i.e. it rendered only when a search keyword matched the label or a property. Advanced search passes no keyword, so that line never appeared there at all. It now renders whenever there is something to show, and the `|` separators appear only between the segments actually present.
- **The class follows the UI language**, read from `entityInfo.classes[type].labels` through `appStringifyStringLiteral`, falling back to the single-language `resourceClassLabel` — which dsp-js also fills in for deleted resources and for classes of ontologies it does not know.
- **No extra API call**: dsp-js already populates both fields when converting a search response.
- **The project sidenav ("browse by class") leaves the class off** via the default-off `showResourceClass` input, since every row there is the same class and the label would just repeat.
- The `.search` padding class now also applies when the class is shown. It was keyed on `searchKeyword !== undefined`, which would have left advanced-search rows with a second line and no vertical padding.

Enabled for global fulltext, project fulltext, and advanced/gravsearch — all three share one item component.

## Testing

- 3 spec cases for the label fallback chain (entityInfo → `resourceClassLabel` → empty)
- 3 stories: enabled, enabled-without-keyword (the advanced-search path), disabled
- `stories.helpers.ts`: `makeReadResource` now populates `entityInfo`/`resourceClassLabel` as dsp-js does, and `STORY_PROVIDERS` gained a `LocalizationService` stub, which the impure pipe needs in the stories that render the item for real
- 64 unit tests pass, 21 data-browser story interaction tests pass, lint clean, `dsp-app:build` succeeds with strict template checking

## Note for the reviewer

The mockup attached to DEV-5452 has an expired image signature, so I could not open it — the placement here was chosen from the layout options rather than confirmed against the design. Worth a look if you can see the original. Verified in Storybook, not yet against a live server with real search data.


derschnee68:
@claude review

domsteinbach (approved):
The issues mentioned are irrelevant as there is always a label. And there is always a class.

PR #3367, opened 2026-08-24
fix: translate the help page and footer (DEV-7006)

domsteinbach:
Closes [DEV-7006](https://linear.app/dasch/issue/DEV-7006/static-page-content-not-translated-for-non-english-ui-languages-help)

## Diagnosis

The ticket offered two hypotheses — missing de/fr/it/rm keys, or hardcoded strings. It is the second: the strings were never translatable.

- `HelpPageComponent` did not import `TranslatePipe` at all. Its headings were literals in the template, and the `docs` / `tools` / `support` card sets held literal English text in the TS class.
- `FooterComponent` was fully hardcoded.

## Changes

**Translations** — added `pages.help.*` and `shared.footer.*` to `en`, `de`, `fr`, `it` (38 keys per file), following the guidelines in `apps/dsp-app/src/assets/i18n/CLAUDE.md`: Swiss High German, combined gender forms (`Entwickler:innen`), British English in `en`. Romansh needs nothing — the fallback loader binds `rm` to `en` (DEV-6629).

**Help page** — headings and all three card sets now use keys. The dev-version banner and the copyright line use interpolation (`{{release}}`, `{{environment}}`, `{{year}}`) rather than being split around markup.

**Footer** — all 11 strings translated. The footer also renders on the **project overview page**, so that page is fixed by the same change.

**Grid** — `GridComponent` passes `title`, `text` and `urlText` through the `translate` pipe, so `GridItem` now carries translation keys; the interface documents this. The DSP-APP / DSP-API / Sipi titles stay verbatim because they carry a version number appended at runtime, and the pipe echoes back anything it cannot resolve. `GridItem` has no consumers outside this library, so the contract change is contained.

## Scope note

The ticket asks to also check the create-project form. It is already fully translated via `pages.project.reusableProjectForm.*` — no gap there, no change made.

## Testing

- `nx run vre-shared-app-help-page:test` — 10 tests, replacing the library's placeholder spec
- `nx run-many --all --target=lint --exclude=vre-open-api` — 28 projects pass
- Key parity against `en.json` — no missing, no extra keys in `de`/`fr`/`it`

**The tests assert in German, not English, and this matters.** My first version asserted against `en.json` and was worthless: reverting a template to a hardcoded English literal still satisfied `toContain(en.…)`, because the English translation *is* that same string. I found this by mutation testing. Asserting in German means a hardcoded English literal cannot pass — verified by reintroducing one and confirming exactly 2 tests fail. Two footer strings ("Impressum", "Downloads") are identical in German and English and so remain untestable this way.

The specs read the translation files from disk via a small helper rather than importing them. Importing `apps/dsp-app/src/assets/i18n/*.json` from a lib spec creates a `lib -> app -> lib` cycle that `@nx/enforce-module-boundaries` rejects — it failed `vre-pages-user-settings-user:lint` until I changed the approach.

## Not verified

I could not run a full `nx run dsp-app:build`: the prerequisite `vre-open-api:generate` fails with `ERR_REQUIRE_ESM` in the openapi-generator CLI. I confirmed by stashing that it fails identically on clean `main`, so it is pre-existing and unrelated to this branch. Template compilation is instead covered by the new specs, which mount both components for real against the actual translation files.

## For the reviewer

The footer's **"Privacy Policy (English)"** link now sits among translated siblings and reads a little oddly — it points at a separate English-only document. Left as-is deliberately, since whether to keep, rename, or drop that link is a content decision.


domsteinbach:
@claude review

mpro7 (approved):

PR #3371, opened 2026-08-25
fix(video): surface initialisation state instead of an empty media box (DEV-7026)

mpro7:
[DEV-7026](https://linear.app/dasch/issue/DEV-7026)

## Summary

- A video resource showed neither a player nor any feedback while it initialised. The progress indicator sat in an `@else` branch on `video`, which is assigned **synchronously** in `ngOnChanges` before the first render — so on every path where that assignment succeeds, the branch was unreachable and no indicator appeared.
- The indicator is now overlaid on the video area while the player is not ready and no error has been reported. The `<video>` element stays mounted underneath; unmounting it would make the browser abort the load.
- Player readiness moves from `canplay` to `loadedmetadata`, matching `AudioComponent`. `onVideoPlayerReady` only needs the duration and the ability to seek, both available at `HAVE_METADATA`; `canplay` additionally waits for a playable buffer, which delays the whole player chrome (slider + toolbar) on a large file.

## Relationship to DEV-7016 — please read before reviewing

[DEV-7016](https://linear.app/dasch/issue/DEV-7016) (endless spinner on restricted media) is **not fixed here**, and its root-cause text points at the `@else` branch this PR removes. Verified against the patched template with `src === null`, exactly as the wrappers pass it when `getFileValue()` returns `null`:

```
threw: "Cannot read properties of null (reading 'fileUrl')"
video: null   isPlayerReady: false   videoError: ""
loaderPresent: true   videoElementPresent: false   errorMessagePresent: false
```

`ngOnChanges` throws on `this.src.fileUrl` *after* setting `isPlayerReady = false` and `videoError = ''`, so the new condition `!isPlayerReady && !videoError` is true and the overlay renders. **The endless spinner is unchanged in behaviour — it just moves** from `@else` on `video` to `@if (!isPlayerReady && !videoError)`.

Two notes for whoever picks up DEV-7016:

1. Its "Root cause" section will be stale once this merges. The line to reason about becomes `@if (!isPlayerReady && !videoError)` in `video.component.html`.
2. Its proposed fix — guarding at the wrapper with `@if (fileValue)` — composes cleanly: `<app-video>` never mounts without a file value, so this overlay never renders in that path. No guard was added here on purpose, to avoid handling the same condition twice in two layers.

Also flagging: the `onDark` stroke below takes that spinner from 2.41:1 to 12.45:1, so the DEV-7016 symptom becomes **more visible** until it is fixed. Cosmetic, on an already-broken state.

## Changes

**Video representation** — `video.component.{html,ts}`
- Loading overlay gated on `!isPlayerReady && !videoError`, so a failed video shows the error state rather than spinning forever.
- Readiness bound to `(loadedmetadata)` instead of `(canplay)`.
- `preload="auto"` deliberately left unchanged (see Follow-ups).

**Shared progress indicator** — `app-progress-indicator.component.ts`
- New opt-in `onDark` input. The default `#336790` stroke measures **2.41:1** against the media container's `rgb(41, 41, 41)`, under the 3:1 WCAG 1.4.11 minimum for non-text contrast — a spinner nobody can see would have defeated the fix. `onDark` switches to `#e8eef4` (**12.45:1**).
- Default unchanged, so the 25+ existing call sites are unaffected. Grepped for any CSS selector or assertion matching the stroke value — none.
- Replaced the self-describing coverage placeholder spec with real tests for the rendered output.

## Test Plan

`nx run-many --target=lint` and `--target=test` on `vre-resource-editor-resource-editor` + `vre-ui-progress-indicator`, plus `nx run dsp-app:build:dev-server` for Angular's strict template checking — all green (411 tests).

**Fix-confirmation** — reverting only the template and re-running fails exactly the two new tests, for the right reasons:
- `shows a loading indicator while the video is still initialising` → no reachable loader existed
- `hides the loading indicator once the video reports its metadata` → `loadedmetadata` was not bound

The other two tests in the new suite (error state instead of a perpetual spinner; `<video>` receives the real URL through the real `DomSanitizer`) pass before and after — they are regression guards, and the second is the video-side counterpart to the DEV-7010 audio guard.

**Verified against the stage asset used by the stories** (`0869/3xUzuLcE9nC…mp4`):
- 368 MB and **faststart** (`ftyp`, then `moov` at offset 32, ~596 KB), so metadata lands within the first 600 KB — `loadedmetadata` is genuinely early, which is what makes the gate change worth making.
- The `knora.json` sidecar returns in ~50 ms, so `fileInfo` was never the bottleneck.

**Not reproduced locally:** stage serves at ~32 MB/s from here, so the multi-second empty box could not be triggered. The empty-box *symptom* is connection-bound; the unreachable indicator branch was a defect regardless of connection speed. Worth a check on a throttled connection during review.

**Not verified locally:** the new `OnDarkSurface` story's `play()`. `playwright-core` pins chromium build 1234 and only 1208 is cached, and installing writes outside the repo — deferred to CI, which has the browsers. Its assertion (the rendered `stroke` attribute) is covered deterministically by the new Jest spec, so only the harness wiring is unproven, and `componentWrapperDecorator` follows the existing `permission-info` story precedent.

## Follow-ups (deliberately out of scope)

- **Endless spinner on restricted media** — [DEV-7016](https://linear.app/dasch/issue/DEV-7016), see above.
- **A stalled download still spins indefinitely.** No timeout was added: a stall is not a failure, the browser can resume, and a timeout would raise false errors on slow-but-working connections. Genuine failures fire `error` and already reach the error state.
- **No `role="status"` on the loading overlay**, so screen readers get no announcement. The right home is `AppProgressIndicatorComponent` itself, which would benefit all 25+ call sites — a broader a11y change than this fix should carry.
- **`preload="auto"` on a 368 MB file** tells the browser to fetch the whole asset in the background. `preload="metadata"` would cut that dramatically and pairs naturally with the `loadedmetadata` gate, but it can change first-frame display and initial playback smoothness, and the issue did not ask for it.

## Screenshots

### DIFF
--- a/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
+++ b/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
@@ -4,7 +4,7 @@
     <app-representation-error-message />
     }
 
-    <div style="margin-bottom: -4px">
+    <div class="video-stage" [class.is-loading]="!isPlayerReady && !videoError">
       @if (video) {
       <video
         style="width: 100%"
@@ -14,11 +14,16 @@
         type="video/mp4"
         preload="auto"
         [muted]="false"
-        (canplay)="onVideoPlayerReady()"
+        (loadedmetadata)="onVideoPlayerReady()"
         (click)="videoPlayer.togglePlay()"
         (error)="handleVideoError($event)"></video>
-      } @else {
-      <app-progress-indicator />
+      }
+      <!-- Overlaid rather than swapped in: the <video> element must stay mounted while it
+           initialises, otherwise the browser stops loading it. See DEV-7026. -->
+      @if (!isPlayerReady && !videoError) {
+      <div class="video-loading" data-cy="video-loading">
+        <app-progress-indicator size="medium" [compact]="true" [onDark]="true" />
+      </div>
       } @if (false) {
       <app-video-overlay />
       }
--- a/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
+++ b/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
@@ -4,7 +4,7 @@
     <app-representation-error-message />
     }
 
-    <div class="video-stage" [class.is-loading]="!isPlayerReady && !videoError">
+    <div class="video-stage" [class.is-loading]="!isMetadataLoaded && !videoError">
       @if (video) {
       <video
         style="width: 100%"
@@ -14,13 +14,14 @@
         type="video/mp4"
         preload="auto"
         [muted]="false"
-        (loadedmetadata)="onVideoPlayerReady()"
+        (loadedmetadata)="onMetadataLoaded()"
+        (canplay)="onVideoPlayerReady()"
         (click)="videoPlayer.togglePlay()"
         (error)="handleVideoError($event)"></video>
       }
       <!-- Overlaid rather than swapped in: the <video> element must stay mounted while it
            initialises, otherwise the browser stops loading it. See DEV-7026. -->
-      @if (!isPlayerReady && !videoError) {
+      @if (!isMetadataLoaded && !videoError) {
       <div class="video-loading" data-cy="video-loading">
         <app-progress-indicator size="medium" [compact]="true" [onDark]="true" />
       </div>
@@ -29,7 +30,7 @@
       }
     </div>
 
-    @if (isPlayerReady) {
+    @if (isMetadataLoaded) {
     <app-media-slider
       [max]="videoPlayer.duration()"
       [currentTime]="myCurrentTime"
--- a/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.spec.ts
+++ b/libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.spec.ts
@@ -170,3 +170,72 @@ describe('VideoComponent — behavior', () => {
     });
   });
 });
+
+describe('VideoComponent — loading and ready state (real template)', () => {
+  let fixture: ComponentFixture<VideoComponent>;
+  let component: VideoComponent;
+
+  const loader = () => fixture.nativeElement.querySelector('[data-cy="video-loading"]');
+  const videoEl = (): HTMLVideoElement => fixture.nativeElement.querySelector('video');
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [VideoComponent],
+      providers: [provideTranslateService()],
+    })
+      .overrideComponent(VideoComponent, {
+        set: {
+          providers: [
+            {
+              provide: SegmentsService,
+              useValue: { onInit: jest.fn(), setSegments: jest.fn(), segments: [], playSegment$: EMPTY },
+            },
+            {
+              provide: MediaControlService,
+              useValue: { play$: EMPTY, watchForPause$: EMPTY, playMedia: jest.fn(), mediaDurationSecs: undefined },
+            },
+            {
+              provide: MediaPlayerService,
+              useValue: { onInit: jest.fn(), navigate: jest.fn(), duration: () => 788, onTimeUpdate$: EMPTY },
+            },
+            // Leaves fileInfo undefined, so the toolbar stays out of this fixture.
+            { provide: RepresentationService, useValue: { getFileInfo: jest.fn().mockReturnValue(EMPTY) } },
+            { provide: NotificationService, useValue: { openSnackBar: jest.fn() } },
+          ],
+        },
+      })
+      .compileComponents();
+
+    fixture = TestBed.createComponent(VideoComponent);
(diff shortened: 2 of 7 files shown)


mpro7:
@claude review

libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
```diff
@@ -14,11 +14,16 @@
         type="video/mp4"
         preload="auto"
         [muted]="false"
-        (canplay)="onVideoPlayerReady()"
+        (loadedmetadata)="onVideoPlayerReady()"
```
domsteinbach: The signal loadedmetadata fires before canPlay, so that changes everything downstream which relies on canPlay. If loadedmetadata is the right hook for the loading spinner pls use that signal separately to show/hide the spinner via a separate onMetaDataLoaded(). loadedmetadata is probably not the correct hook for onVideoplayerReady() and all depending on it.

domsteinbach (changes requested):

libs/vre/resource-editor/resource-editor/src/lib/resource/video/video.component.html
domsteinbach: If other components should rely on loadedmetadata (like the video toolbar) they should also listen to that via a method onMetaDataLoaded or so. If nothing needs canplay everything can listen to that. But loadedmetadata triggering onVideoPlayerReady() is wrong to me.

domsteinbach: loadedmetadata => duration, slider max, seek target, layout, stop showing an empty box
 canplay  =>  play button enabled, autoplay

mpro7: Agreed on the split, the third comment is the right taxonomy and I've implemented it in a04db27c2. Two corrections to the premise of the first comment though, because I think the blast radius is much smaller than it looks.

**What changed**

`isPlayerReady` was doing two jobs behind one name. Split along the readyState boundary:

- `onMetadataLoaded()` — `loadedmetadata` / HAVE_METADATA. Binds the player service, seeks to `start`, publishes the duration, ends the loading state. **The slider moves here**: it needs the duration and seeking, both guaranteed at this readyState, and shouldn't wait on a buffer.
- `onVideoPlayerReady()` — `canplay` / HAVE_FUTURE_DATA. Now only sets `isPlayerReady` and emits `loaded`. **The toolbar stays here**, exactly per your taxonomy, it carries the play button, which shouldn't appear before pressing it starts playback.

So: `loadedmetadata` → duration, slider max, seek target, stop showing an empty box. `canplay` → transport controls. Your line, implemented.

**Correction 1: nothing downstream relied on `canplay`.** I checked before changing it. `onVideoPlayerReady` did four things — bind the service, seek to `start`, read `duration`, subscribe to `timeupdate` — and all four are valid at HAVE_METADATA (the HTML spec guarantees `duration` is set and the seek algorithm is permitted; only HAVE_NOTHING defers). The `(loaded)` output is **bound nowhere in the repo**, so emitting it earlier had literally no effect. Nothing anywhere reads buffering state. The real surface was three template gates, not "everything downstream".

**Correction 2: `AudioComponent` already ships the pattern you're rejecting.** `audio.component.html` binds `(loadedmetadata)="onAudioPlayerReady()"`, and `app-audio-toolbar` has the same play/pause button behind the same single `isPlayerReady` flag. That's where I took the pattern from, it's the existing precedent, not something I invented here. Worth noting because **video and audio now diverge**: video distinguishes the two states, audio still conflates them. I'd rather not fold audio into this PR (DEV-7010 only just landed there), so either it gets the same split in a follow-up or we accept the inconsistency, please let me know.

**One place I'd still push back, and you may want to overrule me.** Native `<video controls>` enables its whole control bar at `loadedmetadata`, precisely because `play()` is legal there — it sets `paused = false`, the browser buffers, then fires `waiting`/`playing`. So gating the toolbar on `canplay` trades a truthful play button for a late-arriving toolbar, i.e. a layout shift on slow connections. I've gone with your call since it's defensible and the shift is minor, but I don't think it's free.

**And a gap neither of us covered:** once the overlay clears at metadata, pressing play on a large file gives no feedback at all until enough data buffers — the "silently empty" problem returns at a different moment. Under your gating that window is smaller but still there (seek into an unbuffered region does the same thing). The actual fix is a buffering indicator on `waiting`/`playing`, which is what YouTube et al. do while keeping play enabled. That's the same follow-up as the stalled-download item already listed in the description — say the word and I'll file it.

domsteinbach: > Correction 1: nothing downstream relied on canplay. ...

> And a gap neither of us covered: once the overlay clears at metadata, pressing play on a large file gives no feedback at all until enough data buffers ...

That was obvious to me, I did not mention it. Imho that is a downstream issue when we don't listen to canplay.

domsteinbach (approved):

PR #3372, opened 2026-08-26
chore: slim CLAUDE.md to non-derivable guidance and repair Claude skill definitions

mpro7:
## What

Three changes to the Claude Code configuration in this repo, all documentation and agent config — no application code is touched.

1. **`fix`** — `.claude/skills/format-project-image/SKILL.md` had no `---` frontmatter delimiters, so the loader discarded every field. The skill's name fell back to its directory name and its description to the first body line (it showed up in the skill listing as `format-project-image: format-project-image`), and the declared `allowed-tools: Bash, Read, Edit, Glob` restriction was never applied. The metadata is now real frontmatter, with the author's `invocable: user` intent carried over as `disable-model-invocation: true` — the skill is reachable through `/format-project-image` only.
2. **`feat`** — the OpenAPI Code Generation section moved out of `CLAUDE.md` into a new `.claude/skills/update-openapi-client/` skill, so the command list and the cross-repo deploy sequence load on invocation instead of in every session. The *prohibition* ("never edit the vendored spec to silence a red `check-openapi-sync`") deliberately stays resident in `CLAUDE.md`: the moment it matters is when a session sees a red check, and that session has no reason to open an OpenAPI skill before reaching for the wrong fix.
3. **`docs`** — trimmed ~11k characters of codebase-derivable content from `CLAUDE.md`.

## Why

Every line of the root `CLAUDE.md` is loaded into context in every session in this repo. Anything a session can reconstruct by reading the repo is paying rent for nothing. The file went from **414 to 135 lines, ~5.0k to ~2.0k est. tokens per session**.

**Cut** — derivable or mechanically enforced: directory layouts (`ls` shows them); dependency and tooling lists that `package.json` already carries; `npm`/`nx` invocations that are standard or already scripted; TypeScript and NX settings from `tsconfig.base.json`; the DSP-JS API signatures and snippets copied out of `libs/dsp-js/`; and the style rules `eslint.config.mjs` / `eslint.config.angular.mjs` already enforce (import ordering, focus-test ban, control-flow syntax). The Environment Configurations list was also already stale — it named five configurations where `apps/dsp-app/project.json` defines six.

**Kept** — everything the codebase cannot teach: the permission and non-repo filesystem safety rules, layer ownership, the Storybook conventions including the container-DI gotcha, the Romansh fallback (DEV-6629), the JSON-LD gotchas, the dual-client rule, and `::ng-deep` / self-closing tags — neither is lint-enforced here, and existing components violate the `::ng-deep` rule, so the code alone would teach the wrong pattern.

Explicitly *not* a reason for any cut: overlap with a maintainer's private machine-level `CLAUDE.md` files. Those are untracked and invisible to anyone cloning this repo, so every cut here stands on its own.

## Verification

- Every one of the 49 section boundaries was asserted against its expected text before any line was removed; the first run aborted on a mismatched anchor and only proceeded once all anchors matched.
- Both `SKILL.md` files parse with `name` and `description` present; `/update-openapi-client` registered and listed immediately in a live session.
- No references to the removed sections exist anywhere else in the repo (only `CHANGELOG.md` history mentions the old "Claude Code Operating Rules" rewrite).
- No markdown linting runs in CI, and no CI job depends on `CLAUDE.md` content.

One thing for reviewers to be aware of: `.github/workflows/claude.yml` points its PR reviewer at this repo's `CLAUDE.md` for "style, conventions, and repository-specific best practices". That context is now smaller by design — the conventions were kept, the derivable facts were removed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,6 +5,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 ## Claude Code Operating Rules
 
 ### 1. Permission & Safety
+
 - The assistant may freely **propose** changes, but must obtain **explicit, unambiguous user approval** before:
   - Modifying code or files  
   - Creating, deleting, modifying or moving files/directories  
@@ -17,6 +18,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - If approval is unclear, the assistant must ask for clarification.
 
 ### 2. Non-Repository File System Safety
+
 - The assistant must **not** modify, create, move, or delete files or directories **outside the project repository** unless explicitly instructed by the user.
 - If the assistant detects that a requested operation would affect the broader system (e.g., user home directory, OS configuration, global environment files, unrelated projects), it must:
   - Clearly **highlight** this fact to the user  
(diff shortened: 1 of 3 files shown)


CLAUDE.md
```diff
@@ -17,66 +18,35 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
...
 - If repository conventions conflict with best practices, ask which to prioritize.  
 - If user instructions conflict with conventions or principles, seek clarification.
 
-### 6. Design Principles
-- Use established principles (SOLID, DRY, KISS, YAGNI) when appropriate to the task.  
-- Avoid introducing complexity or over-engineering.  
-- If applying a principle requires major structural changes, propose the idea first and wait for explicit approval.  
-- When a proposal is influenced by a principle, state which one for transparency.
-
 ### 7. Testing Guidelines
```
domsteinbach: renumbering from here?

domsteinbach (approved):
check my comment pls

CLAUDE.md
mpro7: That's the last paragraph on this level, but indeed it should be updated. Nice catch, thanks!

PR #3383, opened 2026-08-28
fix(representation): show a no-access message instead of an endless spinner for restricted media (DEV-7016)

derschnee68:
Fixes [DEV-7016](https://linear.app/dasch/issue/DEV-7016).

## Problem

When a time-based media resource is viewable but its file value is withheld (`userHasPermission: "V"`, `hasMovingImageFileValue`/`hasAudioFileValue` absent), the media area rendered a black box with an endlessly rotating spinner instead of telling the user they have no access. Reproduced on STAGE while logged out.

`getFileValue()` returns `null` in that case, but all four AV wrappers asserted non-null with `!` and passed the `null` on:

- `<app-resource-legal>` reads `.copyrightHolder` / `.license` on `null` → the `TypeError: Cannot read properties of null` spam on every change-detection cycle.
- `<app-video>` / `<app-audio>` throw inside `bypassSecurityTrustUrl(null.fileUrl)` before the player can become ready, so the progress indicator never gets cleared — the `(error)` handler only fires for a real `<video>` element error.

The existing `<app-resource-restriction>` banner does not cover this: it is gated on `userHasPermission === 'RV'`, and here the resource is `V` with only the file value hidden.

## Change

Treat "no accessible file value" as a first-class state in the AV wrappers:

- Drop the `!` from `getFileValue(...)` so the nullability is visible to the template.
- Guard the legal panel and the player with `@if (fileValue; as file)`.
- Render a new media-neutral `<app-representation-restricted>` in the `@else`, reusing the existing yellow `app-alert-info`. One component serves video, audio and both segment variants.
- New i18n key `resourceEditor.representations.noPermission` in en/de/fr/it.

Applied to all four wrappers: `resource-video`, `resource-video-segment`, `resource-audio`, `resource-audio-segment`.

## Verification

- `nx test vre-resource-editor-resource-editor` — 48 suites, 405 passed, 1 skipped.
- Angular compiler (`ngc --noEmit`, `strictTemplates: true`) on the lib — no new diagnostics; the only remaining errors are pre-existing `TS7016` missing-declaration notes for `jsonld` and `ckeditor5-custom-build`.
- ESLint clean on all changed files; Prettier clean.
- i18n key parity verified across de/fr/it against `en.json`.
- Regression story per wrapper asserting the message renders while neither the player nor a spinner does.

## Notes for review

- **de/fr/it wording is my own, not PM-supplied** — worth a look from whoever owns the translations.
- Composes with #3371 (DEV-7026) as predicted in the issue thread: guarding at the wrapper means `<app-video>`/`<app-audio>` never mount without a file value, so that PR's relocated overlay never renders in this path. No change needed on either side.
- **Out of scope, same latent bug:** `resource-archive`, `resource-pdf`, `resource-document`, `resource-text` and `resource-image` use the identical `getFileValue(...)!` pattern. They do not spin forever, but they will throw on the legal panel under the same conditions. Happy to file a follow-up.

### DIFF
--- a/apps/dsp-app/src/assets/i18n/de.json
+++ b/apps/dsp-app/src/assets/i18n/de.json
@@ -624,6 +624,7 @@
       "text": {
         "title": "Text (csv, txt, xml)"
       },
+      "noPermission": "Sie haben keine Berechtigung, diese Mediendatei anzuzeigen.",
       "placeholder": {
         "title": "Platzhalter – Datei noch nicht verfügbar",
         "message": "Diese Ressource verweist auf einen Platzhalter. Die eigentliche Datei wurde noch nicht geliefert."
--- a/apps/dsp-app/src/assets/i18n/en.json
+++ b/apps/dsp-app/src/assets/i18n/en.json
@@ -641,6 +641,7 @@
       "text": {
         "title": "Text (csv, txt, xml)"
       },
+      "noPermission": "You do not have permission to view this media file.",
       "placeholder": {
         "title": "Placeholder – asset not yet available",
         "message": "This resource references a placeholder. The actual file has not been delivered yet."
--- a/apps/dsp-app/src/assets/i18n/fr.json
+++ b/apps/dsp-app/src/assets/i18n/fr.json
@@ -641,6 +641,7 @@
       "text": {
         "title": "Texte (csv, txt, xml)"
       },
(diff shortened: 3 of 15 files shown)


derschnee68:
@claude review

domsteinbach (approved):
The reviewers first issue is preexisting. Another bug. The PR solves the issue.

PR #3391, opened 2026-09-02
fix: translate the "Found in" property names on search results (DEV-5452)

derschnee68:
Solves DEV-5452.

## Problem

On a fulltext search result the resource class already follows the app language, but the names listed after **Found in:** do not.

Reproduced on https://app.dev.dasch.swiss/search/Brief with the app in French — the first hit reads:

> Portée | Trouvé dans : **Realization** | Project: posepi

while the resource viewer on the right shows that very same property as **Réalisation**.

## Cause

`foundIn` was built from `ReadValue.propertyLabel` — the single-language string dsp-api resolves server-side — instead of the multi-language `entityInfo.properties[...].labels` that the resource class label already uses. The `Label` entry (a hit on the resource's own label) was a hardcoded English string.

## Change

- `foundIn` now holds the multi-language label arrays (or a translation key for the resource's own label), deduplicated by property IRI rather than by the resolved name.
- Rendered through `foundInText$` — `combineLatest(currentLanguage$, translate.stream(...))` — so the names re-resolve when the UI language changes, like the class label does.
- Falls back to `propertyLabel` when `entityInfo` carries no labels for the property.
- New translation key `pages.dataBrowser.resourceListItem.resourceLabel` in en/de/fr/it.

If a project's ontology has no label in the chosen language, the existing fallback chain (`en > de > fr > it > rm`) still applies — that part is a data limitation, not something the app can resolve.

## Testing

Unit tests pass (31 in `resource-list-item.component.spec.ts`), including a new case asserting the French property name wins when the UI language is French. Not verified in a running app yet — the storybook stories for this component are worth a look too.

### DIFF
--- a/apps/dsp-app/src/assets/i18n/de.json
+++ b/apps/dsp-app/src/assets/i18n/de.json
@@ -500,7 +500,8 @@
         "noData": "Es gibt keine Daten, die Ihrer Anfrage entsprechen."
       },
       "resourceListItem": {
-        "foundIn": "Gefunden in: "
+        "foundIn": "Gefunden in: ",
+        "resourceLabel": "Label"
       },
       "downloadDialog": {
         "title": "Herunterladen",
--- a/apps/dsp-app/src/assets/i18n/en.json
+++ b/apps/dsp-app/src/assets/i18n/en.json
@@ -500,7 +500,8 @@
         "noData": "There are no data corresponding to your request."
       },
       "resourceListItem": {
-        "foundIn": "Found in: "
+        "foundIn": "Found in: ",
+        "resourceLabel": "Label"
       },
       "downloadDialog": {
         "title": "Download",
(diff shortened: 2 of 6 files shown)


derschnee68:
@claude review

domsteinbach (changes requested):
I would really like to have things streamlined. c) resource-list-item.component.ts:206-218 — a simpler path already exists in the repo — Non-blocking, impact Low

Has quite some impact imho. translations should all be streamlined. Can you change that?

FYI: I created an issue (assigned to me) concerning the typing and foremost the DSP JS is lying about labels in general. DSP-JS states that labels/label are optional. The base data model and all create and update routes, all ingest have it as required and always existing. I will check that. That will enable us to clean up and shorten a lot downstream in the app. Also the type checks and different types ResourcePropertydefinition and ResourcePropertyDefinitionWithAllLanguages is a pain in the pipeline.

domsteinbach (approved):

PR #3396, opened 2026-09-04
fix(ontology): load project ontologies and lists once per editor visit (DEV-7130)

domsteinbach:
## Problem

The ontology editor's derived streams were built on the cold `ProjectPageService.ontologies$` and a cold lists request. Neither is shared, so both re-issued their full request set on every re-subscription — each class card and property menu triggered its own `getOntologiesByProjectIri` plus one `getOntology` per project ontology. Opening a data model with several ontologies and many classes produced a burst of redundant `allentities` requests.

## Change

`OntologyEditService` now owns this state instead of reading it through a cold stream:

- Project ontologies are held in `_projectOntologiesSubject`, seeded once in the constructor. The service's lifetime is one visit to `OntologyPageComponent`, so that's one request set per visit.
- Lists are held in `_listsInProjectSubject` (a `ReplaySubject(1)`), likewise loaded once.
- The derived streams (`currentOntologyProperties$`, `currentProjectsProperties$`) read from these subjects.
- Mutations keep the snapshot current: `_afterUpdateOntology` upserts via `_upsertProjectOntology`, and `deleteOntology$` removes via `_removeProjectOntology`. Every subscriber still sees each change, without refetching.

The `reloadProject()` calls in the create-property and create-resource-class dialogs became dead weight — the editor no longer reads that stream, so the project GET they triggered was discarded. Removed them along with the `ProjectPageService` injections that existed only to serve them.

## Notes

The editor holds a snapshot for the duration of a visit, so a sibling ontology changed concurrently by another user is picked up on re-entry rather than mid-visit. That was already true — there is no push or invalidation mechanism; the cold streams merely refetched often enough to hide it. Getting both fresh data and few requests needs an actual invalidation mechanism, which is out of scope here.

Follow-up (not in this PR): the remaining direct consumers of `ProjectPageService.ontologies$` — `gui-attr-link`, `create-ontology-form-dialog`, `data-class-view` — still fire a full request set per subscription. That's the broader "where should project-ontology state live" refactor.

## Testing

- `nx run dsp-app:lint` — passes
- `nx run vre-pages-ontology-ontology:test` — passes
- `tsc-errors` output is byte-identical with and without this change (diffed against a stash), so no new type errors and no orphaned symbols

### DIFF
--- a/libs/vre/pages/ontology/ontology/src/lib/forms/property-form/edit-property-form-dialog.component.ts
+++ b/libs/vre/pages/ontology/ontology/src/lib/forms/property-form/edit-property-form-dialog.component.ts
@@ -8,7 +8,6 @@ import {
   MatDialogRef,
 } from '@angular/material/dialog';
 import { StringLiteralV2 } from '@dasch-swiss/vre/3rd-party-services/open-api';
-import { ProjectPageService } from '@dasch-swiss/vre/pages/project/project';
 import { LoadingButtonDirective } from '@dasch-swiss/vre/ui/progress-indicator';
 import { DialogHeaderComponent } from '@dasch-swiss/vre/ui/ui';
 import { TranslatePipe, TranslateService } from '@ngx-translate/core';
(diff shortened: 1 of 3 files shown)


domsteinbach:
@claude review

derschnee68 (approved):

PR #3412, opened 2026-09-08
fix(styles): load the icon font once globally instead of per component (#3411)

### LINKED ISSUE #3411: Move icon-font loading out of _config.scss so it stops leaking into 33 component stylesheets
## Problem

`apps/dsp-app/src/styles/_config.scss` line 2:

```scss
@import url("https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined");
```

`_config.scss` is a **variables partial** — 33 component stylesheets `@use` it to get colour and palette variables. But a CSS `@import` is not scoped like Sass variables: it is inlined into every stylesheet that pulls the partial in. So each of those 33 components ends up carrying its own encapsulated copy of the remote sheet's rules:

```css
.material-icons[_ngcontent-ng-c417689689] { font-size: 24px; ... }
```

At specificity **(0,2,0)** — class + encapsulation attribute — that copy ties Angular Material's own `.mat-mdc-button > .mat-icon` (also 0,2,0) and **wins on source order**, because component styles are injected after the global stylesheet.

## Consequence

Any component that both `@use`s the config partial and puts a `mat-icon` inside a `mat-button` gets a 24px glyph in the 18px-wide `overflow: hidden` box Material sizes button icons to — a clipped icon.

Confirmed on stage (clean page load):

| page | icons | measured |
|---|---|---|
| resource / incoming-links pager | `west`, `east` | 24px in 18×24 box, `18/24` → clipped |
| /help | `launch` ×7, `mail` ×2 | 24px in 18×24 box, `18/24` → clipped |

Icons in buttons whose component does **not** `@use` the partial (e.g. `language`, `unfold_more`) measure 18px/18×18 and are fine — which isolates the import as the trigger.

This is a recurring defect, not a one-off: DEV-7011 was patched per-component in #3410, and the same latent problem sits in `property-info` (`edit`, `delete`) and `sort-button` (`sort`).

It also caused the outage in #3384 → #3406 → #3407: those PRs correctly identified the per-component inlining but tried to fix it in the global sheet, and ended up removing the app's only icon `font-family`, so no icons rendered anywhere on dev and stage (reverted in #3408).

## Proposed fix

Move the icon-font loading out of the variables partial into a stylesheet that is included exactly once, globally.

- The two `@font-face` declarations are harmless duplicated — they declare a font, they do not match elements.
- The two classes (`.material-icons`, `.material-icons-outlined`) are what leak and must exist only once, in the global sheet.

`.material-icons` must keep `font-family` and the ligature setting: `MatIconRegistry` applies `class="material-icons mat-ligature-font"` to every `<mat-icon>` at runtime, and Angular Material never sets the icon `font-family` itself. Removing that class is what broke #3407 — see #3408 for the analysis.

## Acceptance criteria

- [ ] `_config.scss` contains no `@import` (variables only)
- [ ] Icon-font `@font-face` and `.material-icons*` classes are declared once, in the global stylesheet
- [ ] Built output contains exactly one `.material-icons` rule and **no** `.material-icons[_ngcontent-*]` copies
- [ ] Icons render everywhere (regression guard for #3407)
- [ ] Button icons measure 18px/18×18 without per-component `width: 1em` patches
- [ ] The patches added in #3410 can be reverted
- [ ] Verified in a **production** build (`nx build dsp-app --configuration=production`) — the per-component inlining only appears in an optimized build

## Notes

- Requires Node from `.nvmrc` (22.23.1). On Node 20 the build fails with `ERR_REQUIRE_ESM`, which is why the earlier PRs in this chain shipped unverified.
- Worth deciding as part of this: whether to keep loading the icon font from Google Fonts or self-host it. `apps/dsp-app/src/assets/fonts/material-icons/MaterialIcons-Regular.ttf` (356 KB, uncompressed) is in the repo but **never requested** — with `optimization.fonts.inline: true` the build resolves the Google `@import` at build time and inlines a `@font-face` pointing at `fonts.gstatic.com`. Either self-host properly (convert to `.woff2`, ~80 KB) or delete the unused `.ttf`.
- `Material Icons Outlined` has no usages in the codebase; the second family may be droppable.

Refs DEV-7011. Follows #3410, #3408.

domsteinbach:
Closes #3411.

## Problem

`_config.scss` is a **variables partial** that 33 component stylesheets `@use`. Its `@import` of the Google Fonts icon stylesheet was not scoped like a Sass variable — the CSS was inlined into every one of those components, so each carried its own encapsulated copy:

```css
.material-icons[_ngcontent-ng-c417689689] { font-size: 24px; ... }
```

At **(0,2,0)** that copy ties Angular Material's `.mat-mdc-button > .mat-icon` and **wins on source order**, since component styles are injected after the global sheet. Any component that both used the partial and put a `mat-icon` in a `mat-button` rendered a 24px glyph inside the 18px `overflow: hidden` box Material clips button icons to — cropped icon (DEV-7011).

## Change

Declare the font and its classes once, in `_font.scss`, which `styles.scss` is alone in `@use`ing.

## Verification (production build, Node 22.23.1)

| | before | after |
|---|---|---|
| `.material-icons` rules in global sheet | 1 + one per component | **1** |
| `.material-icons[_ngcontent-*]` copies in chunks | one per component | **0** |
| pager arrows (real component) | 24px in 18×24 box, `18/24` clipped | **18px in 18×18, `18/18`, unclipped** |
| `font-family: Material Icons` present | yes | **yes** (regression guard for #3407) |

Measured in the running component, not inferred: both arrows now get Material's sizing uncontested, and the icon-font `@font-face` is intact so glyphs load.

## Deliberate details

- **No `font-size` in the class.** That is what lets Material's contextual rules win on their own — 18px inside buttons, 24px standalone. Adding it back is what caused the shrink in #3406.
- **`.material-icons` keeps `font-family` and `font-feature-settings: 'liga'`.** `MatIconRegistry` adds `class="material-icons mat-ligature-font"` to every `<mat-icon>` at runtime and Material never sets the icon font-family itself. Deleting this class is what left the app with no icons in #3407 (reverted in #3408).
- **Switched to the `.woff2`** the Google sheet now serves, instead of the `.ttf` it used to.
- **Dropped `Material Icons Outlined`** — no usages anywhere in the codebase.
- **Removed `MaterialIcons-Regular.ttf` (356 KB).** It was never requested: with `optimization.fonts.inline: true` the build resolved the remote `@import` and inlined a `gstatic` `@font-face`, so the local file was dead weight. Confirmed no remaining references.

## Follow-up

With this merged, the per-component `width: 1em` patches in #3410 are no longer needed and can be reverted. Suggest merging #3410 first if you want the clipping fixed immediately (it is the smaller, lower-risk change), then this, then a small revert of #3410. Or merge this alone and close #3410 — it fixes the same user-visible bug at the root.

Also worth a decision, out of scope here: the `@font-face` now pins a versioned `fonts.gstatic.com` URL (`materialicons/v145`). That is stable and immutable, which is good for caching, but it does mean the URL must be refreshed manually to pick up a newer icon set — and it keeps the runtime dependency on Google. If self-hosting is wanted (offline support, CSP, no third-party requests), that is a clean follow-up: drop a `.woff2` into assets and point the `@font-face` at it. I did not bundle that here since it changes the loading story again and #3384 showed that deserves its own verification.

## Testing

- [x] Production build succeeds (`nx build dsp-app --configuration=production`)
- [x] Zero `.material-icons[_ngcontent-*]` copies in the built chunks
- [x] Exactly one `.material-icons` rule in the global sheet, with `font-family` + `liga`, no `font-size`
- [x] Pager arrows measured in the real component: 18px/18×18, unclipped
- [x] Icon `@font-face` present so glyphs load (guards #3407)
- [x] `.action-bubble` 18px override still correctly scoped
- [ ] Confirm on dev after deploy: icons render everywhere, pager arrows and /help icons uncropped

Pre-existing Prettier warnings on both touched files are left alone to keep the diff to the fix (verified they warn on `main` too).


domsteinbach:
Pushed `955ab169c`: restored `font-size: 24px` on `.material-icons`.

You were right that this was already tried — the first version of this PR was **byte-identical** to #3384's class (I diffed it), so it reproduced #3384's regression: undersized icons. #3406's commit message had already recorded the exact reason:

> `font-size` must stay: Material's base `.mat-icon` rule sets only a 24px box, no font-size, so a standalone `<mat-icon>` would otherwise inherit the body font-size and render undersized.

I should have read the chain I'd spent the session reverting before rewriting the same fix.

One clarification on which failure was which: the "no icons at all" outcome was **#3407** (it deleted the class, removing the only `font-family`). This PR keeps `font-family` and `liga`, so it rendered icons — just small. Three distinct failures in that chain: #3384 undersized → #3406 sized but clipping → #3407 nothing.

## Why 24px is safe here, unlike #3406

#3406 kept the per-component copies in play. At **(0,2,0)** each copy tied Material's `.mat-mdc-button > .mat-icon` and won on **source order**, so the 24px leaked into buttons and clipped.

Declared once in the global sheet it is only **(0,1,0)**, so Material's (0,2,0) rule out-specifies it — order-independent. The structural change is what makes the same declaration behave correctly.

## Verification

Two independent measurements (Storybook, and a harness built from the production bundle) now agree. Both governing rules confirmed loaded before measuring:

| context | font-size | box | clipped |
|---|---|---|---|
| standalone in 16px container | 24px | 24×24 | no |
| standalone in 14px container | 24px | 24×24 | no |
| standalone in 12px container | 24px | 24×24 | no |
| text button | 18px | 18×18 | no |
| outlined button | 18px | 18×18 | no |
| raised button | 18px | 18×18 | no |
| icon button | 24px | 24×24 | no |
| pager `west` (leading) | 18px | 18×18 | no |
| pager `east` (trailing) | 18px | 18×18 | no |

The 12px/14px cases are the ones that were broken — an icon inheriting 12px in a 24px box. Standalone icons no longer depend on their container's font-size.

Production build re-verified: **0** `.material-icons[_ngcontent-*]` copies, one global rule.

Worth noting two of my earlier harness attempts gave wrong numbers (missing Material's button rule, so icons just inherited from the button and read 13.3px/24px). I only trusted the results once I asserted both rules were actually loaded — that check is in the verification script now. Still needs confirming on dev after deploy.

PR #3415, opened 2026-09-09
fix(resource-editor): show the no-access notice for any restricted media (DEV-7072)

domsteinbach:
Fixes [DEV-7072](https://linear.app/dasch/issue/DEV-7072/restricted-media-representation-resources-route-to-the-plain-view-so). Follow-up to DEV-7016 / #3383.

## The bug

A representation whose file value is withheld (resource viewable, `userHasPermission` V/RV/CR, but the file value absent) never reached its media wrapper. `getResourceType()` gated all six file-based types behind `if (fileValue)`, and the class-IRI fallbacks below covered only `Region` and the two segment types. So a restricted image / video / audio / pdf / document / archive / text resource fell through to the async compound count and rendered the plain view: header and property tabs, **no media area and no message at all**.

This is also why the `@else { <app-representation-restricted /> }` branch added to the video and audio wrappers in #3383 was unreachable *in the app*.

## The fix

**Inverted the sources of truth.** The class definition decides *what kind of thing a resource is* — it is stable schema, whereas the file value is per-request data that permissions can withhold. Classification now reads the declared file-value cardinality; the file value is consulted only for the one distinction the class cannot make: **Document vs Pdf**, which needs the filename.

Previously the file value gated all six types while adding no classification information — `properties` can only hold a file value whose cardinality the class declares, so the two can never disagree.

Also dropped the `getFileValue(...)!` assertions in the five remaining wrappers and guarded both the legal panel and the viewer with `@if (fileValue; as file)`. Without this, fixing the routing alone would surface `TypeError: Cannot read properties of null (reading 'copyrightHolder')` — the DEV-7016 error — across those five media types.

## Why all media types, not just AV

Verified against dsp-api: **value-level permission filtering is entirely type-agnostic.** In `ConstructResponseRdfDataParser.makeRdfPropertyValuesForResource` the permission filter runs *before* `rdf:type` is ever read, and a trailing `filterNot { valObjs.isEmpty }` removes the property key entirely when no values remain — exactly the observed symptom. `PermissionUtilADM` reads only `attachedToUser`, `attachedToProject` and `hasPermissions`, with zero references to any file value type. Any of the six file properties can be absent.

Separately, the **RestrictedView** mechanism (`RestrictedView.Size` / `Watermark`) is *not* the same thing: it is a project-level setting enforced by Sipi, leaves the file value present in the response, and only degrades anything for still images — `sipi.init.lua`'s `file_pre_flight` grants full access for permission code 1 on non-image files.

## Reviewing the three commits

The commits are split so the regression evidence is visible in history rather than asserted:

| commit | logic | specs | result |
|---|---|---|---|
| `46d28e8` fixtures | **unchanged** | corrected | 27/27 → fixtures valid |
| `572989b` the fix | **new** | *identical to previous commit* | 27/27 → **no regression** |
| `83cacfc` coverage | new | +13 new tests | 40/40 |

The first commit exists because the original fixtures set `properties` with a file value but **no `entityInfo`** — a resource dsp-api cannot return (`ResourcesConversionUtil` assigns `entityInfo` unconditionally). Correcting them against the *current* implementation first makes them a usable reference; the second commit then leaves them untouched, so the six types that already worked are provably unaffected.

## Testing

- **48/48 suites, 427 tests** passing; typecheck, prettier and eslint clean.
- New unit coverage: routing by declared cardinality for five types; Document staying Document when the filename is unavailable; a class several levels below its representation superclass (dsp-api flattens inherited cardinalities); segment IRI winning over a declared cardinality; absent `entityInfo` falling back to the compound check; and three dispatcher specs (synchronous classification, no compound-count call, never the plain view).

### Storybook — verified in CI

`nx run dsp-app:build-storybook` fails locally with an ESM/CJS interop error in `@storybook/angular`'s builder (this reproduces on `main` with these changes stashed, so it predates this branch). The play functions therefore could not be executed while developing.

**CI has now run them: 241 suites / 566 tests passing in Chromium.** All eight story files carrying a withheld-file story report `PASS` — the five new wrapper stories, the dispatcher routing story, and the two pre-existing video/audio stories from #3383. Note the runner reports per file rather than per story, so a green file means every story in it passed.

## Note on the existing video/audio stories

They mount their wrapper directly, bypassing the dispatcher — which is why they passed while the app stayed broken (only the dispatcher calls `getResourceType()`). That is the correct scope for a wrapper story, but it means routing needed its own coverage: hence the new dispatcher story and specs.

## Out of scope

Banner precedence — an `app-resource-restriction` banner stacking above the new `app-representation-restricted` — belongs to DEV-6579, which owns that banner's trigger condition across 12 files.


domsteinbach:
@claude review

derschnee68:
@claude review

domsteinbach:
### CI status

All checks relevant to this change are green:

| job | result |
|---|---|
| **Storybook Interaction Tests** | ✅ 241 suites / 566 tests (Chromium) |
| DSP-APP Linting | ✅ |
| DSP-APP E2E (heavy / rest) | ✅ |
| Docs Build Test, CodeQL, Analyze | ✅ |
| Check OpenAPI Spec is Up-to-Date | ❌ — see below |

**The OpenAPI failure is unrelated to this PR.** This branch touches no OpenAPI files. The job detected that dsp-api moved on (`v38.0.0-8-gdf2abf9` → `v38.1.0-5-g35a36e3`) and auto-opened #3416 to bump the vendored spec; its log reads *"An open dsp-api spec-bump PR already exists — skipping"*. It fails by design to signal the drift, and would fire on any PR opened right now. #3416 has auto-merge armed — once it lands, rebasing this branch clears the check. Per CLAUDE.md the vendored spec must not be edited to silence it.

The Storybook result is worth calling out because it was the one thing I could not verify while developing: `build-storybook` fails locally on an ESM/CJS interop error in `@storybook/angular`'s builder, which also reproduces on `main`. CI has now executed the play functions.

derschnee68 (approved):

PR #3422, opened 2026-09-11
fix: fresh clone setup fails loud on the wrong toolchain

jnussbaum:
[DEV-7198](https://linear.app/dasch/issue/DEV-7198)

## Summary
- A committed `.npmrc` makes a fresh clone abort loud (`EBADENGINE`) on the wrong Node or npm version, instead of failing later with an unrelated error.
- `allow-git=root` permits only the project's own `ckeditor5-custom-build` git dependency, which npm 12 refuses by default.
- `allowScripts` allowlists every dependency that needs an install script, using name-only keys so a version bump doesn't invalidate the entry.
- The README gives one instruction to activate the `.nvmrc` pin, instead of offering an alternative.

## Changes
- `.gitignore`: narrowed the bare `.npmrc` rule with `!/.npmrc` so a committed root `.npmrc` is tracked while a nested one stays ignored.
- `.npmrc` (new): `engine-strict=true`, `allow-git=root`.
- `package.json`: widened `engines.npm` to `^10.9.0 || ^11 || ^12`; added a 12-entry `allowScripts` block.
- `package-lock.json`: synced the mirrored root `engines` field.
- `README.md`: Quick Start now gives one instruction to activate `.nvmrc`.

## Test Plan
- Verified live on Node 26.8.1 / npm 12.0.2 (no toolchain manager available on this machine): `npm install` aborts with `EBADENGINE`, naming both the required and actual versions.
- `npm install-scripts ls` reports zero uncovered packages except `@tybys/wasm-util` — confirmed harmless: its `prepare` script (`tsgo build`) has never been runnable by any consumer (`tsgo` is its own devDependency, never installed downstream) and isn't needed, since the published package already ships pre-built output. Full investigation in the plan.
- `git check-ignore -v .npmrc` confirms the root file is tracked; a nested `.npmrc` stays ignored.
- A real install under the `.nvmrc`-pinned Node 22.23.1 was not run locally (no matching Node available here) — CI and a follow-up check on a pinned machine still cover this.
- Plan and execution journal: `specs/2026-09-08-dsp-app-fresh-clone-setup/` in dasch-specs.

## Screenshots
N/A — no UI changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ testem.log
 /typings
 metals*
 .npmrc
+!/.npmrc
 
 #Documentation Files
 docs/_site/
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+engine-strict=true
+allow-git=root
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This monorepo is built on:
 
 ## Quick Start
 
-Use the NodeJs version supported by the installed Angular version ([compatibility table](https://angular.dev/reference/versions)), or the version pinned in [.nvmrc](.nvmrc).
+Activate the Node.js version pinned in [.nvmrc](.nvmrc) — for example `nvm use`, `fnm use`, or `n auto` — before running `npm install`.
 
 ```shell
 npm install
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,7 +114,7 @@
       },
       "engines": {
         "node": "^22.13.0",
-        "npm": "^10.9.0"
+        "npm": "^10.9.0 || ^11 || ^12"
       },
       "optionalDependencies": {
         "@nx/nx-darwin-arm64": "23.0.1",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,21 @@
   },
   "engines": {
     "node": "^22.13.0",
-    "npm": "^10.9.0"
+    "npm": "^10.9.0 || ^11 || ^12"
+  },
+  "allowScripts": {
+    "@openapitools/openapi-generator-cli": true,
+    "@parcel/watcher": true,
+    "@sentry/cli": true,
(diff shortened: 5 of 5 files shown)


mpro7 (approved):
All good from my side. You're right on SC-1, and I can close both open items.

**SC-1: conceded.** `checkEngine()` throws one joint `EBADENGINE` naming the whole `engines` object and both actual versions. My own run proves it against me: with only Node failing it still printed `Required: {"node":"^22.13.0","npm":"^10.9.0"}` / `Actual: {"npm":"10.9.0","node":"v22.12.0"}`. Leave the wording.

**Engines measurement:** 1530 of 2222 installed packages declare `engines`, **zero** fail against Node 22.23.1, npm 10.9.0, 11.19.1 or 12.0.2.

**OQ-2 can close.** Real-lockfile run: fresh clone of `main`, `.npmrc` with `allow-git=root`, npm 12.0.2 on Node 22.23.1. `npm install` exits 0 with the 13-package blocked warning, the root postinstall (`generate-openapi-module && nx reset`) completes inside it, `npx nx --version` prints 23.0.1. Blocked scripts are not a cause of the `WorkspaceContext` crash. Caveat: macOS arm64, but Linux gets its binary via the same `optionalDependencies` mechanism.

**One slip:** "the other two install inside a Docker build, not on the runner" is backwards. `make docker-publish` → `docker-publish-app` → `app-build-prod` → `install-dependencies` → `npm install` runs **on the runner**, then the nx prod build, then `docker buildx build`; the Dockerfile is `FROM nginx:1` and only copies `dist/apps/dsp-app`. So those two run a plain `npm install` with scripts enabled. Your conclusion holds (all five pin Node 22.23.1 → npm 10.9.8), but they are the release path, so they are what meets blocked scripts when CI's npm moves to 12.

PR #3423, opened 2026-09-13
docs: add ARCH-MAP, CONTEXT and ADR-0001 through ADR-0006

subotic:
Adds agent-facing navigation documents and six ADRs recording the architecture this codebase already has.

Nothing here changes a line of application code. The tag migration, the library splits and the e2e project are all recorded as decisions, not carried out.

## What is in here

**`ARCH-MAP.md`** and **`CONTEXT.md`** at the repository root: the topology (36 components, every tracked file mapped to exactly one) and the domain glossary (42 terms, 13 flagged ambiguities).

**`docs/adr/0001` through `0006`**, published through mkdocs under a new nav section.

| ADR | Decision |
| --- | --- |
| 0001 | Library taxonomy and enforced dependency direction |
| 0002 | The barrel is the only public interface |
| 0003 | Feature state is component-provided, with one writer per value |
| 0004 | Cross-feature access |
| 0005 | What `shared/` means, and what it does not |
| 0006 | The end-to-end suite is its own project |

Each rule carries an enforcement level (`structure`, `static-analysis`, `review`, `docs-only`), matching the vocabulary `dasch-ops-platform` already uses. Where a rule is `review` today and could be `static-analysis`, the ADR names the configuration change that would promote it.

## The finding that shaped ADR-0001

The tag decision was scaffolded and then abandoned. All 28 library `project.json` files carry a `"tags"` key; 24 are `[]`. `depConstraints` is still a single `{ sourceTag: '*', onlyDependOnLibsWithTags: ['*'] }`.

The proposed constraint block was **simulated against the real runtime graph** with all 29 projects tagged, applying Nx's own evaluation rules. It rejects 21 edges across 45 files:

| Constraint | Edges | Files |
| --- | --- | --- |
| `type:data-access`, `type:util`, `type:feature`, `type:app` | 0 | 0 |
| `type:ui` | 4 | 6 |
| `scope:*` | 17 | 39 |

The type tier is already clean. Essentially all the cost is cross-page coupling, which is what ADR-0004 exists to remove. ADR-0001 therefore records that it is **not independently landable**: its constraint block waits on 0004 and 0005.

## Two constraints worth reading before anyone starts the migration

- **The rollout order is forced.** Nx refuses dependencies from any project matching no constraint, so all 29 projects must be tagged while `* -> *` is still active, and the constraint block swapped in a separate commit.
- **`*.stories.ts` is currently boundary-checked.** `eslint.config.mjs` exempts environment and spec files but not stories. Several stories import across tiers their runtime code does not touch, so a strict `type:ui` rule fails on libraries whose shipped code is clean.

## Two blockers the simulation surfaced

- **`shared/app-common-to-move` cannot be retagged.** Five page scopes import it across nine files. `scope:app` breaks all nine; `scope:shared` fails the direction test at 6 in / 9 out. It has to be split first, and splitting it surfaces a real design question: two pages import the application header.
- **`ontology -> project` is 12 of the 39 files**, all importing `ProjectPageService`. Relocating that one service clears about a third of the work.

## One open question, deliberately unanswered

Is `pages/data-browser` a page or shared infrastructure? Three other features render its internals. If it is shared, its tags become `scope:shared, type:feature` and **six of the seventeen edges become legal with no code moving**. That needs product knowledge, not graph analysis.

## ADR-0006 can land on its own

Investigating the e2e suite turned up the opposite of what was expected. The Nx wiring is already there and is good: `apps/dsp-app` has an `e2e` target using `@nx/cypress:cypress` with `devServerTarget` for three environments, and `package.json` exposes it.

Nothing reaches it. CI hand-rolls the sequence instead, backgrounding the serve target, polling `localhost:4200` with curl for up to 180 seconds, then running `npx cypress run`. Combined with `eslint.config.mjs:173` excluding `apps/dsp-app/cypress/` and no target ever resolving the tsconfig project reference, the suite's imports have rotted unnoticed: four files import `libs/vre/open-api/src`, a path that has not existed since that library moved. They pass because they are never compiled (DEV-7251).

## Related issues

Filed from this work, all in Triage: DEV-7245 through DEV-7254.

DEV-7254 covers the last finding: `useInferencePlugins: false` was set by an Nx 18 codemod in a version-bump PR, not chosen, and target definition is now split three ways with no rule behind the split.


CONTEXT.md
```diff
@@ -0,0 +1,204 @@
...
+**Resource class**:
+A type of resource defined by a project, declaring which properties its instances may carry.
+_Avoid_: Type, entity type
+
+**Ontology**:
+A project's data model: the set of resource classes and properties it defines.
+_Avoid_: Schema, data model (in code and URLs; see Flagged ambiguities)
+
+**Cardinality**:
+The constraint on how many values of one property a resource of a given class may carry, one of required, optional, may-have-many, or at-least-one.
```
domsteinbach: I would add or mention "cardinalities" which is used by the API's data model, the responses and the DSP-JS for what is commonly know as "property" or "predicate". Imho a big naming flaw calling properties cardinalities, but that is the model.

CONTEXT.md
```diff
@@ -0,0 +1,204 @@
...
+
+- **"List" in code, "controlled vocabulary" in the UI.** The UI rename is incomplete: the property-type picker still shows "Select list" and the property-type group is still labelled "List". Canonical term for code is **List**. The UI inconsistency is a real defect, not a naming choice.
+
+- **"Region"/"Segment" in code, "Annotation" in the UI.** The UI calls both Annotation; the code keeps them distinct because they mark different media and have different geometry. Canonical terms are **Region** and **Segment**. This is the widest gap between code and UI vocabulary in the app.
+
+- **"Cardinality" means two different things.** In `dsp-js` it is the enum of allowed counts, and it is also the DSP-API name for the class-to-property assignment record itself (`UpdateResourceClassCardinality` updates a class's properties). Two code comments in the repo complain about this. Canonical: **Cardinality** is the count constraint; call the assignment a **class property declaration**.
```
domsteinbach, line 186: the API's data model, it's response bodies call "cardinalities" what is commonly know as "property" or "predicate". Imho a big naming flaw calling properties cardinalities. DSP-JS is mirroring that as it operates on those types.

domsteinbach (commented):
I like it, let's open some issues for relocation cleanups and adding tags

derschnee68:
Looks good to me. There are some nx libs that contain miscelleaneous files. This could be restructured and then we can remove those "misc" paragraphs.

derschnee68 (approved):

mpro7 (dismissed):
Read all six ADRs and re-measured every count against today's `main`. The decisions hold; a few facts and one design point need fixing before merge.

**Fix before merge**

- **Rebase, several facts are stale since this morning.** #3434 added the e2e `tsc --noEmit` step, repointed the rotted imports and declared `UploadedFileResponse` locally, so ADR-0006 items 2 and 4 and decision 3 describe finished work. #3437 deleted `status-msg` and the `ontology-properties` alias, so ADR-0005 decision 6 is done, there are 28 projects, and the ADR-0001 tag table has a dead row. #3436 fixed DEV-7249, which ADR-0003 cites as a current defect. The `search -> project` edge is gone: 11 page edges now, not 12.
- **"`app.routes.ts` imports all eight page libraries" is wrong** (ADR-0001 context, ADR-0004 decision 3). It imports seven; `DataBrowserPageComponent` is exported by `pages/project`. ARCH-MAP in this PR already says data-browser is "a component library, not a routed page". That answers the open question without product input: tag it `scope:shared, type:feature` now and six scope violations disappear.
- **ADR-0004 decision 4 contradicts decision 1.** "The provider depends on the consumer's token" is a page-to-page import, which ADR-0001's scope block rejects. Only `scope:app` may import both, so the adapter has to live in `apps/dsp-app`.
- **`ProjectPageService` is pulled two ways.** ADR-0004 calls it stateless and relocates it to shared data-access; ADR-0003 calls its root provision the defect. It holds two `BehaviorSubject`s. Ontology, list and search routes are all children of `ProjectPageComponent`, so the consistent answer is: class moves to shared data-access, provision moves to `ProjectPageComponent`, children inject from the ancestor. Say so in both records.

**Numbers**

- `ProjectPageService` "six importers" / `AllProjectsService` "four" (0001, 0003) vs "4 / 2 page libraries" (0004). Measured: 4 libs in 18 files, 2 libs in 4 files.
- The Nx-20 single-slash risk cites `@dasch-swiss/vre/pages/...`; those are tsconfig aliases. Project names are `vre-pages-project-project` style, no slash. Drop it.
- `resource-editor` fan-in is 3, not 2: `pages/project` reaches it via dynamic `import()` in `data-class-panel.component.ts`.

**Suggestions**

- The per-file inventory in ADR-0001 decision 7 and the `useInferencePlugins` section rotted within one day. Move both into the DEV issues, keep the ADR to the decision, constraint block and tag table.
- Where a rule is `review` and the index promises a promotion path, name it: a grep on `@use .*apps/dsp-app` for Sass, `no-restricted-syntax` on `export *` in new barrels (0002) and on `providedIn: 'root'` under `libs/vre/pages/**` (0003).

CONTEXT.md
```diff
@@ -0,0 +1,204 @@
...
+
+The browser client for the DaSCH Service Platform. It lets humanities researchers define data models, enter and curate research data, and browse and cite it, talking to DSP-API over two HTTP clients. This file fixes the vocabulary so that one word means one thing.
+
+## Language
+
+### Research data
+
+**Project**:
+A research undertaking that owns its own data model, data, and members.
+_Avoid_: Repository, collection
```
BalduinLandolt:
```suggestion
_Avoid_: Repository, collection, dataset
```

CONTEXT.md
```diff
@@ -0,0 +1,204 @@
...
+**Erase**:
+To remove a resource or an entire project from the triplestore irrecoverably.
+_Avoid_: Hard delete, purge
+
+**Standoff**:
+Rich-text markup stored as RDF separately from the text it annotates, which is what makes text values linkable.
+
+**GUI element**:
+The widget a property declares for entering its values, drawn from the salsah-gui vocabulary.
+_Avoid_: Property type (in code; see Flagged ambiguities), input type, control
```
BalduinLandolt, line 204: alongside GUI element, there is also GUI attribute (used in lists)

CONTEXT.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Example dialogue
+
+> **Dev:** "When a user adds a **Property** to an existing **Resource class**, do we need a new **Ontology** version?"
+>
+> **Domain expert:** "No, but you must send the ontology's last modification date, and the **Cardinality** you set decides whether existing **Resources** still validate. Adding a required one to a class that already has instances is refused."
+>
+> **Dev:** "And if the **Property** holds terms from a **List**?"
+>
+> **Domain expert:** "Then the **Value** points at a **List node**, not at a string. If someone later deletes that node the value still resolves, because we **Delete** rather than **Erase**."
```
BalduinLandolt: In my view, this is incorrect on two levels:

- The whole vocabulary is following code, not the domain language, which I already don't like; here we are reenforcing this.
- Our domain experts exactly _do not_ use these terms (they use data model, not ontology) and they definitely don't talk about implementation details such as "last modification date".

I guess this example dialogue is meant to show how devs and domain experts can understand each other with regard to this application; but it's getting this exactly wrong.

CONTEXT.md
BalduinLandolt: by calling the existence of a property class on a resource class "property" rather than "cardinality", this gives a double meaning to the term "property", which is similarly unfortunate

BalduinLandolt (approved):
I'm not too happy about the vocabulary, because it cements _not_ using the domain language. But that's a detail. Other than that, it looks good to me.

BalduinLandolt:
> I'm not too happy about the vocabulary, because it cements _not_ using the domain language. But that's a detail. Other than that, it looks good to me.

It may also be worth noting that the vocabulary defined here does not fully agree with the one defined in https://github.com/dasch-swiss/dsp-api/pull/4331 .

subotic:
@mpro7 Rebased onto main and re-measured everything against `238000ec2`, runtime code only (specs, stories and stories.helpers excluded, since ADR-0001 decision 5 exempts them). That basis is now stated in ADR-0001's context so the numbers are reproducible. Where my counts differ from yours it is that exclusion: `ProjectPageService` is 3 other libraries in 16 files, `AllProjectsService` 2 in 3.

Fixed as you described: 28 projects, seven routed page libraries, data-browser resolved to `scope:shared, type:feature` (ADR-0004 decision 7, six imports across five files become legal), decision 4's adapter moved to `apps/dsp-app`, the Nx 20 slash bullet dropped, resource-editor fan-in 3, DEV-7249/7251/7252 in the past tense, ADR-0006 split into what #3434 patched and what stays structural.

One correction to your `ProjectPageService` resolution: `ProjectPageGuard` seeds the service today, and a root-provided guard cannot inject a component-provided service. Both ADR-0003 and ADR-0004 record your answer (class to shared data-access, provision on `ProjectPageComponent`, children inject the ancestor) plus that constraint for the implementing issue.

Not verified: your "11 edges now, not 12". The `search/search -> project` edge is gone, but `advanced-search -> project` remains in two files, so the table keeps it.

Your two Suggestions (inventory into Linear, naming the grep promotion paths) are not in this pass. Happy to do either as a follow-up if you want them before merge.

CONTEXT.md
subotic: You were right on both counts, and the dsp-api PR being merged settled how to fix it. CONTEXT.md now opens with an ownership rule: dsp-api's `CONTEXT.md` and its per-context files own domain vocabulary, and where a term appears in both repositories the dsp-api definition wins. Concretely: Data Model is the prose term and ontology stays a code and URL name; Cardinality is accepted as the class-to-property binding, dropping the coined "class property declaration"; the permission vocabulary follows dsp-api. The dialogue is rewritten so the domain expert talks about data models and cardinalities, not modification dates.

What dsp-api does not name yet (Representation, Region, ARK URL, GUI element, and so on) is listed as client-owned with the definitions to propose upstream. DEV-7291 tracks that side, including two inconsistencies in dsp-api's Data Model file (singular versus plural Data Model, and Lists as contained by a Data Model).

CONTEXT.md
subotic: Taken, plus tenant from dsp-api's avoid-list.

CONTEXT.md
subotic: Added as its own entry.

CONTEXT.md
subotic: Resolved by accepting dsp-api's meaning: a Cardinality is the binding of Class, Property and count, so the API's naming is its definition rather than a flaw. The dsp-js enum carrying only the count under the same name stays flagged.

subotic:
@BalduinLandolt I agree, but I cant change the code. I'm documenting what is. Just stating that things should be called differently, is not going to change the code. The only thing we can do is map terms. Don't shoot the messenger. If you want dsp-app and dsp-api for that matter to use domain terms, then we should first define them separately and then create projects to rename stuff in code.

PR #3433, opened 2026-09-14
refactor(advanced-search): single source of truth for URL param names

domsteinbach:
Resolves [DEV-7252](https://linear.app/dasch/issue/DEV-7252/advanced-search-url-parameter-names-have-two-sources-of-truth-and-the).

## Problem

The advanced-search query-parameter names existed twice:

- Five `RouteConstants.advancedSearch*` members that **nothing read**
- The string literals in `SearchUrlSyncService`, which actually own the URL contract

The trap was that the constants looked authoritative. Renaming `RouteConstants.advancedSearchFilters` to change a URL would compile clean, pass tests, and change nothing — the live values were the literals. The reverse drift was equally available.

The URL is the durable state of advanced search, so a saved or shared search link is a user-facing contract. Latent rather than broken today, hence tech-debt.

## Change

Delete the five dead constants and replace the literals with a local `ADVANCED_SEARCH_PARAM` object in `SearchUrlSyncService` — the only place that should know the URL shape.

```ts
export const ADVANCED_SEARCH_PARAM = {
  q: 'q',
  ontology: 'ontology',
  class: 'class',
  filters: 'filters',
  orderBy: 'orderBy',
  orderDir: 'orderDir',
} as const satisfies Record<keyof SearchUrlParams, string>;
```

`satisfies Record<keyof SearchUrlParams, string>` ties the set to the decoded param interface, so a new param can no longer be added to one without the other.

Two gaps closed while streamlining:

- **`orderDir`** previously had no constant at all, so the two sets were not even the same shape. It is now included.
- **`'desc'`** appeared as a bare literal four times across the read and write paths. Now `ORDER_DIR_DESC`, typed `OrderDirection`.

Keeping the names local rather than wiring them to `RouteConstants` follows the issue's preference and avoids pulling one page's URL shape into a library with fan-in 19. Not exported from the lib barrel — nothing outside the service needs it.

## No behaviour change

The constants and the literals matched exactly, so no URL changes. Existing links keep working.

## Verification

- `nx test vre-pages-search-advanced-search` — 245 passed, 14 suites
- `nx test vre-core-config` — 9 passed
- `nx lint` on both libs — clean
- Word-boundary grep across `.ts`/`.html` confirms zero surviving references to the removed constants

`nx build dsp-app` fails on the `vre-open-api:generate` prerequisite (openapi-generator-cli needs network to fetch its generator). Verified this fails identically on unmodified `main` — pre-existing and unrelated to this change.


derschnee68:
@claude review

derschnee68 (approved):

PR #3436, opened 2026-09-14
fix(session): stop sending an expired JWT from dsp-js after auto-login rejects it

domsteinbach:
Resolves [DEV-7249](https://linear.app/dasch/issue/DEV-7249/expired-jwt-keeps-being-sent-by-dsp-js-after-auto-login-clears-it-from).

## Problem

The app carries the access token on two independent channels:

- **`authInterceptorFn`** re-reads `localStorage` per request, so it stops sending a token the moment storage is cleared.
- **dsp-js** sends the in-memory copy held on the shared `KnoraApiConfig`, seeded once at startup.

`AutoLoginService`'s invalid-token branch cleared only the first — it called `removeToken()` but left the in-memory copy set. From that point the two channels disagreed for the rest of the page lifetime: dsp-js kept sending `Authorization: Bearer <expired>` while HttpClient correctly sent nothing.

Compounding it, the in-memory copy was seeded moments earlier by `apiConnectionTokenProvider`, which read `localStorage` raw and never checked validity — so dsp-js started the page already holding the very token auto-login was about to reject.

## Fix

Two clears, both in app code. **Nothing in `libs/dsp-js/` changes** — dsp-js has no way to know a token became invalid; keeping its config current is the app's responsibility.

**1. Reactive — `auto-login.service.ts`.** The invalid-token branch now clears the in-memory token alongside `removeToken()`. `afterLogout()` already cleared both; this branch was the only gap.

**2. Preventive — `api-connection-token.provider.ts`.** Seeds through `AccessTokenService.isValidToken()` instead of reading `localStorage` raw, so an expired or malformed token is never seeded at all. This is arguably the more important half: without it, every page load re-seeds an expired token and the reactive clear is only cleanup after the fact.

**3. Provider moved** from `pages/user-settings/user` to `core/session`. Not cosmetic — it is what makes (2) possible, since reaching `AccessTokenService` from a `pages/` lib would be the wrong dependency direction. The connection's token is session state, not user-settings state. Its only consumer was `app.config.ts`, so the move is one import change plus both barrels.

## Rejected alternative

Teaching dsp-js to validate tokens would push app session policy into a shared client. The token's validity is the app's concern.

## A hollow test, repaired

`auto-login.service.spec.ts` already asserted `jsonWebToken` was `''` after the invalid-token branch — and **passed against the broken code**. The mock merely *initialised* to `''`; nothing ever cleared it. It now seeds a real token first, so it tests the behaviour rather than the mock's starting value.

## Verification

- `nx test vre-core-session` — 5 suites, **60 tests pass**
- `nx test vre-pages-user-settings-user vre-core-config` — 2 suites, 9 tests pass
- `nx lint` clean on `vre-core-session`, `vre-pages-user-settings-user`, `dsp-app`

Mutation-tested, since tests passing on correct code prove nothing:

| Mutation | Result |
|---|---|
| Remove the in-memory clear from `auto-login` | 2 failed |
| Revert provider to unvalidated `localStorage` read | 1 failed |
| Both restored | 60 passed |

The two failures are exactly the intended tests — the new regression test and the repaired one.

## Follow-up

**[DEV-7261](https://linear.app/dasch/issue/DEV-7261/session-token-has-two-channels-kept-in-sync-by-convention-fold-them) — fold both token channels behind one owner.** This PR fixes the one place the "clear both channels" convention was forgotten. It does not change the shape that allowed it: three call sites still pair a storage write with an in-memory write by hand, and nothing structural stops the next one from doing only half.

| Call site | Storage | In-memory |
|---|---|---|
| `auth.service.ts:28-29` — login | `storeToken()` | `= encodedJWT` |
| `auth.service.ts:50-51` — `afterLogout()` | `removeToken()` | `= ''` |
| `auto-login.service.ts:39-40` — invalid token | `removeToken()` | `= ''` (this PR) |

A `SessionTokenService.set()/clear()` owning both would make partial updates unrepresentable rather than merely discouraged. Raised by the dune legibility reviewer during review of this PR; filed as tech-debt since nothing is broken today.

## Out of scope

**401 handling.** There is none anywhere in the app, so an expired session still surfaces as generic error snackbars until reload. The issue raises it as an open design question (interceptor vs. error handler), so it belongs in its own ticket rather than being decided here.

**Mid-session expiry.** A token valid at bootstrap that expires while the page is open is not addressed here either — noted on DEV-7261.


domsteinbach:
@claude review

derschnee68 (approved):

PR #3437, opened 2026-09-14
chore: remove verified dead code (DEV-7253)

domsteinbach:
Resolves [DEV-7253](https://linear.app/dasch/issue/DEV-7253/remove-verified-dead-code-one-unused-lib-a-broken-path-alias-and-40).

**41 files, +5 / −1070.** Eleven commits, one per unit of work, each independently revertible.

## Two corrections to the issue

The issue's findings were re-verified against the tree before executing. Two were wrong:

**1. `RouteConstants`: 18 removable, not 21.** Five of the 21 are referenced *inside* `app-constants.ts` by `*Relative` template-literal composition — invisible to a `RouteConstants.<name>` search, because the composition site is in the declaring file.

| Constant | Feeds | Live? | Action |
|---|---|---|---|
| `ontologyRelative` | `ontologyEditorRelative` | `app.routes.ts:68` | **kept** |
| `resourceParameter` | `projectResourceRelative` | `app.routes.ts:186` | **kept** |
| `projectParameter` | `projectResourceRelative` | `app.routes.ts:186` | **kept** |
| `addClassInstance` | `OntologyClassAddRelative` | dead | removed as a pair |
| `modeParameter` | `searchProjectRelative` | dead | removed as a pair |

Deleting the three "kept" would have broken the build. Since that trap is invisible to the grep that produced the original list, **each now carries a comment naming its consumer** — otherwise the next cleanup pass repeats the mistake.

**2. `@dasch-swiss/vre/ontology/ontology-properties`: only the alias is dead.** It points at a directory that does not exist. The real `OntologyPropertiesComponent` lives in the `ontology/ontology` lib, is routed at `app.routes.ts:82`, and has stories. Untouched — one `tsconfig.base.json` line removed.

## What a plan review caught

Review of the plan (consistency + angular + dune auditors) found a defect that would have broken five libraries: **the plan omitted barrel cleanup.** `export * from './deleted-file'` is `error TS2307`, a hard compile break — and a "grep for each removed symbol" verification step cannot catch it, because `export *` lines don't name the symbol. Nine barrel lines are now pruned alongside their files.

Also caught: `nx.json:147` still listed the deleted `status-msg` path, and `ui.common.entities.members` is **still used** by `settings-page.component.ts:72` — so item 8 removes only the dead TS branch and leaves the i18n keys intact.

## Commits

1. Remove unused `status-msg` library (+ alias, + `nx.json` include)
2. Delete broken `ontology-properties` path alias
3. Remove unreachable project-search component island
4–7. Remove unreferenced symbols, split per Nx project: `search-search`, `api`, `session`, `advanced-search`
8. Remove five NgModule-era bulk-import arrays
9. Remove 18 `RouteConstants` + guard the composed ones
10. Remove four satisfied `.gitkeep.spec.ts`
11. Remove dead `member` branch in `itemPluralMapping`

Commit 8 also deletes the only `vre/ui/ui` → `vre/ui/date-picker` graph edge (sole cause: `ui.components.ts:1`). `DateValueHandlerComponent` stays reachable via its direct consumer.

## Verification

- **Lint: 11 projects clean** — `core-config`, `core-session`, `shared-app-api`, `search-search`, `advanced-search`, `ui-ui`, `progress-indicator`, `string-literal`, `system`, `user-settings`, `dsp-app`
- **Zero `TS2307` missing-module errors** across every touched lib (the specific failure mode barrel deletion causes)
- **Tests:** `ui-ui` 46, `error-handler` 78, `app-common` 43, `core-config` 9, `advanced-search` 88, `shared-app-api` 1 — all pass
- Repo-wide grep confirms no surviving reference to any removed symbol

**Pre-existing failure, not from this PR:** some suites can't load due to a missing `libs/vre/3rd-party-services/open-api/src/generated` (needs `nx run vre-open-api:generate` with network). Verified identical failures on unmodified `origin/main` — `search-search`, `system-system`, and `core-session` fail the same way there.

## Not in scope

dsp-js items (`UserCache`, `ListAdminCache`, `Constants.Dsp*`, `PATH_METADATA`). `libs/dsp-js` is published to NPM, so removing exported symbols is a breaking change requiring a deprecation decision.

The five advanced-search query-param constants are left to #3433, which replaces them with a local contract. Disjoint lines — no conflict in either merge order.

## Follow-up

No `knip` / `ts-prune` in the repo, so nothing prevents dead code re-accumulating; this audit was manual and error-prone enough to produce two wrong claims. Worth a scoped trial, tuned for barrel-exported public API and dsp-js's published surface before gating CI.


derschnee68:
@claude review

derschnee68 (approved):

PR #3440, opened 2026-09-14
fix(resource-editor): localize the compound representation tab label (DEV-7043)

derschnee68:
Resolves [DEV-7043](https://linear.app/dasch/issue/DEV-7043/dsp-app-untranslated-placeholder-image-1-shown-in-still-image-resource) in part — see **Scope** below.

## Problem

`resource-compound-tabs.component.ts` labelled the incoming-representation tab with `entityInfo.classes[type].label` — the single-language field. That is whatever one string the server resolved, so a German-authored ontology rendered `Bild Original` to a user whose UI language was English, with no attempt to pick the English variant.

## Fix

Resolve the class label against the current UI language instead of dropping it:

```diff
-<mat-tab [label]="resourceClassLabel(incomingResource.res)">
+<mat-tab
+  [label]="
+    (incomingResourceClass?.labels | appStringifyStringLiteral) || ('resourceEditor.representation' | translate)
+  ">
```

Ontology class labels **are** per-language. `OntologyCache.requestItemFromKnora` fetches with `allLanguages=true` ([`OntologyCache.ts:163`](../blob/main/libs/dsp-js/src/cache/ontology-cache/OntologyCache.ts#L163)), so `ResourceClassDefinitionWithPropertyDefinition` carries `labels: StringLiteralV2[]` — one entry per language the project authored — alongside the single-language `label`. `appStringifyStringLiteral` narrows that array to `LocalizationService.currentLanguage`, falling back through the fixed `en > de > fr > it > rm` chain from DEV-6875.

This is not a new mechanism: `app-resource-header` already renders the compound resource's *own* class label exactly this way ([`resource-header.component.ts:26`](../blob/main/libs/vre/resource-editor/resource-editor/src/lib/header/resource-header.component.ts#L26)). The incoming-representation tab was the one place that didn't.

A new `resourceEditor.representation` key (en/de/fr/it) is the **fallback only**, for a class carrying no `rdfs:label` at all, so the tab can never render blank. It is not the normal label — replacing the class label with a static string would drop the information the tab carries, namely which kind of representation this is (`Bild Original` vs. a page, a PDF original, …).

### Behaviour

| Ontology authored in | UI language | Tab label |
|---|---|---|
| `de` + `en` | EN | `Original image` |
| `de` only | EN | `Bild Original` |
| no label | EN | `Representation` |

`makeEntityInfo` gains a `classLabels` parameter mirroring the array the cache really holds. It defaults to an English literal built from the existing `classLabel` argument, so existing fixtures are unaffected.

## Scope — what this PR does *not* fix

The issue reports two adjacent strings, `Bild Original[Image #1]`. This PR addresses the first.

`[Image #1]` is **not** a frontend string. It appears nowhere in the repo (`git log --all -S "[Image"` over the full history returns nothing). It is the incoming resource's `rdfs:label`, rendered verbatim at [`incoming-resource-header.component.ts:13`](../blob/main/libs/vre/resource-editor/resource-editor/src/lib/header/incoming-resource-header.component.ts#L13) — user data, not localizable by the app. It was a placeholder produced by the project's bulk import.

That data appears to have since been corrected. Gravsearching project `0854`:

| Class | Resources | Labels containing `Image` or `[` |
|---|---|---|
| `daschland:ImageOriginal` | 99 | **0** |
| `daschland:Image` | 108 | **0** |
| `knora-api:StillImageRepresentation` | 221 | **0** |

Two of them are `"Alice finding a little door"` and `"Alice holding the 'Drink Me' bottle"` — the titles the issue's own Note cites. Worth confirming with the reporter before closing DEV-7043 on that half.

## Verification

- **Storybook interaction tests: 244 suites, 581 tests pass**, including three new stories (localized label / fallback to authored language / fallback to the generic label).
- `nx test vre-resource-editor-resource-editor` — 48 suites, 427 pass (1 pre-existing skip)
- `nx build dsp-app` — clean; `nx lint vre-resource-editor-resource-editor` — clean
- i18n key parity: de/fr/it vs en — 0 missing, 0 extra

Mutation-tested, since stories passing against correct code prove nothing:

| Mutation | Result |
|---|---|
| Replace the binding with the static `'resourceEditor.representation' \| translate` | **2 of 3 new stories fail** — `expected [ 'Properties', 'Representation' ] to deeply equal [ 'Properties', 'Original image' ]`. The no-label fallback story correctly still passes. |
| Restored | 5 passed |

## Follow-up

`Accept-Language` appears **nowhere** in dsp-app or dsp-js — `libs/dsp-js/src/api/endpoint.ts:175` `constructHeader` sets only `Authorization` and `Content-Type`, and no interceptor adds it. It does not affect this bug, because the client resolves the language itself from the `allLanguages=true` payload. But it means the app never tells dsp-api what language the user is in, so any string the *server* localizes comes back in the API's fallback language. Worth its own ticket.


derschnee68:
@claude review

mpro7 (approved):

PR #3449, opened 2026-09-16
fix(ci): freeze storybook animations so pixeleye stops reporting phantom diffs (DEV-7238)

derschnee68:
## Problem

Pixeleye reports a change on every build for the stories that render a spinner, even when nothing changed.

The `Loading` stories — resource-fetcher, video, audio, advanced-search-results, search-result, resources-list, property-row — render an indeterminate `mat-spinner`. Its rotation is a CSS keyframe animation that never ends. Pixeleye captures a story as soon as it renders, so each build photographs the spinner at an arbitrary frame and diffs it against a baseline photographed at a different one.

## Why not `pixeleye.config.ts`

The config has a `css` option that would have been the natural home for this, but it is only wired into the `pixeleye urls` command. `@pixeleye/storybook`'s `captureStories` builds its snapshot call by hand and forwards only `name`, `variant`, `selector` and `devices` — `css` and `wait` are dropped.

Per-story `parameters.pixeleye` is not a workaround either: the CLI reads `story.config?.parameters?.pixeleye` where `config` is *already* `parameters.pixeleye`, so it would take `parameters: { pixeleye: { parameters: { pixeleye: {...} } } }`, and only `skip` and `selector` are supported anyway.

So the freeze has to live inside Storybook.

## Fix

`.storybook/preview-head.html` — Storybook injects it into the `iframe.html` that pixeleye loads. Durations are `1ms` rather than `none` so animations settle on their end state instead of snapping back to the base style, and negative delays skip any start offset.

Side benefits: the storybook interaction tests no longer race transitions, and the blinking caret in a focused input stops being its own source of frame noise.

## Testing

Not verified end-to-end locally — this needed a full Storybook build plus a pixeleye capture against the VM. The check is that the `Loading` stories come back unchanged on the next advisory pixeleye run on this PR.

## Not covered

- Angular animations from `provideAnimations()` use the Web Animations API, not CSS keyframes, so `animation-duration` does not touch them. No current story appears to depend on one, but it is the next thing to look at if a diff survives.
- A story captured sometimes mid-load and sometimes loaded is a non-deterministic stub, not an animation frame, and needs the story fixed instead.


derschnee68:
Build #23 flagged a diff on **Ontology Editor / 1. Header / Edit Ontology Form Dialog** — the red region is the `Label` form field, focused in one snapshot and unfocused in the other. That was the CSS-only fix being incomplete, and in this case making things worse.

Angular Material times its own transitions in JS, not from CSS. `MatDialogContainer._startOpenAnimation` does:

```js
this._hostElement.style.setProperty('--mat-dialog-transition-duration', `${this._enterAnimationDuration}ms`);
this._requestAnimationFrame(() => this._hostElement.classList.add(OPENING_CLASS, OPEN_CLASS));
this._waitForAnimationToComplete(this._enterAnimationDuration, this._finishDialogOpen); // setTimeout(cb, 150)
```

`OPEN_ANIMATION_DURATION` is the constant `150`, and since MatDialog sets `delayFocusTrap: true` the initial focus is only trapped in `_finishDialogOpen`. Freezing the CSS made the dialog *look* fully open immediately while its first field stayed unfocused for another 150 ms — a wider window for the capture to land in, not a narrower one.

Second commit provides `MATERIAL_ANIMATIONS: { animationsDisabled: true }` in the storybook preview. Material then resolves the dialog open on a microtask and traps focus at once, and `mat-progress-spinner` switches to the static `_mat-animation-noopable` ring it ships for this purpose (which also zeroes `stroke-dasharray`, so the spinner is a full circle rather than an arc frozen at a keyframe boundary).

Stories adding their own `provideAnimations()` do not override it — that sets `ANIMATION_MODULE_TYPE`, a different token, and storybook's `applicationConfig` concatenates preview and story providers rather than replacing them.

The `preview-head.html` freeze stays for what Material does not own (CDK overlay backdrop fade, hand-written keyframes).

Expect one more round of snapshots to need approval: spinners become full rings, dialogs render settled.

domsteinbach (approved):

PR #3450, opened 2026-09-16
fix(advanced-search): show empty state instead of an endless progress bar when a project has no data model (DEV-6738)

domsteinbach:
## Problem

Opening the **Search** tab of a project that has no data model shows a query-mode progress bar that never resolves.

`OntologyDataService.loadOntologies` settles `_ontologyLoading` only inside `setOntology`, which it calls just when the ontology list is non-empty. For a project without a data model the list comes back empty, `setOntology` never runs, and the loading flag stays `true` forever.

## Fix

- `ontology-data.service.ts`: settle `_ontologyLoading` on the empty-list branch too, so loading terminates whether or not an ontology was found.
- `advanced-search-bar.component.ts`: when loading is done and there are no ontologies, render an explanatory empty state instead of the search form.
- i18n: new `pages.search.advancedSearch.noDataModel` key in de/en/fr/it.

## Testing

- `nx lint vre-pages-search-advanced-search` passes.
- `nx test vre-pages-search-advanced-search`: all 96 tests that run pass. 9 suites fail to resolve `open-api/src/generated/index`, which is absent in a local checkout until `npm run update-openapi` runs — unrelated to this change and reproducible on a clean tree.


derschnee68:
@claude review

mpro7 (approved):

PR #3453, opened 2026-09-17
feat(project-settings): replace the Off image setting with Default (DEV-7244)

subotic:
Part of DEV-7244, gate 3 of the release train. Follows dsp-api#4329 (merged 2026-09-16, deployed to dev).

## Summary

The "Off" radio encoded "no restriction wanted" as the IIIF size `pct:100`, a restriction that restricts nothing. SIPI wave-2 refuses that decision, so the option had to become a real one: **"Default"** clears the project's stored setting and lets the platform default apply.

The access change itself shipped in dsp-api#4329, which migrated `pct:100` out of the triplestore. This is the UI catching up, so it carries no further end-user impact of its own.

### The radio
- `ImageSettingsEnum.Off` becomes `Default`. Selecting it issues `DELETE .../RestrictedViewSettings` instead of posting `pct:100`.
- `getSizeForRequest` has no `pct:100` branch left, so the form cannot express it by any path. Percentage is bounded 1–99 and absolute width 128–1024, so no combination reaches it either.
- The Default radio names the inherited limit. This is new markup: the template's only expandable region was gated on `RestrictImageSize`. The component exposes the limit as structured values (`inheritedDimensions` / `inheritedPercentage`) and the template translates them, so full sentences stay in the i18n files rather than being assembled in TypeScript.

### Bugs found on the way
- **Load would have rendered the default state as a 128px restriction.** The new payload always carries populated `settings`, so the old sparse-payload handling fell through to `setRestrictedSize`. Selection now reads `isDefault` first, then `watermark`, then `size`.
- **Submit was enabled before the user touched anything.** `hasChanges` compared a response object against a request shape via `JSON.stringify`, which never matched. It now compares a tri-state derived from both the loaded response and the current form, and resets after a successful save.
- **A cleared percentage field left the Submit guard open.** imask leaves `''`, not `null`, so `isPercentageSize` stayed true while `parseInt('')/100` was `NaN` — and `NaN <= 0` is false, so Submit stayed enabled and posted `size: "pct:"`. `ratio` now reads any non-numeric value as 0.
- **A stored setting that is neither a watermark nor a size crashed the page.** `size` is optional on the wire; the rewrite asserted past it with `!` after deleting the guard that made that safe. It now falls back to Default.

### Plumbing
- `getRestrictedViewSettingsForProject` moves off `ProjectApiService` onto the generated client; it had a single caller. The loaded response collapses from a union to the one type all three verbs now return.
- **Neither save nor load takes an error callback.** An earlier revision added them, on the premise that a bare `.subscribe(next)` swallowed failures. It does not: RxJS reports an unhandled error to Angular's `ErrorHandler`, and this app's `AppErrorHandler` turns the status into a specific message for 0, 400, 403, 404, 409 and 504. Handling it locally replaced all six with one flat sentence, in a success-styled snackbar. Both call sites now say so, so the omission is not read as an oversight.
- `ARCH-MAP.md`'s recorded spec pin moves with the spec it describes.
- `aria-describedby` ties the Default hint to its radio, so it is not sighted-only.
- `dsp-api_spec.yaml` regenerated (`v38.1.0-5-g35a36e3` → `v39.0.0-17-g9ddad36`). This supersedes #3452, which carried the byte-identical bump but could not build without these component changes.

## Test Plan

- [x] `nx test vre-pages-project-project` — **239 passed, 20 suites**, including 15 new cases
- [x] `nx lint vre-pages-project-project` — clean
- [x] New `image-settings.component.spec.ts`: no spec existed. Covers each load state including the size-less one that used to crash, Submit disabled on a fresh load, the cleared-percentage guard, DELETE vs POST per radio, no `pct:100` on any path, Submit re-disabled after save, and that a failed save confirms nothing.
- [x] New `image-settings.component.stories.ts`: one story per load state, plus one asserting Submit enables after switching to Default.
- [x] New `image-settings.cy.ts`: one linear walk (stored size → reload → Default → DELETE → reload lands on Default with `isDefault: true` → watermark → POST), asserting no request body ever contains `pct:100`. Written as a single test on purpose: every step mutates the project's stored setting, so independent tests would depend on each other's leftovers.
- [x] Registered in the CI matrix (`rest` runner). System-admin specs are enumerated individually, so an unregistered spec never runs.

### Verified by hand

**Rendering**, from the three Storybook states: nothing stored lands on Default with "Images are limited to 128 × 128 pixels, the platform default." and Submit disabled; a stored `pct:13` lands on Restrict image size with the percentage field and ratio preview populated and no hint line; a stored watermark lands on Apply watermark. Console clean.

**The round trip**, against a local stack built from dsp-api `main` at `9ddad36b7` (`webapi v39.0.0-17-g9ddad36`, `sipi v9.1.0`):

```
GET     → {"settings":{"size":"!128,128","watermark":false},"isDefault":true}
POST    {"size":"pct:13"}   → 200, isDefault false
DELETE                      → back to !128,128, isDefault true
POST    {"size":"pct:100"}  → 400 Invalid RestrictedViewSize: pct:100
```

That last line is the premise of this change confirmed end to end: the old "Off" payload is now a hard 400, which is why the radio had to become a real Default.

## A dsp-api bug this surfaced — DEV-7292

The OpenAPI spec describes the POST body as wrapping its values, and the server rejects that shape:

```
POST {"size":"pct:13"}           → 200   ← what this component sends
POST {"size":{"value":"pct:42"}} → 400   ← what the published schema describes
   Invalid value for: body (expected string at 'size')
```

Tapir derives the schema from the `RestrictedView.Size` case class (`{value: string}`) while the codec serialises it as a bare string. So the generated client types this body in a shape the server refuses.

This is pre-existing — `SetRestrictedViewRequest` is unchanged by #4329 — and it is **why this PR leaves the POST payload shape untouched**. `getRequest` casts explicitly rather than hiding behind `any`, and both it and the spec assertion carry a comment naming DEV-7292, so a future "fix the type" pass finds the reason before the compiler pushes it into the wrong shape. Once the schema is fixed, the cast can go.

## Review

Six reviewer agents (angular, typescript, consistency, simplicity, accessibility, devops) plus a dune legibility audit ran over this diff, and every finding below was verified against the source before acting on it. What they caught, all now fixed: the size-less crash; the error handling that bypassed `AppErrorHandler`; error text rendered in a success-styled snackbar; a stale `ARCH-MAP.md` pin; dead `currentSettings` state; the missing `aria-describedby`; and a Cypress load/render race. Contrast on the new hint was computed rather than guessed (5.74:1 on white, 5.66:1 on `#FAFAFA`) and passes.

The `@claude` review then found the cleared-percentage guard, which is the one live bug none of the others reached, plus a Storybook title that was a strict prefix of its siblings' — both fixed here.


BalduinLandolt (approved):
LGTM, though I think it would make more sense to ask for review from someone who actually works on dsp-app

mpro7 (approved):
Nothing from my side to add. Run Claude to test agains recently added ADRs :)

mpro7:
@claude review

subotic:
Thanks both. Everything actionable is now in — force-pushed onto the current `main` (which picked up the flake from #3454).

**@mpro7 on the ADRs** — checked against `docs/adr/0001`–`0006` from #3423. No violations: ADR-0002 (the barrel is the public interface) holds, since every cross-library import goes through a barrel and the one newly exported symbol is consumed by a same-directory relative import; ADR-0006 puts the Cypress suite at `apps/dsp-app/cypress/`, which is where the new spec lives.

**The `@claude` review found one live bug, now fixed.** The cleared-percentage case is real. imask leaves `''` rather than `null`, so `isPercentageSize` stays true, `parseInt('')/100` is `NaN`, and `NaN <= 0` is **false** — Submit stays enabled and posts `size: "pct:"`. It was graded "Suspected" for want of a running app, but the fix is right either way: a getter returning `NaN` is wrong regardless of what imask writes. `ratio` now reads any non-numeric value as 0, with a spec case pinning it.

Also applied from that review: the Storybook title was a strict *prefix* of the sibling `.../Image Display Absolute` titles, which collapses the sidebar group and the component entry into one name — nested to `.../Image Settings / Image Settings Form`. And the Cypress walk now asserts `isDefault === true` on first load, so a contaminated start fails loudly rather than skewing later steps.

**A separate review pass turned up more, all fixed here.** Six reviewer agents plus a dune legibility audit ran over the diff, and I verified each finding against the source rather than taking it on trust:

- **A crash.** `settings.size` is optional on the wire, and the rewrite asserted past it with `!` after deleting the guard that made that safe. Worse, the throw is synchronous inside `next`, which RxJS routes to `reportUnhandledError` — so the error callback I had added for exactly this could never catch it. Now falls back to Default, with a spec case.
- **The error handling was a regression, not a fix.** My commit message claimed a failed save was "silently swallowed". It wasn't: RxJS reports unhandled errors to Angular's `ErrorHandler`, and `AppErrorHandler` here has specific messages for 0/400/403/404/409/504. Adding `error:` callbacks replaced all six with one flat sentence — and, because `openSnackBar` defaults to `type: 'success'`, rendered failures in a blue success toast. Both callbacks are gone; both call sites now carry a comment so the omission isn't "fixed" back.
- Dead `currentSettings` state removed, `ARCH-MAP.md`'s spec pin moved with the spec it describes, and `aria-describedby` now ties the Default hint to its radio so it isn't sighted-only.

Contrast on the new hint was computed rather than guessed — 5.74:1 on white, 5.66:1 on `#FAFAFA` — and passes AA.

239 tests, 20 suites, lint clean. Worth saying plainly: every defect above was in the error handling and robustness I added, not in the `Off` → `Default` change itself, and one of them existed only because I "fixed" a bug that was never there.

subotic:
Thanks, fixes applied.

mpro7:
@subotic can the [open-Api version bump](https://github.com/dasch-swiss/dsp-app/pull/3455) be merged now or the fix lives somewhere else?

PR #3460, opened 2026-09-21
fix(ontology): translate validator error messages instead of rendering the raw i18n key (DEV-7283)

domsteinbach:
Fixes [DEV-7283](https://linear.app/dasch/issue/DEV-7283/validator-error-messages-render-the-raw-i18n-key-instead-of-the).

## Problem

In **Create new data model**, a name violating the name pattern (e.g. `12312323`) rendered the literal translation key as the error text:

```
pages.ontology.ontologyForm.namePatternError
```

Same failure in **Create resource class** (`pages.ontology.resourceClassForm.patternError`).

This is not a missing translation — both keys exist and are correctly translated in all four locales. They were simply never looked up.

## Root cause

`HumanReadableErrorPipe` returns the `message` from `validatorErrors` verbatim, so the convention is that the **caller** resolves the key first. Two ontology call sites passed the raw key; the ~10 others (`user-form`, `reusable-project-form`, `iiif-control`) already call `instant()`.

## Fix

Resolve both messages through `TranslateService.instant()`, consistent with the existing call sites:

- `create-ontology-form-dialog.component.ts` — resolve `ontoNamePatternErrorMsg.message`.
- `resource-class-form.component.ts` — the message was inline in the template, which is why it could not call `instant()`; moved into the class as `namePatternErrorMsg`.

No locale files change.

### Note on injection style

`TranslateService` is injected via `inject()` as a field declared above the initializers that read it, rather than as a constructor parameter. A field initializer reading a constructor-param property only works because this repo sets `useDefineForClassFields: false`; under standard class-field semantics it would be `undefined`. `inject()` is order-explicit, independent of that flag, and matches the repo-dominant style (47 uses repo-wide, and `iiif-control.component.ts` among the conforming call sites).

## Scope

Deliberately limited to the reported bug. The issue lists three related findings in the same pipe, all still open and worth separate issues:

- (a) the built-in branches (`required`, `minlength`, `existingName`, …) return hardcoded English, so they stay English in DE/FR/IT;
- (b) `ontoNameExistsErrorMsg` is unreachable (duplicate `errorKey: 'pattern'`, and `existingName` is caught earlier), making the `nameExistsError` string dead in all four locale files;
- (c) the pipe throws from inside a template expression for unrecognised validator errors.

## Testing

- `nx lint vre-pages-ontology-ontology` passes.
- `nx test vre-pages-ontology-ontology` passes (1 suite — coverage in this lib is thin).
- Typecheck: 118 errors before the change, 118 after — identical. All are the pre-existing missing `open-api/src/generated/` client (absent until `npm run update-openapi`) and missing `@types/jsonld`; neither edited file appears in the output.
- Both keys verified present and translated in `en`/`de`/`fr`/`it`.

Not manually reproduced in a running app.


derschnee68 (approved):

domsteinbach:
Pushed a follow-up: with the message now actually translated, its length exposed a second problem — it was clipped under the field.

**Cause.** `mat-error` renders inside the form field's subscript wrapper. By default that wrapper reserves a single line and positions the error `absolute`, so a longer message overflows and is cut off. Measured in a browser against Material's own rules at the dialog's width:

| | error text needs | wrapper height | clipped |
|---|---|---|---|
| before | 24px (2 lines) | 16px | yes |
| after (`subscriptSizing="dynamic"`) | 24px (2 lines) | 24px | no |

**Fix.** `subscriptSizing="dynamic"` on `CommonInputComponent`'s form field switches the wrapper to static positioning so it grows to fit the wrapped text. This is already the established answer to this problem in the repo — six other form fields use it (`pdf-toolbar`, `multi-language-textarea`, `resource-side-legal-form`, `advanced-search-bar`, `resource-rights-statement`, `property-value-basic-comment`).

**Blast radius.** `CommonInputComponent` has 23 usages across 15 files. The attribute only changes layout while a message is shown, and the component renders no `mat-hint`, so fields are unchanged otherwise. I checked the consumers for styling that assumed a fixed subscript height — none.

**Test.** Added `WrapsLongValidationError` using the real DEV-7283 message at a constrained width. I verified the assertion discriminates by running it against both the old and new CSS: `wrapper.height >= error.height` is **false** before the fix and **true** after, so it fails if this regresses.

Note `nx run vre-ui-ui:storybook` in CLAUDE.md is stale — Storybook is hosted on the `dsp-app` project (`nx run dsp-app:storybook`), with one root `.storybook/` covering all libs. I could not run the interaction tests locally: `dsp-app:build-storybook` fails with an ESM/CJS error in `@storybook/angular` + `@angular-devkit/architect`, identically on a clean tree. So the new story is verified by its underlying CSS measurement, not by a local test-runner pass — worth confirming in CI.

